PR #21, opened 2019-11-18
Switch to First.org JSON schemas for CVSS scores

ericjohnson-tibco:
We should probably wait until others have had a chance to chime in on the various issues raised via email, before this is merged.


santosomar:
Patrick from AT&T observed that there is a typo in the schema. A JSON array should have square brackets "[ ]" and the current schema has it with "{". I am doing a quick edit here to address that concern.

santosomar:
Complete discussion on this item at: https://lists.oasis-open.org/archives/csaf/201911/msg00019.html

PR #32, opened 2020-05-15
Document tracking artifacts

### LINKED ISSUE #24: Document tracking artifacts
The fields "document":"tracking":"version" and "document":"tracking":"revision_history" are not listed as required in CSAF 2.0 (https://github.com/oasis-tcs/csaf/blob/master/sandbox/csaf_2.0/json_schema/csaf_json_schema.json) anymore.  We believe that these information are easy to create and maintain and provide huge benefit to the audience. Therefore, the should stay mandatory (as they were in CVRF 1.2).

tschmidtb51:
- resolves oasis-tcs#24
- add missing required fields in "tracking"

### DIFF
--- a/sandbox/csaf_2.0/json_schema/csaf_json_schema.json
+++ b/sandbox/csaf_2.0/json_schema/csaf_json_schema.json
@@ -293,6 +293,7 @@
             "id",
             "initial_release_date",
             "revision_history",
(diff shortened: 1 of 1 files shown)


sthagen (approved):
The change looks good to me. 👍

PR #35, opened 2020-05-27
Restart required

### LINKED ISSUE #25: Boolean field: restart_required
Add an additional boolean field restart_required in "vulnerabilities":"remediations" to provide them this information in an easily accessible way.

tschmidtb51:
- resolves oasis-tcs/csaf#25
- add "restart_required" field in "remediations"

### DIFF
--- a/csaf_2.0/json_schema/csaf_json_schema.json
+++ b/csaf_2.0/json_schema/csaf_json_schema.json
@@ -622,6 +622,9 @@
                   "$ref": "#/definitions/products_t"
                 },
                 "restart_required": {
+                  "title": "Restart required by remediation",
+                  "description": "The field specifies whether this remediation requires a restart of the vulnerable component to become effective.",
+                  "examples": [true, false],
                   "type": "boolean"
                 },
                 "type": {
--- a/csaf_2.0/json_schema/csaf_json_schema.json
+++ b/csaf_2.0/json_schema/csaf_json_schema.json
@@ -623,9 +623,35 @@
                 },
                 "restart_required": {
                   "title": "Restart required by remediation",
-                  "description": "The field specifies whether this remediation requires a restart of the vulnerable component to become effective.",
-                  "examples": [true, false],
-                  "type": "boolean"
+                  "description": "Provides information on type of restart is required by this remediation to become effective.",
+                  "type": "object",
+                  "required": [
+                    "type"
+                  ],
+                  "properties": {
+                    "type": {
+                      "title": "Type of restart",
+                      "description": "Specifies what type of restart is required by this remediation to become effective.",
+                      "type": "string",
+                      "enum": [
+                        "none",
+                        "vulnerable_component",
+                        "service",
+                        "parent",
+                        "dependencies",
+                        "connected",
+                        "machine",
+                        "zone",
+                        "system"
+                      ]
+                    },
+                    "description": {
+                      "title": "Additional restart information",
+                      "description": "Provides additional information for the restart. This can include details on procedures, scope or impact.",
+                      "type": "string",
+                      "minLength": 1
+                    }
+                  }
                 },
                 "type": {
                   "enum": [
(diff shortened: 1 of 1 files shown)


sthagen:
Maybe add if service or machine restart is required?

tolim:
> Maybe add if service or machine restart is required?

My suggestion here would be to add the following properties to the `restart_required` object:
* add `type` as enum `[system|machine|other]` to describe the component that is affected by the restart. `machine` requires a complete restart of the machine, `service` requires only a restart of the updated service, `other` may be used e.g. if a complete cluster must be restarted (neither machine nor service)
* add `description` as string that allows to give further details about the required restart. If `type: other`, this field should be compulsory

tschmidtb51:
I though about it and came up with a more comprehensive approach. Please see the pull request for more details.

tschmidtb51:
Please have a look at the list of definitions for the meaning of the enum values:
"none" - No restart required.
"vulnerable_component" - Only the vulnerable component needs to be restarted.
"service" -  The vulnerable component and the background service used by the vulnerable component need to be restarted.
"parent" - The vulnerable component and its parent process need to be restarted. This could be the case if the parent process has no build-in way to restart the vulnerable component or process values / context is only given at the start of the parent process. Think of something like jvm parameters.
"dependencies" - The vulnerable component and all components which require the vulnerable component to work need to be restarted. This could be the case e.g. for a core service of a software.
"connected" - The vulnerable component and all components connected (via network or any type of inter-process communication) to the vulnerable component need to be restarted. 
"machine" - The machine on which the vulnerable component is installed on needs to be restarted. This is the value which should be used if an OS needs to be restarted. It is typically the case for OS upgrades.
"zone" - The security zone in which the machine resides on which the vulnerable component is installed needs to be restarted. This value might be useful for a remediation if no patch is available. If the malware can be wiped out by restarting the infected machines but the infection spreads fast the controlled shutdown of all machines at the same time and restart afterwards can leave one with a clean system. 
"system" - The whole system which the machine resides on which the vulnerable component is installed needs to be restarted. This may include multiple security zones. This could be the case for a major system upgrade in an ICS system or a protocol change.

sthagen (approved):

tschmidtb51:
I would like to add some examples for these types:
- **"none"** - kill a process like telnet
- **"vulnerable_component"** - Notepad++ portable
- **"service"** - Nextcloud with nginx issue (php-fpm, see: https://nextcloud.com/blog/urgent-security-issue-in-nginx-php-fpm/)
- **"parent"** - TIA portal when vulnerability in WinCC
- **"dependencies"** - pppd
- **"connected"** - TLS without session renegotiation
- **"machine"** - Windows 10 semi-annual upgrade
- **"zone"** - WannaCry infecting industrial PCs which have the Write Filter enabled; Windows reboots after infection when WannaCry tries to encrypt the disk which removes WannaCry from this PC. However, other PCs in the same security zone still have WannaCry and reinfect each other. To get a clean zone, all equipment must be turned off and be rebooted afterwards.
- **"system"** - Replace a protocol like DeltaV.

tolim:
Adding this many restart types will add much complexity and will cause confusion. We have many different types of vulnerable components, and corner cases will cause problems. Users usually only want to know whether only the application/service or the whole machine (the OS) needs to be restarted. 

So I would suggest to restrict the `types` to `none`, `machine` and `other`, whereby in all cases an additional description with context-specific information is recommended.

sthagen:
In the good old network/json tradition, I suggest to not invent a non-state like otter. Instead trust on consumers only acting on fields they understand and are prepared for and ignore all things they do not. Then we can define several conformance levels for the consumers. Remember, we want automation. Text as value for an other field we have since decades already, kind of.

tschmidtb51:
I agree that it adds complexity. However, I think we should be as specific as possible. This will aid in making the format better processable for computers - running text only brings it back to a human. Don't get me wrong: I agree that we should have a description for additional information for the restart (including details on procedures, scope or impact).
I disagree on your comment regarding the vulnerable components: It was not clear in the definition above - my mistake. I updated the definition. But the "vulnerable_component" is clearly defined by the given values of the array "product_ids" in the remediation item where the "restart_required" object is a part of.

PR #39, opened 2020-06-04
Fixes for example files

tolim:
When running the automated tests, I encountered multiple problems in the two provided examples as they did not conform to the JSON schema. This pull request fixes those problems in the example files.


santosomar:
There are several conflicts in the branch that may need to be addressed before merging.

tolim:
The conflicts were removed.

tschmidtb51:
I just added a few minor changes to be compatible to the current version of the schema:
- add "type": "vendor" in "publisher"
- change "distribution" to be an object
- removed object "identification" in "tracking"

PR #42, opened 2020-06-08
Publisher artifacts

### LINKED ISSUE #41: Publisher artifacts
The field "document":"publisher":"type" is not required in the current JSON-Schema. This breaks compatibility to CVRF 1.2 which requires the field (see CSAF-4.4-1).

tschmidtb51:
- resolves #41 
- set "type" required in "publisher"


santosomar (approved):
Thank you. The changes make sense and does not affect any other elements in the schema. I will merge now and we will discuss the changes in the June monthly meeting.

PR #46, opened 2020-06-13
Version type

### LINKED ISSUE #45: Version type
The definition of "version" is used multiple times. Following the DRY principle this should be extracted to a type.

tschmidtb51:
- resolves #45 
- add type "version_t"
- refactor to use new type


santosomar:
Thank you for your contribution, @tschmidtb51 !

csaf_2.0/json_schema/csaf_json_schema.json
```diff
@@ -182,6 +182,13 @@
           "description"
         ]
         }
+    },
+    "version_t": {
+      "title": "Version",
+      "description": "Sepcifies a version string with a simple hierarchical counter model to denote clearly the evolution of the content of the document. Format must be understood as 'major.minor.patch.build' version.",
+      "examples": ["1", "0.9", "1.4.3", "2.40.0.320002"],
+      "type": "string",
+      "pattern": "^(0|[1-9][0-9]*)(\\.(0|[1-9][0-9]*)){0,3}$"
```
sthagen, line 191: I suggest to modify this regex to allow calendar versioning variants that hint explicitely at that semantic:
>`^(0|[1-9][0-9]*)(\\.([0-9]+)){0,3}$`

that way a version like `21.04` Is valid - which would help producers that have set up their software development to use calendar versioning (because intrinsic semantic versioning just did not work well for them) - the results are still orderable in any packaging relevant context.

Regardless of this proposed change:
> Can we help the ecosystem / communities by distilling actionable rules as guidance from the members working in the field which classes of changes would be expected to impact which degree of change indicator? 

Given the tight coupling between the advisories and the expected reaction upon receipt, I imagine we should be able to help here - if it is only informative it will still help to foster communication between producers and consumers. Even better, when we can normatively state a SHOULD with confidence.

PR #60, opened 2020-06-23
Small schema fixes regarding enums

tolim:
- replaced whitespaces with underlines in enum texts
- added missing type `string` for some enums
- added some titles and descriptions for changed enums


santosomar:
thank you for your contribution @tolim !

PR #63, opened 2020-06-23
CSAF_version artifact

### LINKED ISSUE #23: CSAF version artifact
To be prepared for coming changes in the document specification CSAF 2.0 should contain a version field that indicates the CSAF-Version the document was created with. This enables the parser to check for the correct syntax and find the right fields even if they would change from version to version. Furthermore, it enables customers and asset owners to require a certain minimum standard on the security advisory once it is widely adopted.

tschmidtb51:
- resolves #23 
- add "csaf_version" with value "2.0"
- add "csaf_version" n required fields of "document"
- add "csaf_version" in examples to fix tests


santosomar:
Reviewed and approved by TC during the June 24, 2020 monthly meeting.

PR #65, opened 2020-06-23
CSAF $id property

### LINKED ISSUE #64: CSAF $id property
The schema should have an $id property. 
"The $id property is a URI-reference that serves two purposes:
- It declares a unique identifier for the schema.
- It declares a base URI against which $ref URI-references are resolved.

It is best practice that every top-level schema should set $id to an absolute-URI (not a relative reference), with a domain that you control."
[https://json-schema.org/understanding-json-schema/structuring.html#the-id-property](url)

tschmidtb51:
- resolves oasis-tcs/csaf#64
- add "$id" as property with value pointing to Github


santosomar:
Reviewed and approved by TC during the June 24, 2020 monthly meeting.

santosomar (approved):
Reviewed and approved by TC during the June 24, 2020 monthly meeting.

Issue #72, opened 2020-06-26
Remove system_name


tschmidtb51:
The type definition of "system_name" in vulnerabilities seems to be misplaced. The definition in "id" is the correct one according to CVRF 1.2 Section 6.3 (see [https://docs.oasis-open.org/csaf/csaf-cvrf/v1.2/cs01/csaf-cvrf-v1.2-cs01.html#_Toc493508716](https://docs.oasis-open.org/csaf/csaf-cvrf/v1.2/cs01/csaf-cvrf-v1.2-cs01.html#_Toc493508716). Therefore, the one mentioned first is a duplicate and should be removed.

PR #73, opened 2020-06-26
Remove system_name

tschmidtb51:
- resolves oasis-tcs/csaf#72
- delete duplicate "system_name"


tolim:
@sthagen Done, I just added the corresponding commit.

PR #76, opened 2020-06-26
Product_status artifacts

### LINKED ISSUE #75: Product_status artifacts
The object "product_status" should have a least one property as stated in the specification of CVRF 1.2 (see CSAF-6.10-1).

tschmidtb51:
- resolves #75 
- add "minProperties" to the definition

### DIFF
--- a/csaf_2.0/json_schema/csaf_json_schema.json
+++ b/csaf_2.0/json_schema/csaf_json_schema.json
@@ -673,6 +673,7 @@
           },
           "product_status": {
             "type": "object",
+            "minProperties": 1,
             "properties": {
               "fixed": {
                 "$ref": "#/definitions/products_t"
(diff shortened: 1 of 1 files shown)


tolim:
@sthagen I ordered the properties semantically: first the group for fixed versions, then affected, then not affected. Ordering by name does not really help in my opinion. Do you agree?

sthagen:
> @sthagen I ordered the properties semantically: first the group for fixed versions, then affected, then not affected. Ordering by name does not really help in my opinion. Do you agree?

Ordering by name helps a lot with any longer sequence for humans to read. If we group, then normally the grouping is adding a level (or an aggregation). The ordering itself in the same level not matching the structural/topological one is irritating at best (in my experience).

santosomar:
Discussed in the July 29th, monthly TC meeting. It was agreed to merge the changes as suggested in the pull request.

PR #78, opened 2020-06-26
New Product_status: under_investigation

### LINKED ISSUE #77: New Product_status: under_investigation
Sometimes a vendor is unable to tell which devices are affected but wants to inform that he is investigating the issue. This might be the case for a hardware vulnerability (e.g. Spectre, Meltdown, etc.). Therefore, we suggest a new product status "under_investigation" which allows a vendor (or any other publishing party) to state for which products a customer can expect to get a definitive / authoritative statement on the product status regarding the current vulnerability.

tschmidtb51:
- resolves #77
- add new property "under_investigation" to "product_status"

### DIFF
--- a/csaf_2.0/json_schema/csaf_json_schema.json
+++ b/csaf_2.0/json_schema/csaf_json_schema.json
@@ -694,6 +694,11 @@
               },
               "last_affected": {
                 "$ref": "#/definitions/products_t"
+              },
+              "under_investigation": {
+                "title": "Under investigation",
+                "description": "It is not known yet whether this version is or is not affected by the vulnerability. However, it is still under investigation - the result will be provided in a later release of the document.",
+                "$ref": "#/definitions/products_t"
               }
             }
           },


sthagen (commented):
Questions to those members actually issuing Security Advisories: Is this a Useful state of being in ?Rationale: sounds more like a promise to me not bringing much actionable content to the recipient but putting pressure on the producer.

santosomar:
This could be very useful when vendors are responding to open source or zero day vulnerabilities, where most of their product status is labeled under investigation.

santosomar:
I will mark this for TC discussion. However, the enhancement makes sense from my side.

tolim:
> This could be very useful when vendors are responding to open source or zero day vulnerabilities, where most of their product status is labeled under investigation.

I agree, it could be useful in cases where users have a SBOM of the product and know about 3rd party component vulnerabilities.

sthagen (approved):
The "promise" as stated in the description and conveyed by the status may turn into a liability of unclear magnitude/risk for the vendor ... but from consumer perspective this is a good status to know (as compared to just waiting) :+1:

PR #82, opened 2020-06-26
Prohibit empty arrays

### LINKED ISSUE #81: Prohibit empty arrays
The schema uses arrays in the definition. However, it still validates successful if they are empty which could allow defeating a required information by putting in an empty array. I suggest to prohibit empty arrays - if no information is present, the field should not be present as well.

tschmidtb51:
- resolves #81 
- add "minItems" in properties
- refactoring

### DIFF
--- a/csaf_2.0/json_schema/csaf_json_schema.json
+++ b/csaf_2.0/json_schema/csaf_json_schema.json
@@ -130,6 +130,7 @@
     },
     "notes_t": {
       "type": "array",
+      "minItems": 1,
       "items": {
         "type": "object",
         "properties": {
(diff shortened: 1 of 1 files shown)


sthagen:
Hm, I like empty arrays. Acknowledgements etc. may very well be not available due to various reasons

tolim:
Looking at the case of acknowledgments we have two choices:
1. We do not allow empty arrays: if no acknowledgments are contained in the advisory, then the whole `document->acknowledgments` property must be removed.
2. We allow empty arrays: the author of the advisory now has two choices: he removes the `document->acknowledgments` property, or he keeps it and defines an empty array. Now we have redundancy in the definition.
I prefer to use option 1, as redundancy often adds unneeded complexity. But this is only valid for parent properties that are not required. We still need to check the document for that aspect.

santosomar (approved):

PR #84, opened 2020-06-26
Prohibit empty objects

### LINKED ISSUE #83: Prohibit empty objects
The schema uses objects in the definition. However, it still validates successful if they are empty which could allow defeating a required information by putting in an empty object. I suggest to prohibit empty objects - if no information is present, the object should not be present as well.

tschmidtb51:
- resolves #83 
- add "minProperties"

### DIFF
--- a/csaf_2.0/json_schema/csaf_json_schema.json
+++ b/csaf_2.0/json_schema/csaf_json_schema.json
@@ -7,6 +7,7 @@
   "definitions": {
     "acknowledgment_t": {
       "type": "object",
+      "minProperties": 1,
       "properties": {
         "names": {
           "type": "array",
--- a/csaf_2.0/json_schema/csaf_json_schema.json
+++ b/csaf_2.0/json_schema/csaf_json_schema.json
@@ -420,6 +420,7 @@
             },
             "generator": {
               "type": "object",
+              "minProperties": 1,
               "properties": {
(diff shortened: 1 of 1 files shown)


csaf_2.0/json_schema/csaf_json_schema.json
```diff
@@ -418,6 +420,7 @@
             },
             "generator": {
               "type": "object",
+              "minProperties": 1,
```
sthagen, line 423: The generator might be private to the producer. Not sure we should enforce in the public API - this may lead to disguised or faked or ignored fields in practice. But I have no hard feelings about this.

tolim: > 
> 
> The generator might be private to the producer. Not sure we should enforce in the public API - this may lead to disguised or faked or ignored fields in practice. But I have no hard feelings about this.

In this case the `generator` property can be completely removed from the document. I prefer to remove this redundancy, so I would approve this change.

tolim (approved):

csaf_2.0/json_schema/csaf_json_schema.json
sthagen: Well if some organization wants to publish the generator, I guess they should be able. Others can simply not provide the member and value.

PR #86, opened 2020-06-26
Prohibit empty strings

### LINKED ISSUE #85: Prohibit empty strings
The schema uses strings in the definition. However, it still validates successful if they are empty which could allow defeating a required information by putting in an empty string. I suggest to prohibit empty strings - if no information is present, the field should not be present as well.

tschmidtb51:
- resolves #85 
- add "minLength" in properties

### DIFF
--- a/csaf_2.0/json_schema/csaf_json_schema.json
+++ b/csaf_2.0/json_schema/csaf_json_schema.json
@@ -128,10 +128,16 @@
         "type": "object",
         "properties": {
           "audience": {
+            "title": "Audience of note",
+            "description": "Indicate who is intended to read it.",
+            "examples": ["all", "executives", "operational management and system administrators", "safety engineers"],
             "type": "string",
             "minLength": 1
           },
           "title": {
+            "title": "Title of note",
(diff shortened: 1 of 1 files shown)


csaf_2.0/json_schema/csaf_json_schema.json
```diff
@@ -262,10 +270,12 @@
           "type": "object",
           "properties": {
             "namespace": {
-              "type": "string"
+              "type": "string",
+              "format": "uri"
```
sthagen, line 274: I am so used to namespaces not being URIs that I wonder if we better leave the refined type to the producer and consumer?

santosomar:
Please try to resolve the conflicts before approving the merge of the pull request.

tolim:
> Please try to resolve the conflicts before approving the merge of the pull request.

We had too many changes to the document, so this pull request got some conflicts. Those are removed now.

sthagen:
Document properties like `title`, `description`, etc. of our CSAF DSL make me nervous as part of the mix into json schema language which uses these "keys" for generic meta purposes :wink:

Issue #98, opened 2020-07-07
Fix multiple full_product_names in relationship


tolim:
It is not conclusive to me why we allow multiple `full_product_names` in `product_tree/relationships`. One `relationship` element links two products and may produce one new product name and identifier in `full_product_name`. In this context it does not make sense to allow multiple `full_product_names`.

Btw, this is explicitly allowed by CVRF 1.2:

> The prod:Relationship element MUST be present with cardinality [0, ∞] in prod:Tree and if given MUST contain one or more prod:FullProductName instances. » [CSAF-5.1.3-1]

At the moment we do not require the property `full_product_name` within `product_tree/relationships`. This does not conform to CVRF 1.2, but I propose not to change this.

sthagen:
As stated in my review of the pull request I can easily imagine many components together constructing the insecure situation the advisory tries to mitgate: An nginx server running in a docker container with a specific hypervisor on a specific operating system when ipv6 is active and the application server is a so-and-so in a such-and-such configuration ...

santosomar:
This was addressed in pull request #99

PR #99, opened 2020-07-07
Disallow multiple full_product_names in product relationship

tolim:
Fixes issue #98 
- rename 'full_product_names' to 'full_product_name'
- remove array within node
- fix example

### DIFF
--- a/csaf_2.0/examples/cvrf-rhba-2018-0489-modified.json
+++ b/csaf_2.0/examples/cvrf-rhba-2018-0489-modified.json
@@ -4772,4731 +4772,3871 @@
         "product_reference": "ansible-asb-modules-0:0.1.1-1.el7",
         "relationship_type": "default_component_of",
         "relates_to_product_reference": "7Server-RH7-RHOSE-3.9",
-        "full_product_names": [
-          {
-            "product_id": "7Server-RH7-RHOSE-3.9:ansible-asb-modules-0:0.1.1-1.el7",
-            "name": "ansible-asb-modules-0:0.1.1-1.el7 as a component of Red Hat OpenShift Container Platform 3.9"
-          }
-        ]
+        "full_product_name": {
+          "product_id": "7Server-RH7-RHOSE-3.9:ansible-asb-modules-0:0.1.1-1.el7",
+          "name": "ansible-asb-modules-0:0.1.1-1.el7 as a component of Red Hat OpenShift Container Platform 3.9"
+        }
       },
       {
         "product_reference": "ansible-kubernetes-modules-0:0.4.0-8.el7",
         "relationship_type": "default_component_of",
         "relates_to_product_reference": "7Server-RH7-RHOSE-3.9",
-        "full_product_names": [
-          {
-            "product_id": "7Server-RH7-RHOSE-3.9:ansible-kubernetes-modules-0:0.4.0-8.el7",
-            "name": "ansible-kubernetes-modules-0:0.4.0-8.el7 as a component of Red Hat OpenShift Container Platform 3.9"
-          }
-        ]
+        "full_product_name": {
+          "product_id": "7Server-RH7-RHOSE-3.9:ansible-kubernetes-modules-0:0.4.0-8.el7",
+          "name": "ansible-kubernetes-modules-0:0.4.0-8.el7 as a component of Red Hat OpenShift Container Platform 3.9"
+        }
       },
       {
         "product_reference": "ansible-service-broker-0:1.1.16-1.el7",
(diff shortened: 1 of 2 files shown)


tolim:
The references to the two related products are in the properties `product_reference` and `relates_to_product_reference`. `full_product_names` would only provide a referenceable `product_id` and a corresponding name for the relationship. The property `relationship_type` explicitly defines the type of the relationship. If you want to define two different relationships between components, you would need to create two entries in the `relationships` array.

Here is an excerpt from one of our examples:
```
   "relationships": [
      {
        "product_reference": "ansible-kubernetes-modules-0:0.4.0-8.el7",
        "relationship_type": "default_component_of",
        "relates_to_product_reference": "7Server-RH7-RHOSE-3.9",
        "full_product_names": [
          {
            "product_id": "7Server-RH7-RHOSE-3.9:ansible-kubernetes-modules-0:0.4.0-8.el7",
            "name": "ansible-kubernetes-modules-0:0.4.0-8.el7 as a component of Red Hat OpenShift Container Platform 3.9"
          }
        ]
      }
    ]
```

Still I can't think of any reason why someone would need to define two or more `full_product_names` for this relationship. Could you please give me an example?

sthagen:
I may miss something here, but did we receive input during the previous meeting from members mentioning alias like names or multiple names or even order numbers for the same product which would be examples for relationships most easily noted as a single n-component entry instead of n * (n-1) pair entries into the array. Please help me here. I further remember we wanted to add this reasoning to the description to help readers and implementers alike.

tschmidtb51:
I'm feeling the same way as @tolim - but maybe we miss something.

> Why would we want to replace an array defining a relationship between full_product_names by a single string?

The array is IMO not defining the relationship. It just names the "new" product which is formed by the relationship (see the excerpt @tolim mentioned above). IMO the "new" product should have **one** name and ID. If there are more than one full_product_names the following situation is possible (and syntactically correct):

`
"relationships": [
      {
        "product_reference": "ansible",
        "relationship_type": "default_component_of",
        "relates_to_product_reference": "7Server-RH7-RHOSE-3.9",
        "full_product_names": [
          {
            "product_id": "7Server-RH7-RHOSE-3.9:ansible1",
            "name": "ansible as a component of Red Hat OpenShift Container Platform 3.9"
          },
          {
            "product_id": "7Server-RH7-RHOSE-3.9:ansible2",
            "name": "ansible as a component of Red Hat OpenShift Container Platform 3.9"
          },
          {
            "product_id": "7Server-RH7-RHOSE-3.9:ansible3",
            "name": "ansible as a component of Red Hat OpenShift Container Platform 3.9"
          }
        ]
      }
    ]
`

and 
`
      "product_status": {
            "fixed": [
                 "7Server-RH7-RHOSE-3.9:ansible1"
            ],
            "known_affected": [
                "7Server-RH7-RHOSE-3.9:ansible2"
            ],
            "known_not_affected": [
                "7Server-RH7-RHOSE-3.9:ansible3"
            ]
         }
`
However, this is a situation which we definitely don't want. It would say that the same product is `"fixed"`, `"known_affected"`, `known_not_affected` at the same time. (I think this is even not possible for quantum software ;-))

For the same reason - to prevent such situations - we prohibit the use of `group_id` in `product_status` (see [6.10 Vulnerability – Product Statuses](https://docs.oasis-open.org/csaf/csaf-cvrf/v1.2/cs01/csaf-cvrf-v1.2-cs01.html#_Toc493508733)).

> Likewise, a Status child container cannot be tied to a Product Group due to the fact that a single product can be a member of more than one product group. 

If we need different names for one product I propose to change the specification of `full_product_name_t` to add an array with `aliases`.

sthagen:
I suggest we agree on that proposal during the meeting today - I have no stakes in one or the other. I just try to avoid a situation where we miss users stating use cases that may historically have led to the current (version 1.2) structure. If the TC is good with the change I will support it. Thanks to all for the interesting discussion.

sthagen (approved):

tschmidtb51:
> I suggest we agree on that proposal during the meeting today - I have no stakes in one or the other. I just try to avoid a situation where we miss users stating use cases that may historically have led to the current (version 1.2) structure.

Perfect. We don't want to miss them either.

tschmidtb51 (approved):

santosomar:
During the TC meeting in Sep 2, 2020; it was decided to allow 2-3 more weeks for additional comments on this issue/pull request from additional TC participants. If no comments or objections are provided, we will merge the pull request, as stated.

santosomar:
This pull request was reviewed and approved during the Sept 30, 2020 TC meeting.

Issue #109, opened 2020-08-30
Add title and description to wrapper types


sthagen:
CSAF (and CVRF v1.2) provide the document structure via semantic segmentation - so we should be able to add purpose not only to the names but also a title and a description. Currently we have e.g. for vulnerability sections:
```json
    "vulnerabilities": {
      "type": "array",
      "minItems": 1,
      "items": {
        "title": "Vulnerability",
        "description": "Is a container for the aggregation of all fields that are related to a single vulnerability in the document.",
        "type": "object",
        "minProperties": 1,
        "properties": {
          "acknowledgments": {
            "type": "array",
            "minItems": 1,
            "items": {
              "$ref": "#/definitions/acknowledgment_t"

```

Proposal:
> Add a title and description from prose transformation of the schema element.
> In this case add:
> ```json
> "title": "Vulnerabilities",
> "description": "Represents a list of all relevant vulnerability information items.",
> ```

Rationale:
> Ease future editorial cycles and maintain consistency between schema and prose.

PR #130, opened 2020-09-25
Refactoring of scores

### LINKED ISSUE #129: Refactoring of scores
The current version of  `scores` doesn't reflect the relationship between the list of products and (CVSS-) scores correctly. I suggest the following changes:
- define an object which contains: "products", "cvss_v20", "cvss_v30", "cvss_v31"
- make "products" required (This reduces complexity: An implicit score applies for all products is not longer allowed - there is explicitly specified to which products this score belongs.)
- set "minItems" as 2 ("products" and one score)
- add an editorial remark like: 'Should use "cvss_v31". "cvss_v20" will be deprecated in CSAF 2.1/ further versions'

tschmidtb51:
- resolves #129, addresses #109
- refactor "scores"


mprpic:
@tschmidtb51 I wonder if it's necessary to provide the minor version number for the CVSS properties. There are no differences in schema when you compare CVSS v3.0 and v3.1 and for v2 there were no other minor versions than v2.0. For CVSS v3, the version string is also contained within the vector string as well as defined as a separate property. If the schema only defines the major version, it won't have to be updated each time there is a new v3 minor version released.

More specifically, the `"cvss_v30"` property could be renamed to just `"cvss_v3"` and point to both the v3.1 and v3.0 schemas with the use of `"oneOf"`. The minor version used would be specified within the CVSS object itself.

tschmidtb51:
@mprpic I like the idea. However, the statement
> If the schema only defines the major version, it won't have to be updated each time there is a new v3 minor version released.

is not true. We still need to update the schema to support a new minor version but the consumer doesn't have to change his parser (if there are no major changes.).

mprpic:
> We still need to update the schema to support a new minor version but the consumer doesn't have to change his parser (if there are no major changes.).

True, I guess the additional schema would still have to be added as one of the "oneOf" values. But yes, it would still save parsers/generators from having to be updated to parse a new minor version property.

sthagen:
> More specifically, the `"cvss_v30"` property could be renamed to just `"cvss_v3"` and point to both the v3.1 and v3.0 schemas with the use of `"oneOf"`. The minor version used would be specified within the CVSS object itself.
> -- @mprpic 

From an interface perspective I am +0 on the consolidation proposal from @mprpic as I do not know any massive data processors of CSAF documents. When we explicitly tag the version via a string code, the parsing may be simpler.

Looking at the proposal from the angle of use cases (greetings from my ivory tower 🗼 ) I am -0 on the proposal of providing `oneOf`  containers for minor version alternatives. I imagine some vendor may push updated SAs with CVSS v3.1 and CVSS v3.2 (assuming a metric was added) to empower consumers of the document to pick the scoring system version that matches their existing severity/risk grading procedures in place.

Makes sense? We should discuss this further IMO.

In any case I think we should treat product IDs not like scores.

Having some shared real world samples of SAs would go a long way assisting the editors to provide the optimal schema ...

tschmidtb51:
@sthagen Agreed. Multiple containers may help companies in transition between different minor versions.

mprpic:
@sthagen Speaking from the perspective of a developer that is implementing CVRF (and future CSAF) generators, parsing the version from the CVSS object to figure out the parent-level property name feels a bit less clean I guess :-)

Also, the changes between 3.0 and 3.1 are purely in documentation and some minor changes to the math formulas (rounding was adjusted in 3.1 I think). I don't see anyone out there actually publishing scores for both 3.0 and 3.1 at the same time. Breaking changes in the schema, like the addition of a new metric or a change in any metric values, are reserved for major version updates.

All in all, I'm also fine keeping this as. My suggestion was purely from a convenience standpoint; nothing is actually wrong with the current way scores are represented in the schema. I'm happy to discuss this further.

santosomar:
This item is now in the agenda for the TC meeting tomorrow.

santosomar (approved):

Issue #138, opened 2020-09-30
Rename CSAF properties from description to details


sthagen:
JSON schema mechanics use description as meta information on the weaved in CSAF properties,

We have some same name properties in our CSAF language. One with at least 8 occrurences is description - which for itself is not very specific. Proposal is to replace these occurences to make the CSAF vs. JSON Schema terms separation more clear.

## We have (triplets of TO_BE_REWORDED, title, description) currently:

```javascript
// 1/8: ...
"description": {
    "title": "Description of the acknowledgment",
     "description": "SHOULD represent any contextual details the document producers wish to make known about the acknowledgment or acknowledged parties",
// 2/8: ...
"description": {
    "title": "Description of reference",
    "description": "Indicates what this reference refers to.",
// 3/8: ...
"description": {
    "title": "Description of the revision",
    "description": "Holds a single non-empty string representing a short description of the changes.",
// 4/8: ...
"description": {
    "title": "Description of the product group",
    "description": "Gives a short, optional description of the group.",
// 5/8: ...
"description": {
    "title": "Description of involvement",
    "description": "Contains additional context regarding what is going on.",
// 6/8: ...
"description": {
    "title": "Description of the remediation",
    "description": "Contains a thorough human-readable discussion of the remediation.",
// 7/8: ...
"description": {
    "title": "Additional restart information",
    "description": "Provides additional information for the restart. This can include details on procedures, scope or impact.",
// 8/8: ...
"description": {
    "title": "Description of the threat",
    "description": "Represents a thorough human-readable discussion of the threat.",
```

### TLDR summary proposed changes in pseudo path notation
1. `/acknowledgment_t/description` --&gt; `/acknowledgment_t/details`
2. `/references_t/[]/description`  --&gt; `/references_t/[]/details`
3. `/document/tracking/revision_history/[]/description`  --&gt; `/document/tracking/revision_history/[]/details`
4. `/product_tree/product_groups/[]/description`  --&gt; `/product_tree/product_groups/[]/details`
5. `/vulnerabilities/[]/description`  --&gt; `/vulnerabilities/[]/details`
6. `/vulnerabilities/[]/remediations/description`  --&gt; `/vulnerabilities/[]/remediations/discussion`
7. `/vulnerabilities/[]/remediations/restart_required/description`  --&gt; `/vulnerabilities/[]/remediations/restart_required/details`
[...]

PR #141, opened 2020-10-09
Rename properties "description"

tschmidtb51:
- resolves oasis-tcs/csaf#138
- rename "description" to "details" or "discussion"
- fix examples

### DIFF
--- a/csaf_2.0/examples/CVE-2018-0171-modified.json
+++ b/csaf_2.0/examples/CVE-2018-0171-modified.json
@@ -16,37 +16,37 @@
         {
           "number": "1.0",
           "date": "2018-03-28T15:17:05Z",
-          "details": "Initial public release."
+          "summary": "Initial public release."
         },
         {
           "number": "1.1",
           "date": "2018-03-29T17:13:23Z",
-          "details": "Added the researcher's company name."
+          "summary": "Added the researcher's company name."
         },
         {
           "number": "1.2",
           "date": "2018-04-02T13:18:01Z",
-          "details": "Metadata update."
+          "summary": "Metadata update."
         },
         {
           "number": "1.3",
           "date": "2018-04-06T19:35:44Z",
-          "details": "Added more details to the Workarounds section."
+          "summary": "Added more details to the Workarounds section."
         },
         {
           "number": "1.4",
           "date": "2018-04-09T14:20:12Z",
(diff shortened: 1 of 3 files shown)


sthagen (approved):
We may want to ask the members of the TC, if in some cases even `details` should give way to `summary` for being more adequate.

santosomar:
We can send it via email before the next meeting.

santosomar (approved):

mprpic:
This is more of a nitpick, but the word "discussion" to me implies opinions from multiple parties but in most cases a CSAF document will be produced by a single entity. A discussion where only one person speaks is more of a summary :-)

:+1: on the rest of the changes proposed here.

sthagen:
@mprpic summary is a good resolution for me - I even lean on naming most of these summary, as details has very loose semantics - not clear if those are noise or contribution. What do you think?

tschmidtb51:
> This is more of a nitpick, but the word "discussion" to me implies opinions from multiple parties but in most cases a CSAF document will be produced by a single entity. A discussion where only one person speaks is more of a summary :-)

I'm a little bit worried that if we use the word "summary" we will loose the details (as a summary usually skips the details and provides only an overview). IMHO this doesn't reflect the content we had in mind. For me, a "discussion" can also be done by one person if he/she takes all different aspects of the topic in to consideration and provides a chain of reasoning for the conclusion.

However, I also agree with @sthagen:
> details has very loose semantics

So I hope that the description of the field will provide enough guidance to the author to make it clear what and in which level of detail information should be provided.

tschmidtb51:
There is one point where I would prefer "summary" over "details":
```/document/tracking/revision_history/[]/description```

We don't really need all the details (where did the author change a typo or which whitespace was added) but only the fact that typos were corrected (and there were no changes to the semantic / pragmatics).

santosomar:
Let's discuss this one in the meeting tomorrow. Thank you for all your time and help on this folks!

sthagen:
Discussed and resolved in meeting 2020-10-28 as documented in https://github.com/oasis-tcs/csaf/issues/138#issuecomment-718105908 - @tschmidtb51 volunteered to amend this PR then a maintainer (@santosomar or @sthagen) will merge. Thank you to the whole TC for the dicussion and effort to resolve the issue #138

Issue #143, opened 2020-10-21
Refactoring of Acknowledgement


tschmidtb51:
In CSAF 2.0, we define twice
```
        "acknowledgments": {
          "title": "List of acknowledgements",
          "description": "Contains a list of acknowledgement elements.",
          "type": "array",
          "minItems": 1,
          "items": {
            "$ref": "#/definitions/acknowledgment_t"
          }
```

According to the DRY principle, this should be integrated into the type.

PR #144, opened 2020-10-21
Refactoring of acknowledgement

tschmidtb51:
- resolves oasis-tcs/csaf#143
- extract array defintion into the type
- rename type to "acknowledgments_t"


santosomar (approved):
Inline with other refactoring efforts and renaming it to plural makes sense.

Issue #145, opened 2020-10-21
Refactoring of branches


tschmidtb51:
In CSAF 2.0, we define twice

        "branches": {
          "title": "List of branches",
          "description": "Contains branch elements as children of the current element.",
          "type": "array",
          "minItems": 1,
          "items": {
            "$ref": "#/definitions/branch_branches_t"
          }

According to the DRY principle, this should be integrated into the type.

PR #146, opened 2020-10-21
Refactoring of branches

tschmidtb51:
- resolves oasis-tcs/csaf#145
- extract array definition into type section
- rename type "branch_branches_t" to "branches_t"


santosomar (approved):

santosomar:
I agree, this should be integrated into the type. It should also be called just "branches".

Issue #137, opened 2020-09-30
Fix copy pasta in JSON Text Sequences reference entry


sthagen:
@sthagen fix the leftovers from copy pasta in 

`N. Williams., "JavaScript Object Notation (JSON) Text Sequences", RFC 8259, DOI 10.17487/RFC7464, February 2015, http://www.rfc-editor.org/info/rfc7464.`

to become

`N. Williams., "JavaScript Object Notation (JSON) Text Sequences", RFC 7464, DOI 10.17487/RFC7464, February 2015, http://www.rfc-editor.org/info/rfc7464.`

PR #148, opened 2020-10-27
Correct JSON Text Sequences reference entry

tschmidtb51:
- resolves oasis-tcs/csaf#137
- correct RFC number in reference


sthagen (approved):

sthagen:
This was editorial change correcting a reference error - so I merged it already after review.

Issue #142, opened 2020-10-21
Acknowledg(e)ment spelling


tschmidtb51:
Currently, we use ```acknowledgment``` and  ```acknowledgement``` in the standard. I'm not sure which one we should prefer (@all native speakers: Are there differences in the meaning or use?) but I think we should only use one spelling. In CVRF 1.2, "acknowledgment*" is used 44 time, "acknowledgement*" 90 times.

tschmidtb51:
@santosomar , @mprpic: What do you prefer? Why?

santosomar:
`acknowledgement` is preferred in British English, `acknowledgment` in American English. I do not have any personal preferences. Although, I tend to use `acknowledgement` (even if I am in the U.S.).  I also looked at RFCs (like the good old [TCP RFC 793](https://tools.ietf.org/html/rfc793) ) and other documentation. The most prevalent there is `acknowledgment`.

mprpic:
I prefer `acknowledgment` since that's the US English spelling. But I think both forms are used interchangeably across the industry. I just noticed our CVE pages use the British spelling, for example: https://access.redhat.com/security/cve/cve-2018-1111 (though that may be a relic of the Red Hat Product Security team being started by a person from the UK almost 20 years ago :-)).

In any case, if I had to choose, I'd go with `acknowledgment`.

PR #149, opened 2020-10-28
Acknowledgment

tschmidtb51:
- resolves oasis-tcs/csaf#142
- rename occurance of "acknowledgement" into "acknowledgment" according to TC decision


santosomar:
This pull request addresses the agreement of using `acknowledgment` vs. `acknowledgement`.

PR #159, opened 2021-01-13
Editorial changes to Scores

tschmidtb51:
- addresses discussion and TC decision from #129
 - update prose with new structure from schema
 - correct prose to reflect current stage of schema
 - correct formatting


santosomar (approved):
Discussed during the TC meeting on January 13, 2021. The proposed change makes complete sense. Approving and merging the pull request.

Issue #162, opened 2021-01-13
Remove double CSAF


tschmidtb51:
I suggest to remove the "CSAF" in front of "Common Security Advisory Framework" as it seems to be redundant. However, when we name the CVRF 1.2 specification we should use "CSAF Common Vulnerability Reporting Framework (CVRF) Version 1.2" as this is the official name according to the [standard](https://docs.oasis-open.org/csaf/csaf-cvrf/v1.2/cs01/csaf-cvrf-v1.2-cs01.html).

PR #163, opened 2021-01-13
Editorial double csaf

tschmidtb51:
- resolves #162 
- remove CSAF in front of Common Security Advisory Framework
- add missing CSAF in front of Common Vulnerability Reporting Framework
- depend on #161

PR #166, opened 2021-01-14
Remove double CSAF

tschmidtb51:
- addresses issue related to oasis-tcs/csaf#162
- remove double CSAF in front of Common Security Advisory Framework: change was lost during merge 60406f6
- fix version number


tschmidtb51:
@santosomar: Unfortunately, some changes from #163 were lost during merge. Here is the PR to add them again

PR #167, opened 2021-01-14
Editorial Restart required

tschmidtb51:
- addresses prose related to oasis-tcs/csaf#25, oasis-tcs/csaf#35
- add description when to use which value from PR https://github.com/oasis-tcs/csaf/pull/35#issuecomment-639756156


tschmidtb51:
@sthagen: Can you please check whether the form is correct for the standard?

Issue #91, opened 2020-06-27
CPE regex


tschmidtb51:
The schema is missing a full regular expression to enforce CPE syntax. This issue should track suggestions and develop one / check existing ones.
A starting point could be figures 6-1 to 6-3 from [Common Platform Enumeration: Naming Specification Version 2.3](https://nvlpubs.nist.gov/nistpubs/Legacy/IR/nistir7695.pdf).

sthagen:
We might profit from harvesting some CPR samples in the wild. I would start with the very generous regex (case insensitive):
```javascript
          "pattern": "^(?i)cpe:(/|\\d+\\.\\d+)[^:]*:?[^:]*:?[^:]*:?[^:]*:?[^:]*:?[^:]*:?[^:]*$",
```
The above pattern supports both serialization forms of CPE WFN: URI schema and String Format Binding.

Two additional actions jump into my line of sight (besides resolvng or updating the `$comment` member value):

1. We should update the `minLen` value to the minimum length our final regex will let pass
2. We should look out for validators that choke on the case insensitive "prefix" ...

tolim:
A few comments, as I needed to get into the topic myself:

1. We need to rely on a reduced syntax as defined in Javascript (ECMA-262, https://json-schema.org/understanding-json-schema/reference/regular_expressions.html), so your prefix should not be used.
2. Unfortunately CVRF 1.2 does not define which naming scheme is allowed, so we would need to cover all three: 
    - WFN (`wfn:[part="a",vendor="microsoft",product="internet_explorer",
version="8\.0\.6001",update="beta"]`)
     - URI binding (`cpe:/a:microsoft:internet_explorer:8.0.6001:beta`), or 
     - formatted string binding (`cpe:2.3:a:microsoft:internet_explorer:8.0.6001:beta:*:*:*:*:*:*`)
3. The suggested pattern does not match the formatted string binding example above, as it has more `:` as expected,
4. It gives almost complete freedom on WFN as no `:` are contained in this naming scheme.
5. It is very performance-intensive due to the many possible branches during matching.
6. MITRE already gives us an "official" regex for formatted string binding (https://csrc.nist.gov/schema/cpe/2.3/cpe-naming_2.3.xsd): 
```
[...]

PR #171, opened 2021-01-27
CPE regex

tschmidtb51:
- addresses parts of #91
- take [regex](https://csrc.nist.gov/schema/cpe/2.3/cpe-naming_2.3.xsd) and [specification](https://nvlpubs.nist.gov/nistpubs/Legacy/IR/nistir7695.pdf) from NIST and adopt it
- fix the pattern to be valid JSON

### DIFF
--- a/csaf_2.0/json_schema/csaf_json_schema.json
+++ b/csaf_2.0/json_schema/csaf_json_schema.json
@@ -160,7 +160,7 @@
           "description": "The Common Platform Enumeration (CPE) attribute refers to a method for naming platforms external to this specification.",
           "type": "string",
           "pattern": "^cpe:2\\.3:[aho\\*\\-](:(((\\?*|\\*?)([a-zA-Z0-9\\-\\._]|(\\\\[\\\\\\*\\?!\"#\\$%&'\\(\\)\\+,/:;<=>@\\[\\]\\^`\\{\\|\\}~]))+(\\?*|\\*?))|[\\*\\-])){5}(:(([a-zA-Z]{2,3}(-([a-zA-Z]{2}|[0-9]{3}))?)|[\\*\\-]))(:(((\\?*|\\*?)([a-zA-Z0-9\\-\\._]|(\\\\[\\\\\\*\\?!\"#\\$%&'\\(\\)\\+,/:;<=>@\\[\\]\\^`\\{\\|\\}~]))+(\\?*|\\*?))|[\\*\\-])){4}$",
-          "minLength": 5
+          "minLength": 29
         }
       }
     },
--- a/csaf_2.0/json_schema/csaf_json_schema.json
+++ b/csaf_2.0/json_schema/csaf_json_schema.json
@@ -159,7 +159,7 @@
           "title": "Common Platform Enumeration representation",
           "description": "The Common Platform Enumeration (CPE) attribute refers to a method for naming platforms external to this specification.",
           "type": "string",
-          "pattern": "^(?i)cpe:(/|\\d+\\.\\d+)[^:]*:?[^:]*:?[^:]*:?[^:]*:?[^:]*:?[^:]*:?[^:]*$",
+          "pattern": "^cpe:2\\.3:[aho\\*\\-](:(((\\?*|\\*?)([a-zA-Z0-9\\-\\._]|(\\\\[\\\\\\*\\?!\"#\\$%&'\\(\\)\\+,/:;<=>@\\[\\]\\^`\\{\\|\\}~]))+(\\?*|\\*?))|[\\*\\-])){5}(:(([a-zA-Z]{2,3}(-([a-zA-Z]{2}|[0-9]{3}))?)|[\\*\\-]))(:(((\\?*|\\*?)([a-zA-Z0-9\\-\\._]|(\\\\[\\\\\\*\\?!\"#\\$%&'\\(\\)\\+,/:;<=>@\\[\\]\\^`\\{\\|\\}~]))+(\\?*|\\*?))|[\\*\\-])){4}$",
           "minLength": 5
         }
       }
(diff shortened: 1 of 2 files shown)


sthagen (approved):

mprpic:
So the regex here restricts CPE to version 2.3. Red Hat uses CPE 2.2 exclusively and we haven't yet looked at upgrading to 2.3, which would prevent us from migrating to CSAF 2.0. Is there any reason to *not* also support 2.2? It's still a supported version of CPE as far as I can tell: https://nvd.nist.gov/products/cpe.

Fwiw, we maintain a list of all CPEs for Red Hat products here: https://www.redhat.com/security/data/metrics/cpe-dictionary.xml.

Also, blindly pasting the regex here into https://regex101.com/ with the JavaScript regex flavor selected throws these errors:

```
* The preceding token is not quantifiable
* The preceding token is not quantifiable
* The preceding token is not quantifiable
* The preceding token is not quantifiable
```

I'm not sure which regex flavor is used in the source CPE pattern in https://csrc.nist.gov/schema/cpe/2.3/cpe-naming_2.3.xsd.

sthagen:
With regex dialect we should not forget the use case inside json schema: these rules must be understood by the validator. When I implement my own validator, anything goes, but when using one common tool, this tool may implement the json schema convention only, which is a less powerful dialect (if my memory does not fail me).

mprpic:
JSON schema defines the regex flavor to be used as:

> ECMA 262 regular expression dialect

https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.6.3.3

So the regex here should conform to that.

bentolor:
I tried the examples in the spec, but it seems that _Example 2_ (`cpe:2.3:a:hp:insight_diagnostics:7.4.*.1570:*:*:*:*:*:*`) is not valid according the the regex introduced by this merge request.

Here is my simple test code applied:

```javascript
var fs = require('fs')
var obj = JSON.parse(fs.readFileSync('csaf_json_schema.json', 'utf8'))

var pattern = obj.definitions.full_product_name_t.properties.cpe.pattern
var r = new RegExp(pattern)

var cpeStr = [
    'cpe:2.3:a:microsoft:internet_explorer:8.0.6001:beta:*:*:*:*:*:*',
    'cpe:2.3:a:hp:insight_diagnostics:7.4.*.1570:*:*:*:*:*:*',
    'cpe:2.3:a:microsoft:internet_explorer:8.*:sp?:*:*:*:*:*:*',
    'cpe:2.3:a:foo\\\\bar:big\\$money:2010:*:*:*:special:ipod_touch:80gb:*',
]

cpeStr.forEach(element => {
console.log(element)
console.log(r.exec(element) != null)    
});
```

mprpic:
Why was this merged? I don't think the question of whether CPE 2.2 is no longer supported was answered. Is it intentional that CSAF 2.0 will only support CPE 2.3?

sthagen:
We can use the revert button :wink: ... I think the often loose semantics of CPE in product identification suggest us to be more accepting also for CPE 2.2. Besides that, I do not think that the market that never fully implemented CPE 2.2 in the first place will eagerly push for version 2.3. 

I move we revert this merge and embrace both CPE versions. This should be a good signal for producers and consumers alike.

tschmidtb51:
> I tried the examples in the spec, but it seems that _Example 2_ (`cpe:2.3:a:hp:insight_diagnostics:7.4.*.1570:*:*:*:*:*:*`) is not valid according the the regex introduced by this merge request.
> 
> Here is my simple test code applied. Maybe i overlooked something?

Thanks for looking into that. I checked the spec again and found (right below the sentence):

> This raises an error during unbinding, due to the embedded unquoted asterisk in the version attribute.

So, the asterisk should be quoted: `cpe:2.3:a:hp:insight_diagnostics:7.4.\\*.1570:*:*:*:*:*:*` However, this comes still back as false... I took a moment and split the string:
|ABNF| _Example 2 (corrected)_|
|---|---|
| cpe:2.3: | cpe:2.3:|
| part |a|
| ":" | : |
|vendor |hp|
|":"| : |
|product |insight_diagnostics|
|":"| : |
|version | 7.4.\\*.1570|
|":"| : |
| update | * |
|":"| : |
|edition | * |
|":"| : |
|lang| * |
|":"| : |
|sw_edition | * |
|":"| : |
|target_sw | * |
|":"| : |
|target_hw | * |
|":"|  |
|other |  |

So there is one group (`":" other `) missing. Therefore, it comes back false. The correct one is:
`cpe:2.3:a:hp:insight_diagnostics:7.4.\\*.1570:*:*:*:*:*:*:*`

tschmidtb51:
The following script shows the difference:
```
var fs = require('fs')
var obj = JSON.parse(fs.readFileSync('csaf_json_schema.json', 'utf8'))

var pattern = obj.definitions.full_product_name_t.properties.cpe.pattern
var r = new RegExp(pattern)

var cpeStr = [
    'cpe:2.3:a:hp:insight_diagnostics:7.4.*.1570:*:*:*:*:*:*',
    'cpe:2.3:a:hp:insight_diagnostics:7.4.\\*.1570:*:*:*:*:*:*',
    'cpe:2.3:a:hp:insight_diagnostics:7.4.\\*.1570:*:*:*:*:*:*:*'
]

cpeStr.forEach(element => {
	console.log(element)
	console.log(r.exec(element) != null)    
});
```

Here is the output for your convenience:
```
cpe:2.3:a:hp:insight_diagnostics:7.4.*.1570:*:*:*:*:*:*
false
cpe:2.3:a:hp:insight_diagnostics:7.4.\*.1570:*:*:*:*:*:*
false
cpe:2.3:a:hp:insight_diagnostics:7.4.\*.1570:*:*:*:*:*:*:*
true
```

tschmidtb51:
> Why was this merged? I don't think the question of whether CPE 2.2 is no longer supported was answered. Is it intentional that CSAF 2.0 will only support CPE 2.3?

@mprpic: This was probably my bad. I didn't mark it as a Draft PR when you raised the question. Please see the Draft PR #177 and have a look whether it works.

Issue #126, opened 2020-09-09
Create proposal to add purl, and swid properties to full_product_name_t


sthagen:
Given robust product identification is an important factor for consumers of Security Advisories (SAs) this proposal shall offer other optional properties to the Full Product Name type (`full_product_name_t`) side by side with Common Product Enumeration CPE (`cpe`).

The inputs are the Software Package Data Exchange® ([SPDX](https://spdx.github.io/spdx-spec/)®) and [CycloneDX](https://cyclonedx.org) specifications in their latest published versions.

## Rationale:
CPE is not the only way to identify products. In CycloneDX it is marked as deprecated. 
While the communities seem to split their focus, no one seems to really want to keep CPE.
* some to SWID - ISO-IEC 19770-2 Software Identification (SWID) Tags [spec is paywalled :-1:]
* others to PackageURL (purl - https://github.com/package-url/purl-spec) CycloneDX and SPDX support this already
  * https://spdx.github.io/spdx-spec/appendix-VI-external-repository-identifiers/#purl (ibid also interesting [cpe23](https://spdx.github.io/spdx-spec/appendix-VI-external-repository-identifiers/#cpe23type) regex pattern :wink:)
  * https://cyclonedx.org/docs/1.2/#type_component (XML)
  * https://cyclonedx.org/schema/bom-1.2.schema.json (JSON)
* ...

This is IMO a dangerous situation for CSAF, as we have not many stakes in the discussions. 

So, I suggest we support the main contenders. This way, the sharing communities of SAs have more options than tunneling data of other provenience say through a cpe field.

We had this unfortunate stuation of tunneling already in the past: In the wild producers placing CVSS3 as payload to CVSS2 fields ...

mprpic:
Adding support for purl should be relatively easy since the entire purl string can be represented as a single string, so a hypothetical purl entry may look very similar to the [CPE entry](https://github.com/oasis-tcs/csaf/blob/master/csaf_2.0/json_schema/csaf_json_schema.json#L151-L158) and validate against a generic URL pattern since all purls must also be valid URLs.

[...]

PR #184, opened 2021-02-24
Product identification helper

tschmidtb51:
@sthagen I just started to implement the feature to be able to vote on this today. It doesn't fully support #126 yet.

### DIFF
--- a/csaf_2.0/json_schema/csaf_json_schema.json
+++ b/csaf_2.0/json_schema/csaf_json_schema.json
@@ -167,6 +167,14 @@
               "type": "string",
               "pattern": "^(cpe:2\\.3:[aho\\*\\-](:(((\\?*|\\*?)([a-zA-Z0-9\\-\\._]|(\\\\[\\\\\\*\\?!\"#\\$%&'\\(\\)\\+,/:;<=>@\\[\\]\\^`\\{\\|\\}~]))+(\\?*|\\*?))|[\\*\\-])){5}(:(([a-zA-Z]{2,3}(-([a-zA-Z]{2}|[0-9]{3}))?)|[\\*\\-]))(:(((\\?*|\\*?)([a-zA-Z0-9\\-\\._]|(\\\\[\\\\\\*\\?!\"#\\$%&'\\(\\)\\+,/:;<=>@\\[\\]\\^`\\{\\|\\}~]))+(\\?*|\\*?))|[\\*\\-])){4})|([c][pP][eE]:/[AHOaho]?(:[A-Za-z0-9\\._\\-~%]*){0,6})$",
               "minLength": 5
+            },
+            "purl": {
+              "$comment": "TODO: This should have a full regular expression to enforce purl syntax.",
+              "title": "Common Platform Enumeration representation",
+              "description": "N/A",
+              "type": "string",
+              "pattern": "^pkg:",
+              "minLength": 4
             }
           }
         }
--- a/csaf_2.0/json_schema/csaf_json_schema.json
+++ b/csaf_2.0/json_schema/csaf_json_schema.json
@@ -168,9 +168,64 @@
               "pattern": "^(cpe:2\\.3:[aho\\*\\-](:(((\\?*|\\*?)([a-zA-Z0-9\\-\\._]|(\\\\[\\\\\\*\\?!\"#\\$%&'\\(\\)\\+,/:;<=>@\\[\\]\\^`\\{\\|\\}~]))+(\\?*|\\*?))|[\\*\\-])){5}(:(([a-zA-Z]{2,3}(-([a-zA-Z]{2}|[0-9]{3}))?)|[\\*\\-]))(:(((\\?*|\\*?)([a-zA-Z0-9\\-\\._]|(\\\\[\\\\\\*\\?!\"#\\$%&'\\(\\)\\+,/:;<=>@\\[\\]\\^`\\{\\|\\}~]))+(\\?*|\\*?))|[\\*\\-])){4})|([c][pP][eE]:/[AHOaho]?(:[A-Za-z0-9\\._\\-~%]*){0,6})$",
               "minLength": 5
             },
+            "hashes": {
+              "$comment": "TODO: Array of objects (algorithm, value, optional name)",
+              "title": "List of hashes",
+              "description": "Contains a list of cryptographic hashes usable to identify files.",
+              "type": "array",
+              "minItems": 1,
+              "items": {
+                "title": "Cryptographic hash",
+                "description": "Contains all information to identify a file based on its cryptographic hash value.",
+                "type": "object",
+                "required": [
+                  "algorithm",
+                  "file",
+                  "value"
+                ],
+                "properties": {
+                  "algorithm": {
+                    "title": "Algorithm of the cryptographic hash",
+                    "description": "Contains the name of the cryptographic hash algorithm used to calculate the value.",
+                    "type": "string",
+                    "enum": [
+                      "SHA-256",
+                      "SHA-384",
+                      "SHA-512",
+                      "SHA-3",
+                      "BLAKE3"
+                    ]
+                  },
+                  "file": {
+                    "title": "Filename",
+                    "description": "Contains the filename of the file which is identified by the hash value.",
+                    "type": "string",
+                    "minLength": 1,
+                    "examples": [
+                      "WINWORD.EXE",
+                      "msotadddin.dll",
+                      "sudoers.so"
+                    ]
+                  },
+                  "value": {
+                    "title": "Value of the cryptographic hash",
+                    "description": "Contains the cryptographic hash value.",
+                    "type": "string",
+                    "minLength": 64,
+                    "pattern": "^([0-9a-fA-F]){64,}$",
+                    "examples": [
+                      "4775203615d9534a8bfca96a93dc8b461a489f69124a130d786b42204f3341cc",
+                      "9ea4c8200113d49d26505da0e02e2f49055dc078d1ad7a419b32e291c7afebbb84badfbd46dec42883bea0b2a1fa697c",
+                      "37df33cb7464da5c7f077f4d56a32bc84987ec1d85b234537c1c1a4d4fc8d09dc29e2e762cb5203677bf849a2855a0283710f1f5fe1d6ce8d5ac85c645d0fcb3"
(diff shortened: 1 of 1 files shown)


santosomar:
I have added this to today's meeting agenda.

sthagen:
@tschmidtb51 very good. Thanks.

sthagen (approved):
Looks good to me already.

tschmidtb51:
I will add also the `serial_number` but skip the suggestion for `sbom`.

sthagen:
... and please also the generic URI for vendor specific or future third party standard referencing extensible mapping

sthagen (approved):
Should work.

tschmidtb51:
@santosomar: I suggest to merge it as a first part. Then people can work with the updated JSON schema already. We will resolve the TODOs and address the prose in an extra PR.

csaf_2.0/json_schema/csaf_json_schema.json
```diff
@@ -154,13 +154,127 @@
...
+                      "msotadddin.dll",
+                      "sudoers.so"
+                    ]
+                  },
+                  "value": {
+                    "title": "Value of the cryptographic hash",
+                    "description": "Contains the cryptographic hash value.",
+                    "type": "string",
+                    "minLength": 64,
+                    "pattern": "^([0-9a-fA-F]){64,}$",
```
mprpic, line 215: The capturing group is not really necessary here in my opinion:

```suggestion
                    "pattern": "^[0-9a-fA-F]{64,}$",
```

csaf_2.0/json_schema/csaf_json_schema.json
```diff
@@ -226,7 +226,7 @@
             "purl": {
               "$comment": "TODO: This should have a full regular expression to enforce purl syntax.",
               "title": "package URL representation",
-              "description": "N/A",
+              "description": "The package URL (purl) attribute refers to a method for reliably identifying and locating software packages external to this specification.",
```
sthagen, line 229: :+1:

csaf_2.0/json_schema/csaf_json_schema.json
```diff
@@ -154,13 +154,127 @@
...
+                      "9ea4c8200113d49d26505da0e02e2f49055dc078d1ad7a419b32e291c7afebbb84badfbd46dec42883bea0b2a1fa697c",
+                      "37df33cb7464da5c7f077f4d56a32bc84987ec1d85b234537c1c1a4d4fc8d09dc29e2e762cb5203677bf849a2855a0283710f1f5fe1d6ce8d5ac85c645d0fcb3"
+                    ]
+                  }
+                }
+              }
+
+            },
+            "purl": {
+              "$comment": "TODO: This should have a full regular expression to enforce purl syntax.",
```
mprpic, line 227: Purls are valid URLs so any regex we include here will be a generic URL regex with the restriction on the protocol part, i.e. `^pkg:`. I probably wouldn't bother with a URL regex and leave it as is, i.e. `"pattern": "^pkg:",`.

csaf_2.0/json_schema/csaf_json_schema.json
```diff
@@ -154,13 +154,127 @@
...
+              "type": "array",
+              "minItems": 1,
+              "items": {
+                "title": "Serial number",
+                "description": "Contains a part or a full serial number of the component to identify.",
+                "type": "string",
+                "minLength": 1
+              }
+            },
+            "x_other_ids": {
```
mprpic, line 247: How about:

```suggestion
            "x_generic_uris": {
```

"Other" seems a bit vague, but maybe that's just personal preferences :-)

In the properties, it could also then be `namespace` and `uri`.

csaf_2.0/json_schema/csaf_json_schema.json
```diff
@@ -154,13 +154,127 @@
...
           ]
         },
-        "cpe": {
-          "$comment": "TODO: This should have a full regular expression to enforce CPE syntax.",
-          "title": "Common Platform Enumeration representation",
-          "description": "The Common Platform Enumeration (CPE) attribute refers to a method for naming platforms external to this specification.",
-          "type": "string",
-          "pattern": "^(cpe:2\\.3:[aho\\*\\-](:(((\\?*|\\*?)([a-zA-Z0-9\\-\\._]|(\\\\[\\\\\\*\\?!\"#\\$%&'\\(\\)\\+,/:;<=>@\\[\\]\\^`\\{\\|\\}~]))+(\\?*|\\*?))|[\\*\\-])){5}(:(([a-zA-Z]{2,3}(-([a-zA-Z]{2}|[0-9]{3}))?)|[\\*\\-]))(:(((\\?*|\\*?)([a-zA-Z0-9\\-\\._]|(\\\\[\\\\\\*\\?!\"#\\$%&'\\(\\)\\+,/:;<=>@\\[\\]\\^`\\{\\|\\}~]))+(\\?*|\\*?))|[\\*\\-])){4})|([c][pP][eE]:/[AHOaho]?(:[A-Za-z0-9\\._\\-~%]*){0,6})$",
-          "minLength": 5
+        "product_identification_helper": {
```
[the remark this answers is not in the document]
sthagen: I like the renaming @mprpic

csaf_2.0/json_schema/csaf_json_schema.json
```diff
@@ -154,13 +154,127 @@
...
+                  "algorithm",
+                  "file",
+                  "value"
+                ],
+                "properties": {
+                  "algorithm": {
+                    "title": "Algorithm of the cryptographic hash",
+                    "description": "Contains the name of the cryptographic hash algorithm used to calculate the value.",
+                    "type": "string",
+                    "enum": [
```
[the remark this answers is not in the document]
tschmidtb51: Actually not. I wanted to prevent the use of insecure hash algos... How often do they need to change? With this version a consumer is able to have a fixed set of strings for each software installed in his asset database and can easily compare them.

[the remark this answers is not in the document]
sthagen: I opt to go for examples (hashing algorithms we should not prescribe).

csaf_2.0/json_schema/csaf_json_schema.json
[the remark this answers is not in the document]
tschmidtb51: There is a big chance that `product_identifiers` are mistaken for `product_id` resp. the difference is not clear to the user. With the `_helper` it becomes more clear that it is a method the support the user in identifying the product.

csaf_2.0/json_schema/csaf_json_schema.json
sthagen: Me neither - as indicated by my comment - reading the README.md spec of Steve I do not feel confident to write a terse and useful regex. The parsers of these documents will do this anyway  they like (and with security aspects implemented) - our patterns are more like shapes (my view). Not a copy-paste template for implementations.

csaf_2.0/json_schema/csaf_json_schema.json
sthagen: I like the generic part and dislike other (which has that smell of not caring, whilst we are opening up for extension in the future)

csaf_2.0/json_schema/csaf_json_schema.json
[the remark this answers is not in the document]
sthagen: You both have a point. The dog sharing my flat reminds me to stop working now. I will suggest a third name tomorrow which we might all love. Helper I do not really like, and identifier seems to trump id (which is not what we intend).

csaf_2.0/json_schema/csaf_json_schema.json
tschmidtb51: If it is a `uri` then the format should also be `uri`. Correct?

csaf_2.0/json_schema/csaf_json_schema.json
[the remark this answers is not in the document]
tschmidtb51: What about a reasonable default value like "SHA-3"?

tschmidtb51:
> @santosomar: I suggest to merge it as a first part. Then people can work with the updated JSON schema already. We will resolve the TODOs and address the prose in an extra PR.

Looks like we have still a discussion ongoing - so please provide feedback but don't merge yet.

csaf_2.0/json_schema/csaf_json_schema.json
tschmidtb51: > I like the generic part and dislike other (which has that smell of not caring, whilst we are opening up for extension in the future)

I agree. Does it actually always have to be an uri?

csaf_2.0/json_schema/csaf_json_schema.json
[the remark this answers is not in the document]
mprpic: +1 for a reasonable default.

csaf_2.0/json_schema/csaf_json_schema.json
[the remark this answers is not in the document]
mprpic: Perhaps each of the properties could simply be defined on `full_product_name_t` itself, or is there a particular reason to have them grouped into this container? `product_identification_helper` isn't listed among the required properties of `full_product_name_t` so I assume it would be fine to simply mark all of its properties as optional as well.

csaf_2.0/json_schema/csaf_json_schema.json
sthagen: I would go for a URI and hope there will always sufficient automatic resolution of those across the community.
That would allow (citing from RFC 3986 section 1.1.2. Examples:
>ftp://ftp.is.co.za/rfc/rfc1808.txt
>http://www.ietf.org/rfc/rfc2396.txt
>ldap://[2001:db8::7]/c=GB?objectClass?one
>mailto:John.Doe@example.com
>news:comp.infosystems.www.servers.unix
>tel:+1-816-555-1212
>telnet://192.0.2.16:80/
>urn:oasis:names:specification:docbook:dtd:xml:4.1.2

especially URNs often did not help XML editing tools to resolve automatically in practice (on my little planet) - but nevertheless the producer of the CSAF using this field could even say: *Call jane!* (per the `tel:` scheme) :wink:

santosomar (approved):
It looks like this addresses all previous comments and discussion here.

mprpic (approved):

PR #215, opened 2021-04-12
Anonymous types

tschmidtb51:
- fixes #169
- remove headings introduced for array items
- separate array from item structure
- minor format changes

PR #216, opened 2021-04-12
Markdown

tschmidtb51:
- completes markdown fixes started with oasis-tcs/csaf#215
- resolves 3 additional comments from @mprpic


sthagen (approved):
Thank you.

PR #222, opened 2021-04-13
Improve wording

tschmidtb51:
- add change from 4951338 into the schema


mprpic (approved):

sthagen (approved):
Versions of software not the version of the advisory :+1:

PR #225, opened 2021-04-13
TC Charter

tschmidtb51:
I had some ideas what items could / should be changed in the charter.

@mprpic, @tolim, @sthagen, @santosomar: Please provide feedback.


charter/Charter.md
```diff
@@ -7,19 +7,19 @@
...
+    The TC will use CSAF CVRF 1.2 as the basis for creating OASIS Standards Track Work Products. One important consideration will be to ensure that the specification provides for sufficient interoperability to allow any consuming application to reliably process vulnerability-related remediation advisories from multiple sources without special semantic handling for each source. This includes also the provisioning, distribution and retrieving process. Knowning that there are changes which have an influence on backwards compatibility with CSAF CVRF 1.2, the TC will be attempting, where possible, to provide guidance on how to deal with these issues.
```
mprpic, line 16:
```suggestion
    The TC will use CSAF CVRF 1.2 as the basis for creating OASIS Standards Track Work Products. One important consideration will be to ensure that the specification provides for sufficient interoperability to allow any consuming application to reliably process vulnerability-related remediation advisories from multiple sources without special semantic handling for each source. This includes also the provisioning, distribution and retrieving process. Knowing that there are changes which have an influence on backwards compatibility with CSAF CVRF 1.2, the TC will be attempting, where possible, to provide guidance on how to deal with these issues.
```

mprpic (commented):
I wonder if the scope of work should also include the production of supporting tooling to parse CSAF documents, validate them, or generate them? Or do we want to purposefully leave that out to allow the larger community to produce these tools?

sthagen:
> I wonder if the scope of work should also include the production of supporting tooling to parse CSAF documents, validate them, or generate them? Or do we want to purposefully leave that out to allow the larger community to produce these tools?

I think the TC should provide encouragement for tooling to emphasize the importance of actionable advisory culture and optimal automation end to end.

For this the TC may produce tooling and I would put this as such into the charter.
... as long as we do not put target dates in ... in one TC I participated, we removed this year a „2008 ready“ clause from the charter which ...

Rationale for including: What is not in the charter should not be done, and as some want to challenge the ecosystem by pushing tools, we should include the option IMO.

I think some members already have a shopping list aligned with our conformance target roles ....

charter/Charter.md
```diff
@@ -7,19 +7,19 @@
...
+    The TC will make substantive additions and other changes to the CSAF CVRF 1.2 input specification to create and maintain the CSAF specification (Version 2.0 and later) to correct errors and evolve capabilities based on requirements and capabilities identified by OASIS TC members. The TC will rename the framework to more closely align to the primary use (e.g., Common Security Advisory Framework - CSAF). Deliverables will include a major revision of the framework. In addition to the specification deliverables, the TC may deliver supporting documentation and open source tooling on an ongoing basis in support of the TC's published standard(s). The TC expects to produce a major revision of the framework within 3 months of its next meeting after approving the charter.
```
sthagen, line 22: Well, I for one would (a) remove any temporal info from the last sentence in this paragraph, (b) not only include 2.0 (as it reads now), and (c) focus on the "why" we do that. Something like: 

>The TC expects to produce major revision of the framework to support the advisory sharing communities.

sthagen (approved):
Despite the one comment I placed this looks good to me.

charter/Charter.md
tschmidtb51: Thank you @sthagen. I added the change.

sthagen (approved):
❤️🧡💛💚💙💜

charter/Charter.md
tschmidtb51: Given that change, I extended the statement about backwards compatibility to following major revisions.

sthagen (approved):
As a German, succeeding makes me feel unsure if it is a success or a successor we mean with the term ... I would rather write `subsequent` 
I approve with this comment and am sure, our English Natives will help us here when they start their day ...

tschmidtb51:
> > I wonder if the scope of work should also include the production of supporting tooling to parse CSAF documents, validate them, or generate them? Or do we want to purposefully leave that out to allow the larger community to produce these tools?
> 
> I think the TC should provide encouragement for tooling to emphasize the importance of actionable advisory culture and optimal automation end to end.
> 
> For this the TC may produce tooling and I would put this as such into the charter.
> ... as long as we do not put target dates in ... in one TC I participated, we removed this year a „2008 ready“ clause from the charter which ...
> 
> Rationale for including: What is not in the charter should not be done, and as some want to challenge the ecosystem by pushing tools, we should include the option IMO.

I just added a paragraph regarding that.

> I think some members already have a shopping list aligned with our conformance target roles ....

I really hope that and am looking forward to seeing that happening :smile:

charter/Charter.md
tschmidtb51: > The TC expects to produce a major revision of the framework within 3 months of its next meeting after approving the charter.

This is my (personal) hope.

sthagen (approved):
Well stated - I like the challenge in there for both TC and communities - I tend to use the term audience less and less often these days. There is nothing good to be expected from active specification but  passive consumption. Vibrant communities and committees should be eager (but not forced) to produce tooling to yield the promises in real life.

mprpic (approved):
LGTM!

santosomar (commented):
My vote does not count, as chair. Although, the changes suggested are aligned with previous discussions and suggestions by TC members. This will be a core item of discussion during the TC's meeting today 2021-04-14. Thank you all!

santosomar:
This pull request was discussed and the motion to accept the changes were approved during the CSAF TC meeting on 2021-04-14.

PR #228, opened 2021-04-15
Limit acknowledgment organization to a single string value

### LINKED ISSUE #227: Proposal for acknowledgments attributes to be single values instead of arrays
Currently, the schema defines the `acknowledgments_t` type as a list of objects that themselves may contain lists of names and organizations. This may produce valid data such as:

```
acks: [
  {names: [Joe, Mary, Ann], organizations: [BigCorp, CoolStartup]},
]
```

I think the schema should enforce this to be split into multiple objects with the name and organization as single values so that the relationship between a person and their organization is explicit. Alternatively, we could limit just the `organizations` attribute to be a single value so that multiple persons affiliated with the same organization could be specified in a single object, but that seems overly complicated in my opinion.

The above represented as single values would look like this:

```
acks: [
  {name: Joe, organization: BigCorp},
  {name: Mary, organization: CoolStartup},
  {name: Ann}
]
```

The `summary` and `urls` attributes of the acknowledgment object would remain unchanged.

mprpic:
The previous value of an array allowed for ambiguous relationships between
names, URLs or summaries to any of the organization in the array.
Limiting this to a single organization per acknowledgment object makes the
association to names, URLs and a summary clear.

Resolves #227

### DIFF
--- a/csaf_2.0/prose/csaf-v2-editor-draft.md
+++ b/csaf_2.0/prose/csaf-v2-editor-draft.md
@@ -563,6 +563,8 @@ Value type is string with format URI (`uri`).
 
 ##### 3.1.1.5 Acknowledgments Type - Example
 
+Example:
+
 ```
 "acknowledgments": [
   {
--- a/csaf_2.0/prose/csaf-v2-editor-draft.md
+++ b/csaf_2.0/prose/csaf-v2-editor-draft.md
@@ -593,6 +593,14 @@ Value type is string with format URI (`uri`).
   },
 ]
 ```
+The example above should lead to the following outcome in a human-readable advisory:
+
+> We thank the following parties for their efforts:
+> 
+> * Johann Sebastian Bach, Georg Philipp Telemann, Georg Friedrich Händel from Baroque composers for wonderful music
+> * CISA for coordination efforts (see: https://cisa.gov)
+> * BSI for assistance in coordination
+> * Antonio Vivaldi for influencing other composers
 
 ### 3.1.2 Branches Type
 
--- a/csaf_2.0/prose/csaf-v2-editor-draft.md
+++ b/csaf_2.0/prose/csaf-v2-editor-draft.md
@@ -2383,10 +2383,11 @@ Firstly, the program:
 
 Secondly, the program for all items of:
 
+* `/document/acknowledgment/organization` and `/vulernabilities[]/acknowledgment/organization`: If more than one cvrf:Organization instance is given, the CVRF CSAF converter converts the first one into the `organization`. In addition the converter outputs a warning that information might be lost during conversion of document or vulnerability acknowledgment.
 * `/document/publisher/name`: Sets the value as given in the configuration of the program or the corresponding argument the program was invoked with. If both values are present, the program should prefer the latter one.
 * `/vulernabilities[]/scores[]`: If no `product_id` is given, the CVRF CSAF converter appends all Product IDs which are listed under `../product_status` in the arrays `known_affected`, `first_affected` and `last_affected`.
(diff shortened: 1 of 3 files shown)


csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -562,6 +561,39 @@ List of URLs (`urls`) of acknowledgment is a container (value type `array`) for
...
+  },
+  {
+    "names": [
+      "Antonio Vivaldi"
+    ],
+    "summary": "influencing other composers"
+  },
+]
+```
+
```
tschmidtb51, line 596:
```suggestion

The example above should lead to the following outcome in a human-readable advisory:

This should lead in a human-readable advisory to the following outcome:

> We thank the following parties for their efforts:
> 
> * Johann Sebastian Bach, Georg Philipp Telemann, Georg Friedrich Händel from Baroque composers for wonderful music
> * CISA for coordination efforts (see: https://cisa.gov)
> * BSI for assistance in coordination
> * Antonio Vivaldi for influencing other composers

```

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -562,6 +561,39 @@ List of URLs (`urls`) of acknowledgment is a container (value type `array`) for
 Any URL of acknowledgment contains the URL or location of the reference to be acknowledged.
 Value type is string with format URI (`uri`).
 
+##### 3.1.1.5 Acknowledgments Type - Example
+
```
tschmidtb51, line 565: According to the rules in the document this should be an informative section. Therefore, we should add before the example the following:
```suggestion

Example:

```

tschmidtb51 (commented):
@mprpic: Thank you for providing the PR. We need also to add the conversion rule in the CVRF CSAF converter.

tschmidtb51:
> @mprpic: Thank you for providing the PR. We need also to add the conversion rule in the CVRF CSAF converter.

I just committed a conversion rule. Please provide feedback.

csaf_2.0/prose/csaf-v2-editor-draft.md
mprpic: Can you please fix the double leading clause here so I can just accept this suggestion? :-)

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -2351,10 +2383,11 @@ Firstly, the program:
 
 Secondly, the program for all items of:
 
+* `/document/acknowledgments[]/organization` and `/vulernabilities[]/acknowledgments[]/organization`: If more than one cvrf:Organization instance is given, the CVRF CSAF converter converts the first one into the `organization`. In addition the converter outputs a warning that information might be lost during conversion of document or vulnerability acknowledgment.
```
mprpic, line 2386:
```suggestion
* `/document/acknowledgments[]/organization` and `/vulnerabilities[]/acknowledgments[]/organization`: If more than one cvrf:Organization instance is given, the CVRF CSAF converter converts the first one into the `organization`. In addition the converter outputs a warning that information might be lost during conversion of document or vulnerability acknowledgment.
```

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: Sorry that was by accident.

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: Found the root cause which will be resolved in the next editor revision.

sthagen (approved):
LGTM.

santosomar:
Thank you, @mprpic ! We do not have to vote on this, since we discussed this issue previously. I am merging the pull request, as it is very straightfw.

PR #230, opened 2021-04-19
Rename properties "type"

### LINKED ISSUE #212: Proposal to rename type properties to category
CSAF currently exposes enumerations as type properties in some objects.

Examples are in document and in publisher as well as others. A handful.

I propose to instead name that field `category`.

Rationale: The values of these domains are categorial data. There is no ordering implied and the values are all of type string.

Also, we would gain some cognitive separation from the type fields of the core vocabulary of JSON Schema, which would help implementers and users alike IMO.

tschmidtb51:
- resolves #212
- rename "type" to "category"
- change prose and examples to match schema

### DIFF
--- a/csaf_2.0/examples/CVE-2018-0171-modified.json
+++ b/csaf_2.0/examples/CVE-2018-0171-modified.json
@@ -1,13 +1,13 @@
 {
   "document": {
     "title": "Cisco IOS and IOS XE Software Smart Install Remote Code Execution Vulnerability",
-    "type": "Cisco Security Advisory",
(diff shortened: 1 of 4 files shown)


tschmidtb51:
I hope I caught all occurrences.

@sthagen, @mprpic, @santosomar: Please review.

sthagen (approved):
Thank you. I like this very much and am very happy we could agree on naming the key to categories `category`. especially so late in the process towards version 2.0.

santosomar:
+1! Thank you so much!

santosomar (approved):

PR #232, opened 2021-04-21
Run Actions on PR

tschmidtb51:
Currently, we only run the actions during `push`. As a result, we don't see whether a PR might break the tests. Therefore, I suggest to add PRs to the triggers of actions.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,4 +1,4 @@
-on: [push]
+on: [push, pull_request]
 
 jobs:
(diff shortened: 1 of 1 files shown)


santosomar (approved):
Yes. We should definitely include pull requests. Thank you for catching this!

sthagen (commented):
LGTM

santosomar:
Thank you all! I will merge now, since the change is fairly straightforward.

Issue #111, opened 2020-08-30
Include any version system that allows meaningful ordering for the consumer


sthagen:
The current schema contains a restriction on the semantics of version identifiers that is not refelcted in the pattern and IMO is not beneficial to the producers and consumers.

```json
"version_t": {
    "title": "Version",
    "description": "[...] Format must be understood as 'major.minor.patch.build' version.",
    "type": "string",
    "pattern": "^(0|[1-9][0-9]*)(\\.(0|[1-9][0-9]*)){0,3}$"
```

Proposal:
> Rewrite description above text value as:
>```json
>"[...] Format must be interpretable as 'major[.minor[.patch[.build]]]' version, calendar version, or any other simple sequential numbering scheme."
>```
> or similarly.

Rationale:
> Any calendar version matches the ordering expectation within the security advisories sharing community.

santosomar:
This is a good suggestion and provides good guidance.

tolim:
It is important that CSAF parsers are able to determine the order of advisory releases (i.e. which is the latest one?). In your suggestion you include `or any other simple sequential numbering scheme`. How do we ensure that used algorithms will work correctly here as well? 

So I propose to change the description of `version_t` as follows:
```
"[...] Format must be interpretable as 'major[.minor[.patch[.build]]]' version. Alternatively, a calendar version or any other simple sequential numbering scheme may be used that allows ASCII sorting in the order of the document releases."
```

A small side note:
In `revision_history`, we also require the property `date` that allows to determine this order:

```
            "revision_history": {
              "type": "array",
              "minItems": 1,
              "items": {
                "type": "object",
                "properties": {
                  "number": {
                    "$ref": "#/definitions/version_t"
                  },
                  "date": {
                    "title": "Date of the revision",
                    "description": "The date of the revision entry",
                    "type": "string",
                    "format": "date-time"
                  },
                  "description": {
                    "type": "string"
                  }
                },
                "required": [
                  "number",
                  "date",
                  "description"
                ]
              }
            },
```
However, if two releases are published on the same day, we will have a problem once again. We might require people to also specify the time of the revision, but this is not needed if we provide a clear way how to sort the versions.

tschmidtb51:
I agree on the first part. The documentation should be changed to:

`"[...] Format must be interpretable as 'major[.minor[.patch[.build]]]' version.`

[...]

Issue #214, opened 2021-04-11
Allow explicit calendar versioning and provide actionable rules for version semantics


sthagen:
PR #46 nicely made the version concept actionable.

But, it leaves producers employing calendar versioning in a more explicit way outside in the rain.

I suggest to modify the version type value regex to allow calendar versioning variants that hint explicitely at that semantic:
>`^(0|[1-9][0-9]*)(\\.([0-9]+)){0,3}$`

that way a version like `21.04` Is valid - which would help producers that have set up their software development to use calendar versioning (because intrinsic semantic versioning just did not work well for them) - the results are still orderable in any packaging relevant context.

Regardless of this proposed change:
> Can we help the ecosystem / communities by distilling actionable rules as guidance from the members working in the field which classes of changes would be expected to impact which degree of change indicator? 

Given the tight coupling between the advisories and the expected reaction upon receipt, I imagine we should be able to help here - if it is only informative it will still help to foster communication between producers and consumers. Even better, when we can normatively state a SHOULD with confidence.

tschmidtb51:
I would refrain from introducing calendar versioning since we already a field `date` in the `revision_history`. Moreover, it would leave to consumer alone not knowing whether the rules from semantic versioning apply or the issuer wanted to use calendar versioning in this document. It might be easy to guess for `21.04` that this is probably a calendar versioning string, but what about `10.12`? This is also a valid semantic versioning string.

Therefore, I suggest to offer 2 versioning systems:

- semantic versioning (as the preferred one, with the rules we provide)
- integer version (just increment the number for each new `final` version)

A document must not mix both versions. Draft versions before the initial release should have a version string starting with `0.`. There should only be `final` or `interim` versions in the `revision_history` when the document is released.

tschmidtb51:
I added PR #233 as a suggestion. It is still in draft status (so that it doesn't get merged accidentally) and does not implement the changes in the JSON schema yet. I'm interested whether this is the way to go. If the answer is yes, I'm happy to make the changes to the schema and fix the examples.

sthagen:
I will take a look at the proposed wording in the new PR. If we can come up with clear and actionable rules that relate an advisory over the full life span to a semantic version, there is no need for using a calendar version.

But, if we cannot provide that regime and producers start to just fake semantics on the version field for formal validity, ...

[...]

PR #233, opened 2021-04-22
Versioning rules

tschmidtb51:
- suggests prose related to oasis-tcs/csaf#214
- introduce semantic and integer versioning

### DIFF
--- a/csaf_2.0/prose/csaf-v2-editor-draft.md
+++ b/csaf_2.0/prose/csaf-v2-editor-draft.md
@@ -2520,14 +2520,14 @@ A CSAF content management system satisfies the "CSAF content management system" 
   * show an audit log for each CSAF document
   * identify the latest version of CSAF documents with the same `/document/tracking/id`
   * suggest a `/document/tracking/id` based on the given configuration.
-  * track of the version of CSAF documents automatically and increments for each change at least the patch_version.
+  * track of the version of CSAF documents automatically and increment according to the versioning scheme (see also subsections of 3.1.11) selected in the configuration.
   * support the following workflows:
 
     * "New Advisory": create a new advisory, request a review, provide review comments or approve it, resolve review comments; if the review approved it (draft->interim), the approval for publication can be requested; if granted (manual or time-based) the advisory is provided for publication (interim -> final)
     * "Update Advisory": open an existing advisory, create new revision & change content (interim), request a review, provide review comments or approve it, resolve review comments; if the review approved it (draft->interim), the approval for publication can be requested; if granted (manual or time-based) the updated advisory is provided for publication (interim -> final)
 
 * publication may be immediately or at a given date/time.
-* handling of date/time and version is be automated.
+* handling of date/time and version is automated.
 * provide an API to retrieve all CSAF documents which are currently in the status published.
 * should provide an API to import or create new advisories from outside systems (e.g. bug tracker, CVD platform,...).
 * provide a user management and support at least the following roles:
@@ -2553,7 +2553,7 @@ A CSAF content management system satisfies the "CSAF content management system" 
   * `/document/tracking/initial_release_date` with the current date
   * `/document/tracking/revision_history`
     * `date` with the current date
-    * `number` (based on the templates from configuration; default: 0.1)
+    * `number` (based on the templates according to the versioning scheme configured)
     * `summary` (based on the templates from configuration; default: "Initial version.")
   * `/document/tracking/status` with `draft`
   * `/document/tracking/version` with the value of `number` the latest `/document/tracking/revision_history[]` element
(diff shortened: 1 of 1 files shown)


tschmidtb51:
@mprpic, @sthagen, @santosomar, @tolim, @zmanion: Please provide feedback.

This is just a draft - it was easier for me to provide that instead of explaining what I think... If we agree on that direction, I'll follow up and provide the changes in the schema and fix the examples.

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -1146,6 +1153,91 @@ Examples:
...
+9. Build metadata MAY be denoted by appending a plus sign and a series of dot separated identifiers immediately following the patch or pre-release version. Identifiers MUST comprise only ASCII alphanumerics and hyphens [0-9A-Za-z-]. Identifiers MUST NOT be empty. Build metadata MUST be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence. Examples: 1.0.0-alpha+001, 1.0.0+20130313144700, 1.0.0-beta+exp.sha.5114f85, 1.0.0+21AF26D3—-117B344092BD.
+10. Pre-release and build metadata SHOULD NOT be included if the `/document/status` is `final`.
+11. Precedence refers to how versions are compared to each other when ordered.
```
mprpic, line 1209: I'd say items 8, 9, and 11 should just refer to the semver documentation instead of being copy-pasted here. E.g.:

```markdown
8. A pre-release version of a CSAF document (document status `draft`) may be denoted by
appending a hyphen and and a series of dot separated identifiers immediately following the
patch version, as specified by the Semantic Versioning Specification (https://semver.org/#spec-item-9).
```

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -1146,6 +1153,91 @@ Examples:
...
+9. Build metadata MAY be denoted by appending a plus sign and a series of dot separated identifiers immediately following the patch or pre-release version. Identifiers MUST comprise only ASCII alphanumerics and hyphens [0-9A-Za-z-]. Identifiers MUST NOT be empty. Build metadata MUST be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence. Examples: 1.0.0-alpha+001, 1.0.0+20130313144700, 1.0.0-beta+exp.sha.5114f85, 1.0.0+21AF26D3—-117B344092BD.
+10. Pre-release and build metadata SHOULD NOT be included if the `/document/status` is `final`.
```
mprpic, line 1208: Why allow build metadata on draft but not on final? Someone may want to refer to a Git commit hash associated with the version even when the document is in its final form.

mprpic (commented):
Overall, looks good to me and I agree with the change!

csaf_2.0/prose/csaf-v2-editor-draft.md
sthagen: I am +0 on providing the copy as a service to the reader. I once started to support the GeoJSON community to make their well-working but informal  specification into an RFC so as to allow OASIS specifications to refer to them normatively. To semver.org we can only refer informally, and shall provide the content as a copy. This makes the future OASIS standard stable and the reference serves still the purpose of indicating the source (no plagiarism)

sthagen (approved):
I would keep optional fields regardless of state as pointed out by @mprpic to ease identification back to even consumer invisible sources.
Otherwise I like this change and consider it well worth to merge and have the members verify, that they can ensure proper semantics to their advisories (produced as well as consumed) across the full life span of an advisory.

Resolves #111

csaf_2.0/prose/csaf-v2-editor-draft.md
mprpic: Fair enough, if we feel it's better to include the content as is to be more future-proof I'm fine with that.

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: Thanks for the catch. I moved it from here to the `revision_history` where it should be (see 7f49db9).

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: @sthagen: Thanks for the comment. I moved [SemVer] from the "Normative References" to the "Informative Reference". Could you please check that I got the right format for OASIS?

sthagen (approved):

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: @sthagen: Thank you.

tschmidtb51:
@santosomar: Please add to the agenda for the next TC meeting.

santosomar:
@tschmidtb51 - Consider it done ;-)

sthagen (approved):

mprpic (approved):

Issue #209, opened 2021-04-09
Add remark on hash match but filename deviation


tschmidtb51:
@tlimmer stated in https://github.com/oasis-tcs/csaf/pull/199#commitcomment-49170258:

> What happens if the hash matches but the file name is incorrect?

---

@sthagen replied in https://github.com/oasis-tcs/csaf/pull/199#commitcomment-49171722:

> To me the ideal case is all-three-match. Another interesting use case IMO is where the file name information is more a comment and the consumer might act on the hash and document internally the deviation in name. Say, an admin renamed a file to not override upon update, created links, or the filesystem did name mappings (case sensitivity/or-not comes to mind and also automatic versioning like renamings as in windows and mac when the user is prompted upon placing a file with colliding name to replace an existing file or not ...

---

@tolim replied in https://github.com/oasis-tcs/csaf/pull/199#commitcomment-49181517:

> Yes, I agree. Using this method of identification, we will always be able to construct cases that will produce false positive matches.
> I would suggest that, if the hash is specified, it should be the sole criteria to decide on. Then the accompanying file name should be informative, and maybe even defined as an optional property.

---

As this is something which needs more discussion I added it as a ticket.

tschmidtb51:
I agree with @sthagen that the ideal case is all-three-match.

I suggest to add a remark that the hash value should be preferred in such situations:

```suggestion
If the value of the Value of the cryptographic hash matches and the Filename does not, a user should prefer the hash value. In such cases the filename should be used an informational property.
```

IMHO, I would refrain from making the file name optional as there is the motion (see #208) to allow weak hash algorithms. This would lead to too many false positives.

sthagen:
Editorial nit:
> ```
> If the value of the hash matches and the Filename does not, a user should prefer the hash value. In such cases the filename should be used as informational property.
> ```

Question: Could that filename property then also be used as a range identifier? Like `bytes_[123,4200]`?

sthagen:
Another idea:
> Offer an array of hash values.

Benefits: 
* producer offers either multiple algorithm values not knowing which of these the consumer prefers or is capable of using
* across versions variations of same name files exist that are vulnerable and others which are not; this way the producer can state hashes vor all known impacted files

Good idea, bad idea? Any feedback appreciated - e.g. to spawn of another low priority issue ...

tschmidtb51:
> Could that filename property then also be used as a range identifier? Like `bytes_[123,4200]`?

[...]

Issue #208, opened 2021-04-09
Change Hash default and names


tschmidtb51:
@tolim stated in https://github.com/oasis-tcs/csaf/pull/199#commitcomment-49170069:

> `SHA-3` is ambiguous, as the bit length is not specified. Should we reference the IANA Named Information Hash Algorithm Registry here? https://www.iana.org/assignments/named-information/named-information.xhtml

---

@sthagen replied in https://github.com/oasis-tcs/csaf/pull/199#commitcomment-49171473:

> Yes, I think reusing the ids from the iana registry is much better. I will propose using the bitlengths matching the „other“ sha algorithms and maybe use sha-256 as default instead, because we will not want to edufy the producers but instead offer a meaningful default matching many existing uses already.

---

@tolim stated in https://github.com/oasis-tcs/csaf/pull/199#commitcomment-49180351:

> After a little bit of thought, I suggest to do it differently: we do not require a cryptographically secure hash algorithm here. In contrast, quite many file hashes are available based on MD5 or SHA-1, and the IANA Named Information Hash Algorithm Registry does not list those algorithms. The IANA Hash Function Textual Tables (https://www.iana.org/assignments/hash-function-text-names/hash-function-text-names.xhtml) would list those, but not the newer ones.
> 
> After looking a bit around, I found the best way to get a good list of hash identifiers is to look at OpenSSL. The command `openssl dgst -list` (Version 1.1.1f from 2020-03-31) outputs the following:
> 
> ```
> Supported digests:
> -blake2b512                -blake2s256                -md4                      
> -md5                       -md5-sha1                  -ripemd                   
> -ripemd160                 -rmd160                    -sha1                     
> -sha224                    -sha256                    -sha3-224                 
> -sha3-256                  -sha3-384                  -sha3-512                 
> -sha384                    -sha512                    -sha512-224               
> -sha512-256                -shake128                  -shake256                 
> -sm3                       -ssl3-md5                  -ssl3-sha1                
> -whirlpool      
> ```
> 
> If there is not a better source available, I would suggest to use those identifiers for the moment.
> 
> As for the default value, I agree to use sha256.

---

As this is something which needs more discussion I added it as a ticket.

tschmidtb51:
Personally, I always prefer a cryptographic secure hash algorithm and I don't like setting the default low when it comes to security - therefore I suggested the `SHA-3`.

[...]

PR #239, opened 2021-04-28
Hashes

tschmidtb51:
- resolves comments from #208 and #209
- provide an array with hashes for each filename
- add remark about interpretation of filename mismatch at hash match
- change default value to `sha256`
- use digest list from OpenSSL

### DIFF
--- a/csaf_2.0/json_schema/csaf_json_schema.json
+++ b/csaf_2.0/json_schema/csaf_json_schema.json
@@ -168,50 +168,66 @@
               "type": "array",
               "minItems": 1,
               "items": {
-                "title": "Cryptographic hash",
-                "description": "Contains all information to identify a file based on its cryptographic hash value.",
+                "title": "Cryptographic hashes",
+                "description": "Contains all information to identify a file based on its cryptographic hash values.",
                 "type": "object",
                 "required": [
-                  "algorithm",
-                  "file",
-                  "value"
+                  "file_hashes",
+                  "file_name"
                 ],
                 "properties": {
-                  "algorithm": {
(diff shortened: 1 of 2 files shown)


tschmidtb51:
@sthagen: Should we add an informational reference to OpenSSL?

@native_speakers: Is it `filename` or `file name`?

@all: Do you have a better suggestion for the `title` of items in the `hashes` array?

tschmidtb51:
@santosomar: We should discuss these questions tonight.

sthagen:
Yes @tschmidtb51 I think an informative reference to openssl is in order when we suggest the use of the nomenclature from that project / toolbox

tschmidtb51:
> Yes @tschmidtb51 I think an informative reference to openssl is in order when we suggest the use of the nomenclature from that project / toolbox

Does that suit our purpose or should we refer to the project's GitHub repo?

sthagen (approved):
LGTM.

sthagen:
> Does that suit our purpose or should we refer to the project's GitHub repo?

I think it does. Github is a brand in itself. So, if a project operates an own web presence I would always prefer that as to be explicit about the diversity and the contributions of individual natural persons technology really depends upon.

sthagen (approved):

tschmidtb51:
It was decided during the TC meeting at 28.04.2021 to wait for comments for one week.

santosomar:
There are some minor conflicts in the pull request. Please review https://github.com/oasis-tcs/csaf/pull/239/conflicts and I will merge right after.

santosomar:
I have addressed the conflict and will merge shortly.

PR #240, opened 2021-04-28
Versioning (into schema & examples)

tschmidtb51:
- addresses part of #214 
- include pattern into schema
- fix pattern errors
- adopt description
- fix examples

### DIFF
--- a/csaf_2.0/examples/CVE-2018-0171-modified.json
+++ b/csaf_2.0/examples/CVE-2018-0171-modified.json
@@ -12,40 +12,40 @@
     "tracking": {
       "id": "cisco-sa-20180328-smi2",
       "status": "final",
-      "version": "1.6",
+      "version": "3.0.0",
       "revision_history": [
         {
-          "number": "1.0",
+          "number": "1.0.0",
           "date": "2018-03-28T15:17:05Z",
           "summary": "Initial public release."
         },
         {
-          "number": "1.1",
+          "number": "1.1.0",
           "date": "2018-03-29T17:13:23Z",
           "summary": "Added the researcher's company name."
         },
         {
-          "number": "1.2",
+          "number": "1.2.0",
           "date": "2018-04-02T13:18:01Z",
           "summary": "Metadata update."
         },
         {
-          "number": "1.3",
+          "number": "1.3.0",
           "date": "2018-04-06T19:35:44Z",
           "summary": "Added more details to the Workarounds section."
         },
         {
-          "number": "1.4",
+          "number": "1.4.0",
           "date": "2018-04-09T14:20:12Z",
           "summary": "Emphasized that Smart Install is enabled by default. Added a link to the list of devices that support Smart Install."
         },
         {
-          "number": "1.5",
+          "number": "2.0.0",
           "date": "2018-04-16T18:21:34Z",
           "summary": "Updated IOS Software Checker with products found to be non-vulnerable."
         },
         {
-          "number": "1.6",
+          "number": "3.0.0",
           "date": "2018-04-17T15:08:41Z",
           "summary": "Updated IOS Software Checker with products found to be vulnerable."
         }
--- a/csaf_2.0/json_schema/csaf_json_schema.json
+++ b/csaf_2.0/json_schema/csaf_json_schema.json
@@ -444,14 +444,15 @@
     },
     "version_t": {
       "title": "Version",
-      "description": "Specifies a version string with a simple hierarchical counter model to denote clearly the evolution of the content of the document. Format must be understood as 'major[.minor[.patch[.build]]]' version.",
+      "description": "Specifies a version string to denote clearly the evolution of the content of the document. Format must be either integer or semantic versioning.",
       "type": "string",
-      "pattern": "^(0|[1-9][0-9]*)(\\.(0|[1-9][0-9]*)){0,3}$",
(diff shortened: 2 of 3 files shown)


tschmidtb51:
@santosomar: Please review also the example as I changed the advisory versions in `revision_history` based on the summary given.

sthagen (approved):
Nice.

santosomar (commented):
I have not seen anyone use more than minor releases of an advisory. In other words, you will typically see 1.0 vs 1.0.0 or 2.1 vs 2.1.0. I wonder if by default we should have that granularity or the tool just assume that 1.0 is 1.0.0 if there is no third digit?

Examples: 
- Cisco: https://tools.cisco.com/security/center/content/CiscoSecurityAdvisory/cisco-sa-memc-dos-fncTyYKG
- Arista: https://www.arista.com/en/support/advisories-notices/security-advisories/12135-security-advisory-61
- Juniper: https://kb.juniper.net/InfoCenter/index?page=content&id=JSA11125&cat=SIRT_1&actp=LIST
- Palo Alto: https://security.paloaltonetworks.com/CVE-2021-3035 (it has a revision history, but no version)
- NetApp: https://security.netapp.com/advisory/ntap-20210423-0002/
- Siemens" https://security-advisory-tracker.dc.siemens.com/advisories/SSA-875726/60772e4bdc2fa6001263071c

A lot of other "advisories" also do not have any versioning information:
- Apple: https://support.apple.com/en-us/HT212321
- ASUS: https://www.asus.com/content/ASUS-Product-Security-Advisory/
- Lenovo: https://support.lenovo.com/us/en/product_security/LEN-29406
- IBM: https://www.ibm.com/support/pages/node/270463
- 

The interesting part is that some of the "newer types" of advisories are not really advisories. For example: 
- https://nvd.nist.gov/vuln/detail/CVE-2021-29251
- which is: https://github.com/btcpayserver/btcpayserver/releases/tag/v1.0.7.1

and 
- https://nvd.nist.gov/vuln/detail/CVE-2021-28164
- which is: https://github.com/eclipse/jetty.project/security/advisories/GHSA-v7ff-8wcx-gmc5

santosomar (commented):
At the very minimum, please don't change `csaf_2.0/examples/CVE-2018-0171-modified.json` since we do not do any version updates as granular as 1.2.0. We do 1.0, 1.2, 1.3, 2.0, 2.1, etc... I want to make sure that we do not  enforce vendors to have an extra semantic position (patch) when there will never be a "patch" for a document revision of an advisory. In other words:

From https://semver.org/
```
Given a version number MAJOR.MINOR.PATCH, increment the:
- MAJOR version when you make incompatible API changes,
- MINOR version when you add functionality in a backwards compatible manner, and
- PATCH version when you make backwards compatible bug fixes.
```
I will expect that no vendor will have a "patch" for a document/advisory. Vendors and producers will absolutely have these versioning for the actual software fixes/affected versions, etc... but I do not expect that they will have it for a document (as per my examples above).

mprpic (approved):
LGTM!

tschmidtb51:
@santosomar wrote:
> I have not seen anyone use more than minor releases of an advisory. In other words, you will typically see 1.0 vs 1.0.0 or 2.1 vs 2.1.0. 

> I will expect that no vendor will have a "patch" for a document/advisory. Vendors and producers will absolutely have these versioning for the actual software fixes/affected versions, etc... but I do not expect that they will have it for a document (as per my examples above).

>I wonder if by default we should have that granularity or the tool just assume that 1.0 is 1.0.0 if there is no third digit?

---

It is true that you typically see Major.Minor version. But do they follow a semantic which is common and the user is able to rely on? Moreover, is that the semantic [we defined](https://github.com/oasis-tcs/csaf/blob/master/csaf_2.0/prose/csaf-v2-editor-draft.md#31112-version-type---semantic-versioning)?

**Excerpt from 3.1.11.2 Semantic versioning**
> 5. Patch version Z (x.y.Z | x > 0) MUST be incremented if only backwards compatible bug fixes are introduced. A bug fix is defined as an internal change that fixes incorrect behavior.
>
>   > In the context of the document this is the case  e.g. for spelling mistakes.
>
>6. Minor version Y (x.Y.z | x > 0) MUST be incremented if the content of an existing element changes except for those which are covert through rule 7. It MUST be incremented if substantial new information are introduced or new elements are provided. It MAY include patch level changes. Patch version MUST be reset to 0 when minor version is incremented.
>7. Major version X (X.y.z | X > 0) MUST be incremented if a new comparison with the end user's asset database is required. This includes:
>
>   * changes (adding, removing elements or modifying content) in `/product_tree` or elements which contain `/product_tree` in their path
>   * adding or removing items of `/vulnerabilities`
>   * adding or removing elements in:
>     * `/vulnerabilities[]/product_status/first_affected`
>     * `/vulnerabilities[]/product_status/known_affected`
>     * `/vulnerabilities[]/product_status/last_affected`
>   * removing elements from:
>     * `/vulnerabilities[]/product_status/first_fixed`
>     * `/vulnerabilities[]/product_status/fixed`
>     * `/vulnerabilities[]/product_status/known_not_affected`
>
>   It MAY also include minor and patch level changes. Patch and minor version MUST be reset to 0 when major version is incremented.

If a user sees a semantic version with only Major.Minor, he can't be sure whether it follows the rules from CSAF 2.0. By introducing the Patch-level as mandatory, we resolve that and make sure we provide enough guidance to it.

tschmidtb51:
As a side note: We silently fix spelling mistakes without issuing a new minor version. However, if we continue to do that while all stuff is going to be automated - there will be problems with signature as an old signature won't match the new file...

PR #242, opened 2021-05-03
Versioning

tschmidtb51:
- addresses part of oasis-tcs/csaf#214
- insert rules for CSAF asset matching systems
- add rule regard "do a new run whenever there is a change in the Major version if semantic versioning is used"


santosomar (approved):
Thank you @tschmidtb51 !

sthagen (approved):
LGTM

Issue #245, opened 2021-05-12
JSON Schema reference should be in HTTPS


santosomar:
As a best practice, the JSON Schema reference should be in HTTPS instead of HTTP. In other words, changing:

```
"$schema": "http://json-schema.org/draft-07/schema#",
```

to

```
"$schema": "https://json-schema.org/draft-07/schema#",
```

PR #246, opened 2021-05-12
JSON schema namespace in HTTPS

santosomar:
Addressing #245 

As a best practice, the JSON Schema reference should be in HTTPS instead of HTTP. In other words, changing:
```
"$schema": "http://json-schema.org/draft-07/schema#",
```
to

```
"$schema": "https://json-schema.org/draft-07/schema#",
```


tschmidtb51 (approved):
LGTM!

sthagen (approved):
Thanks, one little `s` can make all the difference :+1:

PR #250, opened 2021-05-20
Versioning

tschmidtb51:
- addresses part of oasis-tcs#214
- add rules for semantic versioning before initial public release


sthagen (approved):
I am a bit lost as to why we target master directly. Shouldn't that go into the May revision branch?

tschmidtb51:
> I am a bit lost as to why we target master directly. Shouldn't that go into the May revision branch?

You are right. It should actually target the May revision.

PR #251, opened 2021-05-20
Tests

tschmidtb51:
- addresses part of oasis-tcs#195
- add test "Multiple Use of Same CVE"


sthagen (approved):
The change looks good (of course) just wondering if the branch target is really master?

tschmidtb51:
> The change looks good (of course) just wondering if the branch target is really master?

You are right. I changed that to May revision.

santosomar (approved):

Issue #152, opened 2020-11-25
Proposal: Guidance on how and where to find CSAF documents


tschmidtb51:
We should provide some guidance how and where a publisher should provide the security advisories. This will also help the users to find the security advisories at a publisher’s / vendor’s website. Moreover, it aids in making the format machine processable.

Therefore, I suggest to define several ~conformance targets~ levels:
- **CSAF publisher**: publishes CSAF documents.
> Idea: Lowest level, easy to accomplish.
- **CSAF provider**: provides CSAF documents under a given path.
> Idea: Helps to automate it.

- **CSAF trusted provider**: provides CSAF documents under a given path and has them hashed and signed.
> Idea: Level of trust and confidence we actually want to have.

- **CSAF aggregator**: provides CSAF documents from other CSAF publishers
> Idea: Provide a point of contact so that the end user doesn't have to crawl multiple websites.

- ~~**CSAF modifier**: provides CSAF documents which base on CSAF documents which have been published by another CSAF publisher.~~
> ~~Idea: A company might want to share with their customers a in-depth or branch / sector specific analysis which takes the context in to account. It takes the original CSAF and produces a new one which also includes its information.~~

- ~~**CSAF translator**: translates a CSAF document from one language into another. See #156 for details.~~
> ~~Idea: Not everyone speaks English. Government might be required to issue advisories in more than one language.~~

**Update:** I removed CSAF translator and CSAF modifier from here and pasted them in #140 since they base on the definition of CSAF producers. Therefore, they are not related to the question where to find CSAF documents.

**Edit**: I removed the term conformance targets as there was some confusion. To clarify: This is independent from #140 - it does not specify anything at the document level.

tschmidtb51:
This should also be linked to #140 .

tschmidtb51:
Here are some idea for the requirements in general (which should be divide for the conformance targets as stated above):

- A publisher must provide valid CSAF documents. The file name shall be the value of `document/tracking/id` in lower case where non-printable characters and non-ASCII characters have been replaced with the character underscore. The file extension must be json.
The URL https://[www.]domain.tld/ .well-known/security/data/csaf/latest/ must link to a folder where CSAF documents which use the latest version of CSAF are located. CSAF documents of version <version> should also be accessible at https://[www.]-domain.tld/ .well-known/security/data/csaf/<version>/. The index file must provide a list of all file names of CSAF documents which are located in this directory.
[...]

PR #254, opened 2021-05-22
Distributing

tschmidtb51:
- addresses the first part of #152
- add rules agreed on TC meeting Jan 27th, 2021

### DIFF
--- a/csaf_2.0/prose/csaf-v2-editor-draft.md
+++ b/csaf_2.0/prose/csaf-v2-editor-draft.md
@@ -3289,7 +3289,7 @@ Informatiove tests provide insights in common mistakes and bad practices. They M
 
 # 5 Distributing CSAF documents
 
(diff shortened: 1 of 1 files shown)


sthagen (approved):
I think this is a good start. I wonder if other transports than per TLS might be desirable? What about QUIC?

tschmidtb51:
> I think this is a good start. I wonder if other transports than per TLS might be desirable? What about QUIC?

AFAIK: QUIC incorporates TLS 1.3 (see [QUIC](https://quicwg.org/base-drafts/draft-ietf-quic-http.html#name-delegation-to-quic).

tschmidtb51:
@santosomar: Please review.

sthagen (approved):
:+1:

santosomar (approved):
Looks good! Thank you again for all your time and help!

Issue #204, opened 2021-04-08
Consider a set of proven values for maxima of string and array lengths


sthagen:
When offering an API everyone and her dog will set practical limits for security reasons on the length of strings and arrays.

I suggest we do not leave the consumers and producers alone in the rainy situation of agreeing on some set of such limits.

can the members of the TC evaluate their corpus of existing advisories to determine at least a set of minimal sizes every process parsing csaf documents shall be able to handle?

I think this is an important contribution of the spec and the best place also.

I will suggest an amended safety, security, and data protection section regardless. But, if we can identify and agree on such a set of at least accept that set of limits for conforming roles / profiles it would be much easier to write the security aspect into the dedicated section (short, sweet, and clearly motivated).

Good idea, bad idea? Please give feedback - much appreciated.

tschmidtb51:
Under the impulse of the moment, I fear that a limitation in length of strings might contradict our intention of detailed descriptions, notes,...

Let me give 2 examples why I think it is really hard to decide:

- `document/publisher/name` (for the sake of simplicity just assuming that #201 comes through): Is 50 enough? Definitely not (since we have a longer name - even without mention the section)...
- `branches[]`: How many items should we allow here? A company could easily have >50000 products. If there is a vulnerability which affects a core component you have a lot of products to describe in a advisory...

However, I agree that people will implement certain limitations - and we should provide the guidance. For example, as a developer I would start looking into the recursion of `branches_t`. I think a minimal size of 25 should be enough.

We should be aware that if we set such minimal sizes anyone can only produce advisories within that boundaries to make sure that the consumer are able to process the documents.

Nevertheless, I'm happy to define such boundaries if they exist. I'm open to the TC's discussion.

tschmidtb51:
Maybe we should add that to the agenda of the next meeting?

santosomar:
Considered done. It will be added to the agenda.

sthagen:
Added the email tag for consistency 🙈

sthagen:
To help maintain the focus of the discussion: I do not believe finding a set of smallest useful maxima for interoperability is easy, but if we do not even try, the communities have to take over, which I think is not a good message.

I imagine that aggregators might want to use that format to, and there we may have astronomical collections. But, for interchange there has to be a way 🚀

sthagen:
We can try to start from the amount of bytes accepted per POST in typical REST API settings, and also from file size info by those active in the field in case transport is per rsync, sftp, or message queues.

[...]

PR #256, opened 2021-05-22
Add soft limits as appendix C

tschmidtb51:
In #204 we discussed limits. This PR provides a suggestion to introduce soft limits. They should be use for interoperability but do not render a bigger CSAF document invalid. (However, an end user might not be able to process a CSAF document which exceeds the limits.) As we don't have much experience here - I suggest to keep this as guidance for soft limits.

### DIFF
--- a/csaf_2.0/prose/csaf-v2-editor-draft.md
+++ b/csaf_2.0/prose/csaf-v2-editor-draft.md
@@ -3693,3 +3693,288 @@ Zach | Turk | Microsoft
 | Revision | Date | Editor | Changes Made |
 | :--- | :--- | :--- | :--- |
 | csaf-v2.0-wd20210521 | 2021-05-21 | Stefan Hagen | Editor revision for TC review |
+
+# Appendix C. Guidance on the Size of CSAF Documents
+
+The TC carefully considered all aspects in regard to suggesting size limits for CSAF documents. It was decided that hard limits should not be enforced. However, since there is the need for guidance to ensure interoperability in the eosystem, the TC provides a set of soft limits. A CSAF document which exceeds those, can still be valid but it might not be processable for some parties.
+
+All _CSAF consumers_ should be able to process CSAF documents which comply with the limits below. All _CSAF producers_ should not produce CSAF documents which exceed those limits.
+
+> If you come across a case where these limits are exceeded, please provide feedback to the TC.
+
+## File size
+
+A CSAF document should be smaller than 15.9 MB.
+
+## Array length
+
+An array should not have more than:
+
+* 10 000 items for
+  * `/document/acknowledgments`
+  * `/document/acknowledgments[]/names`
+  * `/document/acknowledgments[]/urls`
+  * `/document/tracking/aliases`
+  * `/product_tree/branches[]/product/product_identification_helper/hashes`
+  * `/product_tree/branches[]/product/product_identification_helper/hashes[]/file_hashes`
+  * `/product_tree/branches[]/product/product_identification_helper/x_generic_uris`
+  * `/product_tree/branches[](/branches[])*/product/product_identification_helper/hashes`
+  * `/product_tree/branches[](/branches[])*/product/product_identification_helper/hashes[]/file_hashes`
+  * `/product_tree/branches[]/branches[]/product/product_identification_helper/x_generic_uris`
+  * `/product_tree/full_product_names[]/product_identification_helper/hashes`
+  * `/product_tree/full_product_names[]/product_identification_helper/hashes[]/file_hashes`
+  * `/product_tree/full_product_names[]/product_identification_helper/x_generic_uris`
+  * `/product_tree/relationships[]/full_product_name/product_identification_helper/hashes`
+  * `/product_tree/relationships[]/full_product_name/product_identification_helper/hashes[]/file_hashes`
+  * `/product_tree/relationships[]/full_product_name/product_identification_helper/x_generic_uris`
+  * `/vulnerabilities[]/acknowledgments`
+  * `/vulnerabilities[]/acknowledgments[]/names`
+  * `/vulnerabilities[]/acknowledgments[]/urls`
+  * `/vulnerabilities[]/id/system_name`
+  * `/vulnerabilities[]/id/text`
+  * `/vulnerabilities[]/remediations[]/entitlements`
+
+* 40 000 items for
+  * `/document/notes`
+  * `/document/references`
+  * `/vulnerabilities[]/involvements`
+  * `/vulnerabilities[]/notes`
+  * `/vulnerabilities[]/references`
+
+* 100 000 for
+  * `/document/tracking/revision_history`
+  * `/product_tree/branches`
+  * `/product_tree(/branches[])*/branches`
+  * `/product_tree/branches[]/product/product_identification_helper/serial_numbers`
+  * `/product_tree/branches[](/branches[])*/product/product_identification_helper/serial_numbers`
+  * `/product_tree/full_product_names`
+  * `/product_tree/full_product_names[]/product_identification_helper/serial_numbers`
+  * `/product_tree/product_groups[]/product_ids`
+  * `/product_tree/relationships[]/full_product_name/product_identification_helper/serial_numbers`
+  * `/vulnerabilities`
+
+* 10 000 000 for
+  * `/product_tree/relationships`
+  * `/product_tree/product_groups`
+  * `/vulnerabilities[]/remediations[]/group_ids`
+
+* 100 000 000 for
+  * `/vulnerabilities[]/product_status/first_affected`
+  * `/vulnerabilities[]/product_status/first_fixed`
+  * `/vulnerabilities[]/product_status/fixed`
+  * `/vulnerabilities[]/product_status/known_affected`
+  * `/vulnerabilities[]/product_status/known_not_affected`
+  * `/vulnerabilities[]/product_status/last_affected`
+  * `/vulnerabilities[]/product_status/recommended`
+  * `/vulnerabilities[]/product_status/under_investigation`
+  * `/vulnerabilities[]/remediations`
+  * `/vulnerabilities[]/remediations[]/product_ids`
+  * `/vulnerabilities[]/scores`
+  * `/vulnerabilities[]/scores[]/products`
+  * `/vulnerabilities[]/threats`
+  * `/vulnerabilities[]/threats[]/group_ids`
+  * `/vulnerabilities[]/threats[]/product_ids`
+
+## String length
+
+A string should not have a length greater than:
+
+* 1000 for
+  * `/document/acknowledgments[]/names[]`
+  * `/document/acknowledgments[]/organization`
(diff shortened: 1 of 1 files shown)


sthagen (approved):
Thank you very much, @tschmidtb51 - I think this was a lot of work and will enable the members of the TC and of the wider public to comment on specifics. This way we may gather important information that allows us to provide both best effort and accepted limits by and to the communities using CSAF.

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -3692,4 +3692,291 @@ Zach | Turk | Microsoft
...
+
+All _CSAF consumers_ should be able to process CSAF documents which comply with the limits below. All _CSAF producers_ should not produce CSAF documents which exceed those limits.
+
+> If you come across a case where these limits are exceeded, please provide feedback to the TC.
+
+## File size
+
+A CSAF document in the specified JSON format encoded in UTF-8 should be not lager than 15 MB.
+
+>At least one database technology in wide use for storing CSAF documents rejects insert attempts when the transformed BSON size exceeds 16 megabytes. The BSON format optimizes for accessibility and not size. So, small integres and small strings may incur more overhead in the BSOn format than in JSON. In addition, the BSON format adds lebgth information for the entries inside the document which adds to the size when storing CSAF document content in a BSON format.
```
mprpic, line 3709:
```suggestion
> At least one database technology in wide use for storing CSAF documents rejects insert attempts when the transformed BSON size exceeds 16 megabytes. The BSON format optimizes for accessibility and not size. So, small integers and small strings may incur more overhead in the BSON format than in JSON. In addition, the BSON format adds length information for the entries inside the document, which adds to the size when storing CSAF document content in a BSON format.
```

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -3692,4 +3692,291 @@ Zach | Turk | Microsoft
...
+  * `/vulnerabilities[]/notes[]/text`
+
+* 250 000 for
+  * `/vulnerabilities[]/remediations[]/details`
+  * `/vulnerabilities[]/remediations[]/restart_required/details`
+  * `/vulnerabilities[]/threats[]/details`
+
+## URI length
+
+A string with format `uri` should not have a length greater than 20000. This applies to:
```
mprpic, line 3887: Most browsers have a limit of 2,000 characters on URLs, so 20,000 seems very generous.

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -3692,4 +3692,291 @@ Zach | Turk | Microsoft
...
+* `/product_tree/relationships[]/full_product_name/product_identification_helper/x_generic_uris[]/uri`
+* `/vulnerabilities[]/acknowledgments[]/urls[]`
+* `/vulnerabilities[]/references[]/url`
+* `/vulnerabilities[]/remediations[]/url`
+
+## Enum
+
+A string which is an enum has a fixed maximum length given by its longest value.
+
+> Later versions of CSAF migth add, modify or delete possible value. Therefore, this sizes should not be implemented as fixed limits.
```
mprpic, line 3909: Not sure I understand this note and this section in general. The maximum size of a string that can use only one of the values from an enum is not variable. So as long as the CSAF spec does not exceeds the string soft limit for a particular enum, then there is nothing to adhere to as far as consumers/producers go.

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -3692,4 +3692,291 @@ Zach | Turk | Microsoft
...
+* `/vulnerabilities[]/scores[]/cvss_v3/modifiedScope`
+* `/vulnerabilities[]/scores[]/cvss_v3/modifiedConfidentialityImpact`
+* `/vulnerabilities[]/scores[]/cvss_v3/modifiedIntegrityImpact`
+* `/vulnerabilities[]/scores[]/cvss_v3/modifiedAvailabilityImpact`
+* `/vulnerabilities[]/scores[]/cvss_v3/environmentalSeverity`
+* `/vulnerabilities[]/threats[]/category`
+
+## Date
+
+The maxima length of strings representing a temporal value is given by the format specifier. This applies to:
```
mprpic, line 3973:
```suggestion
The maximum length of strings representing a temporal value is given by the format specifier. This applies to:
```

mprpic (commented):
LGTM!

The limits themselves look fine. I'm curious how they were chosen though :-)

csaf_2.0/prose/csaf-v2-editor-draft.md
sthagen: Well, that is a URI :wink: not necessarily a URL ... but, the limits we will publish after discussing specific first (possibly other) limits will be the real thing and gain of this nice proposal from @tschmidtb51 ... :+1:

csaf_2.0/prose/csaf-v2-editor-draft.md
sthagen: Yes, I think this is maybe "belt and braces" style - I understood this information as: Look, we have now a specific set of enumerations i.e. strings with a certain fixed length as of CSAF v2.0 - do not implement a filter rejecting longer entries when you are not willing to adjust that limit later (future revisions).

I can imagine that we will have a different wording "here" when we publish.

mprpic: Ah, ok. That makes sense. Perhaps rewording it to how you described it would help make that a bit clearer :-)

csaf_2.0/prose/csaf-v2-editor-draft.md
mprpic: True, but if you look at the list of attributes below, more than a third are URLs even though the type of the field is technically `uri`.

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: I hope the new wording makes it clearer...

sthagen (approved):
Seen already in another PR (which is OK).

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -3307,7 +3324,46 @@ Example which fails the test:
...
+
+The document is a valid CSAF document (cf. Conformance clause 1).
+
+### 5.1.2 Requirement 2: Filename
+
+The CSAF document has a filename according to the rules in section 3 (cf. section 3.2.1.12.4).
+
+### 5.1.3 Requirement 3: TLS
+
+The CSAF document is retrievable from a website which uses TLS for encryption and server authenticity. The CSAF document MUST not be downloadable from a location which does not encrypt the transport.
```
tolim, line 3345: I do not understand why we put emphasis on encryption here. In many cases the document will be public, so we should focus on server authentication and data integrity. 
Do we want to also force TLS? What about PGP to ensure those requirements? I am thinking about the typical package mirrors of Linux distributions that are still available via HTTP and FTP.

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -3730,4 +3786,291 @@ Zach | Turk | Microsoft
 
 | Revision | Date | Editor | Changes Made |
 | :--- | :--- | :--- | :--- |
-| csaf-v2.0-wd20210521 | 2021-05-21 | Stefan Hagen | Editor revision for TC review |
+| csaf-v2.0-wd20210521 | 2021-05-21 | Stefan Hagen and Thomas Schmidt| Editor revision for TC review |
+
+# Appendix C. Guidance on the Size of CSAF Documents
+
+The TC carefully considered all known aspects to provide size limits for CSAF documents for this version of the specification. It was decided that hard limits should not be enforced. However, since there is the need for guidance to ensure interoperability in the eosystem, the TC provides a set of soft limits. A CSAF document which exceeds those, can still be valid but it might not be processable for some parties.
```
tolim, line 3793:
```suggestion
The TC carefully considered all known aspects to provide size limits for CSAF documents for this version of the specification. It was decided that hard limits should not be enforced. However, since there is the need for guidance to ensure interoperability in the ecosystem, the TC provides a set of soft limits. A CSAF document which exceeds those, can still be valid but it might not be processable for some parties.
```

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -3730,4 +3786,291 @@ Zach | Turk | Microsoft
...
+
+All _CSAF consumers_ should be able to process CSAF documents which comply with the limits below. All _CSAF producers_ should not produce CSAF documents which exceed those limits.
+
+> If you come across a case where these limits are exceeded, please provide feedback to the TC.
+
+## File size
+
+A CSAF document in the specified JSON format encoded in UTF-8 should be not lager than 15 MB.
```
tolim, line 3801:
```suggestion
A CSAF document in the specified JSON format encoded in UTF-8 should not be larger than 15 MB.
```

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -1706,6 +1706,23 @@ Examples:
     cisco-sa-20190513-secureboot
 ```
 
+This value is also used to define the filename for the CSAF document. The following rules MUST be applied to determine the filename for the CSAF document:
```
tolim, line 1709: We specify the filename of our CSAF document in a quite hidden section under `Document Property - Tracking - ID`. I would not have expected that as a reader. Can we find some other, more adequate place for this specification so that it can also be found in the TOC?

tolim (commented):
Also LGTM! I like that we only specify soft limits.

sthagen (approved):
LGTM.

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: We changed that to "be smaller".

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: We should track that as a separate issue. Therefore: See #277.

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: @mprpic: Is it now clearer?

sthagen (approved):
LGTM.

santosomar (approved):

mprpic (approved):

santosomar:
from Martin Prpic Red Hat (Guest) to Everyone:    2:10  PM
...in UTF-8 SHOULD conform to known size limits of current technologies parsing JSON content, e.g.: <mongo example>

tschmidtb51:
@santosomar: Ready for Merge.

sthagen (approved):
LGTM.

Issue #262, opened 2021-05-25
Consider broaden the definition of impact 


tschmidtb51:
The VEX subgroup suggested to broaden the definition of the category `impact` in `/vulnerabilities[]/threats[]` to also cover descriptions why something is not exploitable / vulnerable.

The current CSAF definiton is:
> The value `impact` indicates that the `details` field contains an assessment of the impact on the user or the target set if the vulnerability is successfully exploited. If applicable, for consistency and simplicity, this section can be a textual summary of the three CVSS impact metrics. These metrics measure how a vulnerability detracts from the three core security properties of an information system: Confidentiality, Integrity, and Availability.

The suggested change adds the words: "or a description why it cannot be exploited."
The new definition would be:

> The value `impact` indicates that the `details` field contains an assessment of the impact on the user or the target set if the vulnerability is successfully exploited or a description why it cannot be exploited. If applicable, for consistency and simplicity, this section can be a textual summary of the three CVSS impact metrics. These metrics measure how a vulnerability detracts from the three core security properties of an information system: Confidentiality, Integrity, and Availability.

Alternative: Construct a new category for such statements.

PR #263, opened 2021-05-25
Definition of impact

tschmidtb51:
- resolves oasis-tcs/csaf#262
- add suggested extension to the definition


sthagen (approved):
Good addition, I think.

sthagen:
Merged during meeting of TC following motion to resolve.

Issue #260, opened 2021-05-25
Consider VEX definition alignment for product_status


santosomar:
Original email from Allan:

From: "Friedman, Allan" <AFriedman@ntia.gov>
Reply-To: VEX Subgroup <ntia_sbom_vex@osmmail.ntia.doc.gov>
Date: Monday, May 24, 2021 at 8:47 PM
To: VEX Subgroup <ntia_sbom_vex@osmmail.ntia.doc.gov>
Subject: VEX - finalizing definitions (pls review by 12:01am May 26)

Hi Folks,

We had a productive discussion going through most of the rest of the CSAF fields, and making sure they met our needs for at least a VEX 1.0. Link to the details below with some notes.

We now need to win over the CSAF community for the core definition of VEX: not affected and affected. These currently have a bit of a circular definition, so hopefully, we'll win some friends.

Below are the definitions that we've workshopped. Please review and share any urgent critical comments, ideally with a suggested solution. If anyone wants more background on these, and what we're trying to do, let me know, and I'll give you a few min.

Defining /vulnerabilities[]/product_status

known_not_affected 
No mitigation is required regarding this vulnerability. 
This could be because the code referenced in the vulnerability is not present, not exposed, compensating controls exist, or other factors. 
See /vulnerabilities[]/notes and /vulnerabilities[]/threats for more details

known_affected
Actions are recommended to mitigate or address this vulnerability 
This could include learning more about the vulnerability and context, and/or making a risk- based decision. See /vulnerabilities[]/notes, /vulnerabilities[]/threats, and /vulnerabilities[]/remediations for more details.

Fixed  (definition from current CSAF docs)
These versions contain a fix for the vulnerability but may not be the recommended fixed versions.

Under_Investigation  (definition from current CSAF docs)
It is not known yet whether these versions are or are not affected by the vulnerability. However, it is still under investigation - the result will be provided in a later release of the document.

CSAF / VEX mapping: https://docs.google.com/spreadsheets/d/109NHx0FBkgmgVGrDCKPmj3SitP75_9WeMrtYSEXGGRc/edit#gid=0

santosomar:
These definitions look very good. One minor suggestion: perhaps add the word “remediation” to the known_not_affected definition, as follows:

known_not_affected 
No mitigation or remediation is required regarding this vulnerability. 

The reason of my silly request is because often people refer to a “mitigation” as a workaround and they refer to the ultimate fix as the final “remediation”.

[...]

PR #266, opened 2021-05-25
Suggestion for Definition of Product Status

tschmidtb51:
- resolves oasis-tcs/csaf#260
- add definitions to explain affected
- need decision on oasis-tcs/csaf#263

### DIFF
--- a/csaf_2.0/prose/csaf-v2-editor-draft.md
+++ b/csaf_2.0/prose/csaf-v2-editor-draft.md
@@ -2191,9 +2191,14 @@ First fixed (`first_fixed`) of value type Products (`products_t`) represents tha
 
 Fixed (`fixed`) of value type Products (`products_t`) represents that these versions contain a fix for the vulnerability but may not be the recommended fixed versions.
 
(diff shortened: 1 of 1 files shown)


tschmidtb51:
@allan-ntia: Please review.

allan-ntia:
I think this approach makes sense. 

I would just add this: I think some of the folks in the VEX world liked the set of examples of potential actions for known_affected. In particular, I've heard the idea that the consumer may have to learn more to take action. e.g. addressing vulns that are specific to configuration or other details, as well as the explicit acknowledgement that the action may be something other than a traditional patch. 

Now, we can always add the "for instance" text for the VEX profile. Just wanted to try to capture this feedback.

tschmidtb51:
I tried to incorporate the feedback.

sthagen (approved):
I like it. Thanks.

santosomar:
Same here! Thank you @tschmidtb51 @allan-ntia @sthagen!

santosomar (approved):

tschmidtb51:
As discussed in VEX subgroup at Jun 7th, 2021: Okay. Ready to merge.

PR #269, opened 2021-05-26
Distributing CSAF documents

tschmidtb51:
This PR tries to summarize the discussions from #152 and sketches an actionable solution which we could include in the standard to support automation.

### Note

This does **not add** any additional burden for vendors as it is **truly optional**. However, we provide a standard solution for people to ask from their vendors if they want to automate the handling of CSAF documents.

#### Summary
- addresses parts of oasis-tcs/csaf#152
- add requirements with draft description
- add roles CSAF provider and CSAF trusted provider

### DIFF
--- a/csaf_2.0/prose/csaf-v2-editor-draft.md
+++ b/csaf_2.0/prose/csaf-v2-editor-draft.md
@@ -3412,13 +3412,13 @@ Directory listing SHALL be enabled to support manual navigation.
 
 ### 5.1.11 Requirement 11: ROLIE service document
 
-The ROLIE service document MUST conform with [RFC8322].
+The ROLIE service document MUST be a JSON file that conforms with [RFC8322].
 
 **TODO: Provide Example**
 
 ### 5.1.12 Requirement 12: ROLIE category document
 
-Each ROLIE category document MUST conform with [RFC8322].
+Each ROLIE category document MUST be a JSON file that conforms with [RFC8322].
 
 **TODO: Provide Example**
 
--- a/csaf_2.0/prose/csaf-v2-editor-draft.md
+++ b/csaf_2.0/prose/csaf-v2-editor-draft.md
@@ -3395,7 +3395,7 @@ Examples:
 
 ### 5.1.9 Requirement 9: changes.csv
 
-The file changes.csv must contain the filename as well as the value of `/document/tracking/current_release_date` for each CSAF document in the sub-directories without a heading; lines must be sorted by the latter one.
+The file changes.csv must contain the filename as well as the value of `/document/tracking/current_release_date` for each CSAF document in the sub-directories without a heading; lines must be sorted by the `current_release_date` timestamp with the latest one first.
 
 Examples:
 
--- a/csaf_2.0/prose/csaf-v2-editor-draft.md
+++ b/csaf_2.0/prose/csaf-v2-editor-draft.md
@@ -3442,7 +3442,16 @@ Example:
 
 ```
 File name of CSAF document: example_company_-_2019-yh3234.json
-File name of hash file: example_company_-_2019-yh3234.json.sha512
+File name of SHA-256 hash file: example_company_-_2019-yh3234.json.sha256
+File name of SHA-512 hash file: example_company_-_2019-yh3234.json.sha512
+```
+
+The file content SHALL start with the first byte of the hexadecimal hash value. Any subsequent data (like a filename) which is optional SHALL be separated by at least one space.
+
+Example:
+
+```
+ea6a209dba30a958a78d82309d6cdcc6929fcb81673b3dc4d6b16fac18b6ff38  example_company_-_2019-yh3234.json
 ```
 
 ### 5.1.16 Requirement 16: Signatures
--- a/csaf_2.0/prose/csaf-v2-editor-draft.md
+++ b/csaf_2.0/prose/csaf-v2-editor-draft.md
@@ -321,6 +321,8 @@ N. Williams., "JavaScript Object Notation (JSON) Text Sequences", RFC 7464, DOI 
 ###### [SCAP12]
 _The Technical Specification for the Security Content Automation Protocol (SCAP): SCAP Version 1.2_, D. Waltermire, S. Quinn, K. Scarfone, A. Halbardier, Editors, NIST Spec. Publ. 800‑126 rev. 2, September 2011, 
 http://dx.doi.org/10.6028/NIST.SP.800-126r2.
+###### [SECURITY-TXT]
+Foudil, E. and Shafranovich, Y., _Security.txt Project_, https://securitytxt.org/.
 ###### [SemVer]
 _Semantic Versioning 2.0.0_, T. Preston-Werner, June 2013, https://semver.org/.
 ###### [XML]
@@ -3352,7 +3354,7 @@ This does not exclude that such a document is also available in an access protec
 
 ### 5.1.5 Requirement 5: security.txt
 
-In the security.txt there MUST be at least one field `CSAF` which points to either the ROLIE service document or a directory with CSAF files. If this field indicates a web URI, then it MUST begin with "https://" (as per section 2.7.2 of [RFC7230]).
+In the security.txt there MUST be at least one field `CSAF` which points to either the ROLIE service document or a directory with CSAF files. If this field indicates a web URI, then it MUST begin with "https://" (as per section 2.7.2 of [RFC7230]). See [SECURITY-TXT] for more details.
 
 > At the time of this writing, the security.txt is still a proposed standard. The `CSAF` field has not been officially added yet.
 
--- a/csaf_2.0/prose/csaf-v2-editor-draft.md
+++ b/csaf_2.0/prose/csaf-v2-editor-draft.md
@@ -3414,27 +3414,84 @@ Directory listing SHALL be enabled to support manual navigation.
 
 ### 5.1.11 Requirement 11: ROLIE service document
 
-The ROLIE service document MUST be a JSON file that conforms with [RFC8322].
+Resource-Oriented Lightweight Information Exchange (ROLIE) is a standard to ease discovery of security content. ROLIE is built on top of the Atom Publishing Format and Protocol, with specific requirements that support publishing security content. The ROLIE service document MUST be a JSON file that conforms with [RFC8322] and lists the ROLIE feed documents.
 
 **TODO: Provide Example**
 
-### 5.1.12 Requirement 12: ROLIE category document
+### 5.1.12 Requirement 12: ROLIE feed
 
-Each ROLIE category document MUST be a JSON file that conforms with [RFC8322].
+All CSAF documents with the same TLP level MUST be listed in a single ROLIE feed. At least one of the feeds
+
+* TLP:WHITE
+* TLP:GREEN
+* unlabeled
+
+MUST exist. Each ROLIE feed document MUST be a JSON file that conforms with [RFC8322].
+
+Example:
+
+```
+{
+  "feed": {
+    "id": "example-csaf-feed-tlp-white",
+    "title": "Example CSAF feed (TLP:WHITE)",
+    "link": [
+      {
+        "rel": "self",
+        "href": "https://psirt.domain.tld/advisories/csaf/feed-tlp-white.json"
+      }
+    ],
+    "category": {
+      "scheme": "urn:ietf:params:rolie:category:information-type",
+      "term": "security advisory"
+    },
+    "updated": "2021-01-01T12:00Z",
+    "entry": [
+      {
+        "id": "2020-ESA-001",
+        "title": "Example Security Advisory 001",
+        "link": [
+          {
+            "rel": "self",
+            "href": "https://psirt.domain.tld/advisories/csaf/2020/ESA-001.json"
+          }
+        ],
+        "published": "2021-01-01T11:00Z",
+        "updated": "2021-01-01T12:00Z",
+        "summary": {
+          "content": "Vulnerabilities fixed in ABC 0.0.1"
+        },
+        "content": {
+          "type": "application/json",
+          "src": "https://psirt.domain.tld/advisories/csaf/2020/ESA-001.json"
+        },
+        "format": {
+          "schema": "https://raw.githubusercontent.com/oasis-tcs/csaf/master/csaf_2.0/json_schema/csaf_json_schema.json",
+          "version": "2.0"
+        }
+      }
+    ]
+  }
+}
+```
+
+### 5.1.13 Requirement 13: ROLIE category document
+
+The use and therefore the existence of ROLIE category document is optional. If it is used, each ROLIE category document MUST be a JSON file that conforms with [RFC8322]. It should be used for to further dissects CSAF documents by their document categories.
 
 **TODO: Provide Example**
 
-### 5.1.13 Requirement 13: TLP:AMBER and TLP:RED
+### 5.1.14 Requirement 14: TLP:AMBER and TLP:RED
 
 CSAF documents labeled TLP:AMBER or TLP:RED MUST be access protected. If they are provided via a webserver this SHALL be done under a different path than for TLP:WHITE, TLP:GREEN and unlabeled CSAF documents. TLS client authentication, access tokens or any other automatable authentication method SHALL be used.
 
-### 5.1.14 Requirement 14: Redirects
+### 5.1.15 Requirement 15: Redirects
 
 Redirects SHOULD NOT be used. If they are inevitable only HTTP Header redirects are allowed.
 
 > Reasoning: Clients, as e.g. `curl`, do not follow any other kind of redirects.
 
-### 5.1.15 Requirement 15: Integrity
+### 5.1.16 Requirement 16: Integrity
 
 All CSAF documents SHALL have at least one hash file computed with a secure cryptographic hash algorithm (e.g. SHA-512 or SHA-3) to ensure their integrity. The filename is constructed by appending the file extension which is given by the algorithm.
 
@@ -3456,7 +3513,7 @@ Example:
 ea6a209dba30a958a78d82309d6cdcc6929fcb81673b3dc4d6b16fac18b6ff38  example_company_-_2019-yh3234.json
 ```
 
-### 5.1.16 Requirement 16: Signatures
+### 5.1.17 Requirement 17: Signatures
 
 All CSAF documents SHALL have at least one OpenPGP signature file which is provided under the same filename which is extended by the appropriate extension.
 
@@ -3467,7 +3524,7 @@ File name of CSAF document: example_company_-_2019-yh3234.json
 File name of signature file: example_company_-_2019-yh3234.json.asc
 ```
 
-### 5.1.17 Requirement 17: Public PGP Key
+### 5.1.18 Requirement 18: Public PGP Key
 
 The public part of the PGP key used to sign the CSAF documents MUST be available. It SHOULD also be available at a publc key server.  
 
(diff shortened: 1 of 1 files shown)


csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -3352,6 +3350,116 @@ This does not exclude that such a document is also available in an access protec
...
+
+### 5.1.14 Requirement 14: Redirects
+
+Redirects SHOULD NOT be used. If they are inevitable only HTTP Header redirects are allowed.
+
+> Reasoning: Clients, as e.g. `curl`, do not follow any other kind of redirects.
+
+### 5.1.15 Requirement 15: Integrity
+
+All CSAF documents SHALL have at least one hash file computed with a secure cryptographic hash algorithm (e.g. SHA-512 or SHA-3) to ensure their integrity. The filename is constructed by appending the file extension which is given by the algorithm.
```
sthagen, line 3437: I suggest to later allow a text file hold more than just the hexadecimal hash value but only enforce that value to start at the first byte and any subsequent data (like a filename) be separated by at least one space.
Rationale: the typical generators like sha256sum produce such output and any reader parsing such a file will expect the first hash length bytes to contain the value.

sthagen (approved):
I think this is a very good start to receive constructive and fast converging input from the TC members and the wider public. Thanks a lot for providing @tschmidtb51

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -3352,6 +3350,116 @@ This does not exclude that such a document is also available in an access protec
 
 > Reasoning: If an advisory is already in the media, an end user should not be forced to collect the pieces of information from a press release but be able to retrieve the CSAF document.
 
+### 5.1.5 Requirement 5: security.txt
+
+In the security.txt there MUST at least one field `CSAF` which points to either the ROLIE service document or a directory with CSAF files. If this field indicates a web URI, then it MUST begin with "https://" (as per section 2.7.2 of [RFC7230]).
```
tolim, line 3355: Small fix:
```suggestion
In the security.txt there MUST be at least one field `CSAF` which points to either the ROLIE service document or a directory with CSAF files. If this field indicates a web URI, then it MUST begin with "https://" (as per section 2.7.2 of [RFC7230]).

File: csaf_2.0/prose/csaf-v2-editor-draft.md (hunk 1 above)
Comment (tschmidtb51) [reply csaf_2.0/prose/csaf-v2-editor-draft.md:3437]: I added a suggestion to address that.

**Review by sthagen — APPROVED** [review]
💚

**Review by sthagen — APPROVED** [review]
LGTM. Especially enforcing the JSON variant of ROLIE will help the CSAF tools to keep a minimal toolchain.

File: csaf_2.0/prose/csaf-v2-editor-draft.md [hunk 3]
```diff
@@ -3352,6 +3350,125 @@ This does not exclude that such a document is also available in an access protec
...
+
+Example:
+
+```
+File name of CSAF document: example_company_-_2019-yh3234.json
+File name of SHA-256 hash file: example_company_-_2019-yh3234.json.sha256
+File name of SHA-512 hash file: example_company_-_2019-yh3234.json.sha512
+```
+
+The file content SHALL start with the first byte of the hexadecimal hash value. Any subsequent data (like a filename) which is optional SHALL be separated by at least one space.
```
mprpic, line 3449: Instead of doubling the amount of files, could the hash be specified as a second field in the index.txt file with a hash prefix specifying the algorithm used? Something like:

```
2020/example_company_-_2020-yh4711.json sha256:321691c4527cb4510e1365d2a982cd141ddd73c2be7a04893f08e56006a40b2e
2019/example_company_-_2019-yh3234.json sha512:7fe6b968ec273dec239437f43ad7662f24484b1fd401c38509993a2e1e2972d47bb5543f7ba6d7bd41503731641e9200a8dcc49b9875ff53dd64098e902bafb9
2018/example_company_-_2018-yh2312.json sha384:f7f4840727a9f3be895dbe246275506fb550ec9b55f8264db606f2f81b6fd5931f7bb82c338df163730de7de6072f214
```

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -3352,6 +3350,125 @@ This does not exclude that such a document is also available in an access protec
 
 > Reasoning: If an advisory is already in the media, an end user should not be forced to collect the pieces of information from a press release but be able to retrieve the CSAF document.
 
+### 5.1.5 Requirement 5: security.txt
+
+In the security.txt there MUST be at least one field `CSAF` which points to either the ROLIE service document or a directory with CSAF files. If this field indicates a web URI, then it MUST begin with "https://" (as per section 2.7.2 of [RFC7230]).
```
mprpic, line 3355: Perhaps this could refer to https://securitytxt.org/ for more information about security.txt files.

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -3352,6 +3350,125 @@ This does not exclude that such a document is also available in an access protec
...
+2020/example_company_-_2020-yh4711.json
+2019/example_company_-_2019-yh3234.json
+2018/example_company_-_2018-yh2312.json
+```
+
+> This can be used to download all CSAF documents.
+
+### 5.1.9 Requirement 9: changes.csv
+
+The file changes.csv must contain the filename as well as the value of `/document/tracking/current_release_date` for each CSAF document in the sub-directories without a heading; lines must be sorted by the latter one.
```
mprpic, line 3398:
```suggestion
The file changes.csv must contain the filename as well as the value of `/document/tracking/current_release_date` for each CSAF document in the sub-directories without a heading; lines must be sorted by the `current_release_date` timestamp with the latest one first.
```

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -3352,6 +3350,125 @@ This does not exclude that such a document is also available in an access protec
...
+
+Directory listing SHALL be enabled to support manual navigation.
+
+### 5.1.11 Requirement 11: ROLIE service document
+
+The ROLIE service document MUST be a JSON file that conforms with [RFC8322].
+
+**TODO: Provide Example**
+
+### 5.1.12 Requirement 12: ROLIE category document
```
mprpic, line 3419: I don't think I would make this a requirement because categories using within a Category document have to be registered:

https://datatracker.ietf.org/doc/html/rfc8322#section-8.4

For CSAF specifically, I could imagine registering "security advisory" as an information type:

https://www.iana.org/assignments/rolie/rolie.xhtml

and then a subregistry with all the CSAF document categories, but seeing as they are not enforced in the schema, the two sources of information (ROLIE registry vs CSAF schema) might get out of sync really fast.

All in all, you can publish ROLIE feeds without any categorization information.

mprpic: Ok, I think I read that wrong, the specific information types don't have to be registered as indicated by this example:

https://datatracker.ietf.org/doc/html/rfc8322#appendix-B.1

which uses the term "vulnerability" despite not being included in the list of potential types:

https://datatracker.ietf.org/doc/html/rfc8322#section-7.1.2

A global `urn:ietf:params:rolie:category:information-type` feed attribute actually determines whether feed is a ROLIE feed or not so the last sentence of my previous comment can be ignored. For CSAF feeds, it can I guess be set to:

```json
    "category": {
      "scheme": "urn:ietf:params:rolie:category:information-type",
      "term": "security advisory"
    }
```

In any case, category documents are not special in any way to ROLIE and are general to the Atom protocol specification:

https://datatracker.ietf.org/doc/html/rfc5023#section-7

We could consider coming up with guidance on how to publish category documents to cover some basic category types for 2.x, but I don't think I would include any category document requirements in 2.0.

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -3352,6 +3350,125 @@ This does not exclude that such a document is also available in an access protec
...
+
+### 5.1.10 Requirement 10: Directory listings
+
+Directory listing SHALL be enabled to support manual navigation.
+
+### 5.1.11 Requirement 11: ROLIE service document
+
+The ROLIE service document MUST be a JSON file that conforms with [RFC8322].
+
+**TODO: Provide Example**
```
mprpic, line 3417:
```suggestion
### 5.1.11 Requirement 11: ROLIE feed document

Resource-Oriented Lightweight Information Exchange (ROLIE) is a proposed standard to ease discovery of security content. ROLIE is built on top of the Atom Publishing Format and Protocol, with specific requirements that support publishing security content. All CSAF documents MUST be listed in a single ROLIE feed (in JSON format) that conforms with [RFC8322]. You MAY include the CSAF feed in a Service Document and create a Category Document that further dissects CSAF documents by their document categories.

An example of a CSAF ROLIE feed:

{
  "feed": {
    "id": "example-csaf-feed",
    "title": "Example CSAF feed",
    "link": [
      {
        "rel": "self",
        "href": "https://psirt.domain.tld/advisories/csaf/feed.json"
      }
    ],
    "category": {
      "scheme": "urn:ietf:params:rolie:category:information-type",
      "term": "security advisory"
    },
    "updated": "2021-01-01T12:00Z",
    "entry": [
      {
        "id": "2020-ESA-001",
        "title": "Example Security Advisory 001",
        "link": [
          {
            "rel": "self",
            "href": "https://psirt.domain.tld/advisories/csaf/2020/ESA-001.json"
          }
        ],
        "published": "2021-01-01T11:00Z",
        "updated": "2021-01-01T12:00Z",
        "summary": {
          "content": "Vulnerabilities fixed in ABC 0.0.1"
        },
        "content": {
          "type": "application/json",
          "src": "https://psirt.domain.tld/advisories/csaf/2020/ESA-001.json"
        },
        "format": {
          "schema": "https://raw.githubusercontent.com/oasis-tcs/csaf/master/csaf_2.0/json_schema/csaf_json_schema.json",
          "version": "2.0"
        }
      }
    ]
  }
}
```

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: I'm not sure whether that is still as easy to generate/parse as it was intended. However, you do not need to fulfill requirement 8 (index.txt) if you provide a ROLIE document... IMHO having that as a separate file helps to use existing tools.

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: @sthagen: Please adopt the format if needed.

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: Is there a specific reason why we want to skip the service document?

I have a bunch of use cases in mind where that would help (e.g. the TLP:AMBER vs. TLP:WHITE discussion)... From my understanding it doesn't really hurt ;-) But I might be wrong here.

tschmidtb51:
@mprpic: Thank you for adding the ROLIE stuff! It's really appreciated.

csaf_2.0/prose/csaf-v2-editor-draft.md
sthagen: Generation is easy as `sha256sum *.whatever | sed s/^/sha256:/g > index.txt` and the collation of file to hash mappings into one file per "intent" is folklore I think. But, traditionally the hash is noted first and last the resource name (as per the example call I note on a Linux system ...). As an implementer I could easily live with bundled info in files with well-known names. Instead of "per file" one would analyze "per line".

I see the complexity rising though as @tschmidtb51 states. E.g. when sibling files are processed the target is fully specified. In such a bundled file, the interpretation of the target names may deviate from what one finds at the assumed path.

sthagen (approved):
LGTM - and the ongoing discussion.

csaf_2.0/prose/csaf-v2-editor-draft.md
mprpic: > I'm not sure whether that is still as easy to generate/parse as it was intended. However, you do not need to fulfill requirement 8 (index.txt) if you provide a ROLIE document... IMHO having that as a separate file helps to use existing tools.

I think the complexity isn't that much different as Stefan illustrated, it's just appending to a file instead of writing out a new one while iterating over a list of CSAF files.

What existing tools process hashes in separate files?

> I see the complexity rising though as @tschmidtb51 states. E.g. when sibling files are processed the target is fully specified. In such a bundled file, the interpretation of the target names may deviate from what one finds at the assumed path.

I'm not sure I understand this. What do you mean by a sibling file?

csaf_2.0/prose/csaf-v2-editor-draft.md
mprpic: To me, a service document would be created across an entire set of security content where each type (e.g. CSAF documents, OVAL documents, CVE documents, SBOM documents, etc.) is represented as a separate feed. And for that kind of a service document, you're well outside the scope of CSAF.

Though, I guess you're thinking of a way more granular use case where the service document would advertise several disparate CSAF feeds that would contain subsets of all CSAF documents with certain properties (e.g. TLP), right?

In any case, adding all CSAF ROLIE feeds to a service document certainly is not discouraged, but I wouldn't make it a requirement in this section.

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: > What existing tools process hashes in separate files?

I was thinking of `sha256sum` or `sha512sum`... They would use e.g. `sha256sum $i > ${i}.sha256` to generate and `sha256sum -c ${i}.sha256` to check the file. (It is assumed that `$i` is the filename of the current CSAF document.)

IMHO, adding the hashes into the index.txt makes updates a little bit more complex (now you would have to edit a line). I fear that this hinders SME to provide integrity as it is then not only uploading a bunch of files (CSAF+hash file) to the web server.

There are 3 other points I would like to bring up:
1. How would we deal then with providing multiple hashes for one file? 
    Tools might make all kind of assumptions (e.g. sha256 is always in the third field) or need to be specific for each vendor (which we tried to prevent). With separate files, we don't have this issue as everyone can choose the hash according to one's wishes / preferences. When you use a certain file extension, you either get a hash file or not. But you don't have to parse through a long list.
2. Does it scale for CSAF aggregators?
    If a CSAF aggregator has to do this, he basically needs to join all this from different files. I think this is more complex (also with updates and maybe deletions) than "just" copying more files. 
3. Big open source projects are already doing their releases that way, e.g.:
   * https://nextcloud.com/changelog/#latest21
   * https://httpd.apache.org/download.cgi
   * https://www.openssl.org/source/
   * http://www.releases.ubuntu.com/21.04/

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: I agree that such broad service document would be out of scope for CSAF.

> Though, I guess you're thinking of a way more granular use case where the service document would advertise several disparate CSAF feeds that would contain subsets of all CSAF documents with certain properties (e.g. TLP), right?

Yes. As we require that CSAF documents of different TLP versions must not be present under the same URL they should not be present in the same feed. (Otherwise, you can't access some documents without knowing why or even worse - you're able to access TLP:RED stuff that you should not have access to by accident...) Therefore, I would like to provide the opportunity for different feeds. To me, the RFC sounded like I have to do that via a service document (see [B.1](https://datatracker.ietf.org/doc/html/rfc8322#appendix-B.1)). Is there another way to do that?

> As a side note: I thought we could then use the categories for filtering. E.g. a big manufacturer puts each product family into a single category. Now the end user only selects a subset of categories which are relevant for him. Or one wants to group them by profile type.
> Nevertheless, an issuing party shall provide one feed with all CSAF documents (but based on their TLP level).
> Thoughts?

IMHO, the CSAF field in the security.txt should always link to same type of document to make it easier for automation. If we can provide the standard service document which is used by 80% - awesome.

tschmidtb51:
@mprpic: I integrated your content for the ROLIE requirements. I did make the Service Document required to be able to split the feeds to be inline with requirement 14 (TLP:AMBER and TLP:RED). I kept the ROLIE category document optional.

csaf_2.0/prose/csaf-v2-editor-draft.md
mprpic: > * How would we deal then with providing multiple hashes for one file?

True. Though, the ease of use of any hashing algorithm on any modern system makes me think that providing more than one is sort of pointless :-)

> * Does it scale for CSAF aggregators?
>   If a CSAF aggregator has to do this, he basically needs to join all this from different files. I think this is more complex (also with updates and maybe deletions) than "just" copying more files.
> 

I guess so, but I would assume that there will be a lot more CSAF publishers than aggregators so the spec should slightly favor the former when it comes to ease of use.

> * Big open source projects are already doing their releases that way, e.g.:
>   
>   * https://nextcloud.com/changelog/#latest21
>   * https://httpd.apache.org/download.cgi
>   * https://www.openssl.org/source/

All of these distribute just a handful of artifacts for which it makes no sense to produce an index file, so creating separate hash files is really the only option. For projects with 1000s of files though, I would guess single with all hashes may be more common (though I can't find a good example of this).

>   * http://www.releases.ubuntu.com/21.04/

This one actually uses the index with hashes (file `SHA256SUMS`) approach :-)

All in all, I see the use cases for single file hashes but I would still slightly prefer a single index of them. I'm fine keeping this as is though. It would be nice to hear from others on what they find more useful.

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -3352,6 +3352,184 @@ This does not exclude that such a document is also available in an access protec
...
+2018/example_company_-_2018-yh2312.json, "2019-03-01T06:01:00Z"
+```
+
+### 5.1.10 Requirement 10: Directory listings
+
+Directory listing SHALL be enabled to support manual navigation.
+
+### 5.1.11 Requirement 11: ROLIE service document
+
+Resource-Oriented Lightweight Information Exchange (ROLIE) is a standard to ease discovery of security content. ROLIE is built on top of the Atom Publishing Format and Protocol, with specific requirements that support publishing security content. The ROLIE service document MUST be a JSON file that conforms with [RFC8322] and lists the ROLIE feed documents.
```
mprpic, line 3417: Note that Service Documents aren't specific to ROLIE, they are merely mentioned in RFC8322 as a way to discover ROLIE feeds. So technically this section should be named "AtomPub service document" with a requirement that it points to specific ROLIE CSAF feeds.

That was my point about requiring a service document. If as a vendor you publish only CSAF documents of one TLP level, your service document will be a pointer to that one feed and that's it, making it unnecessary since you could just point to the feed in the first place in e.g. security.txt.

sthagen (approved):

santosomar:
This pulled requests was discussed and approved in the June 30, 2021 TC Monthly meeting. It was suggested to add example documents to support the change.

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -3352,6 +3352,184 @@ This does not exclude that such a document is also available in an access protec
...
+Example:
+
+```
+File name of CSAF document: example_company_-_2019-yh3234.json
+File name of signature file: example_company_-_2019-yh3234.json.asc
+```
+
+### 5.1.18 Requirement 18: Public PGP Key
+
+The public part of the PGP key used to sign the CSAF documents MUST be available. It SHOULD also be available at a publc key server.
```
tolim, line 3531:
```suggestion
The public part of the PGP key used to sign the CSAF documents MUST be available. It SHOULD also be available at a public key server.  
```

sthagen: We will ensure this gets fixed with a subsequent editorial changes PR

Issue #241, opened 2021-04-30
Considering inclusion of a signature field


lukehinds:
I would like to propose adding a signature / pubkey field to the csaf_2.0/json_schema to provide non-repudiation and some level of integrity verification of the claim. 

This way it will allow assurance that the claims within the schema have a degree of tamper resistance. If someone is able to amend an entries they could use that to mislead over vulnerability data and perform some level of targeted attack.

On the later note, we would be interested in having [rekor](https://github.com/sigstore/rekor) consume csaf data and validate the signing and store the claims into an immutable , tamper resistant transparency log. We already have a custom schema interface which would mean the csaf claims can be version controlled and referenced https://github.com/sigstore/rekor/tree/main/pkg/types

rekor supports [multiple signing tools](https://github.com/sigstore/rekor/blob/main/pkg/types/rekord/v0.0.1/rekord_v0_0_1_schema.json), so we have flexibility there as well (pgp, minisign, x509, ssh, openidconnect etc)

mprpic:
I'm a little confused as to how you imagine that field would be used. If I were to generate a CSAF document, then sign that document, and put that signature as a value of a hypothetical `signature` field then verifying that signature involves first stripping that field from the document and using the value to verify the modified document.

Normally, file signatures are provided in separate files so that the signature represents the state of the file in its complete form. You wouldn't expect to find the checksum of an ISO within the ISO itself, no? Maybe I'm just not understanding the ask here correctly :-)

lukehinds:
@mprpic sorry, should have clarified. The sig would be generated by each value part of all key / values concatenated together. So it is an attached signature, but not a signature of the entirety of the file itself (which is not possible as you rightly point out).

tschmidtb51:
Thank you for the suggestion.

I agree that signing CSAF documents is important to accomplish the goal of integrity protection. Given that there are a lot of signature wrapping attacks and signatures may not always be produced within the same (online) tool (e.g. [Secvisogram](https://secvisogram.github.io), I prefer detached signatures as suggested in https://github.com/oasis-tcs/csaf/issues/152#issuecomment-733804674:

>   All files shall have a cryptographic hash (e.g. SHA-512 or SHA-3) as well as an OpenPGP signature file which are provided under the same file name which is extended by the appropriate extension.
>    Example: File names
>    ```
>    Value of document/tracking/id: "Example Company - 2019-YH3234"
>    File name of CSAF document: example_company_-_2019-yh3234.json
[...]

PR #274, opened 2021-06-12
Publisher namespace

### LINKED ISSUE #267: Vendor ID issued by FIRST
In CVRF there was a informative comment:
> At the time of this writing, OID issuance by FIRST is still a work in progress, thus some samples are provided below, that use OIDs from other standard MIBs.

Can someone confirm that FIRST now issues the OIDs? Or do we need to change something? E.g. refer to [OID](http://oid-info.com)

tschmidtb51:
- resolves #267 
- remove Vendor ID as it was never really used
- introduce publisher namespace to allow for easier identification of publisher => leads in combination with `/document/tracking/id`to globally unique identifiers for CSAF documents


tschmidtb51:
@mprpic, @tolim, @zmanion: Please review.

tschmidtb51:
@allan-ntia: This might also be helpful for VEX.

santosomar (approved):

sthagen (approved):
LGTM.

sthagen (approved):
Good addition.

Speaking of which, I notice we mix in  a hierarchical configuration poem repetitively:
>Sets the value as given in the configuration of the program or the corresponding argument the program was invoked with. If both values are present, the program should prefer the latter one.

One aspect I find questionable is if "this" is maybe less our business in the spec and more some requirements engineering task between users and executable processors.

The other perspective that makes these two sentences stand out a bit when I look at them, is the 
1. inclusion of sources like invocation argument and configuration entry which exposes quite some detail and 
2.  exclusion of environment variables or even hard coded defaults as other typical sources for parameters 

... with the usual expected hierarchical relationships on overriding across sources.

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -1621,9 +1621,25 @@ Example:
      Siemens ProductCERT
 ```
 
-##### 3.2.1.8.5 Document Property - Publisher - Vendor ID
+##### 3.2.1.8.5 Document Property - Publisher - Namesapce
+
+The Namespace of publisher (`namesapce`) of value type `string` and format `uri` contains a URL which is under control of the issuing party and can be used as a globally unique identifier for that issuing party. The URL SHALL be normalized.
```
elear, line 1626: You should consider how this identity binds to a signing identity.  What happens when these things radically differ?  Which do you trust?  As a matter of tooling I am likely to discard this field for any automated processing unless the the document is secured in some way.

mprpic (approved):

zmanion:
> Sets the value as given in the configuration of the program or the corresponding argument the program was invoked with. If both values are present, the program should prefer the latter one.

This seems like new language used in a handful of places?

I'll say I'm confused by it and have spent ~20 minutes looking through the spec and the repo. Is "program" the program that is parsing CVRF? And the idea is that the program should use e.g., the namespace read from the CVRF data, but the command line (or maybe environment variables) overrides?

If so I'd suggest backing off to something like "Can be used by a CVRF parser" and leave the details to the implementation. Although I'm not sure even that is necessary, clearly a CVRF parser is going to use lots/all of the data it finds in a document. Or is it "name/namespace can be set here for the parsing invocation?"

Maybe this needs to be a separate issue?

csaf_2.0/prose/csaf-v2-editor-draft.md
zmanion: Isn't this true for the entire document? I vaguely recall XML DSIG was discussed but I don't see a digital signature specification/requirement for CVRF. Maybe the decision was to rely on transport layer security? See: https://github.com/oasis-tcs/csaf/issues/241.

zmanion (approved):

zmanion (approved):

tschmidtb51:
> > Sets the value as given in the configuration of the program or the corresponding argument the program was invoked with. If both values are present, the program should prefer the latter one.
> 
> This seems like new language used in a handful of places?
> 
> I'll say I'm confused by it and have spent ~20 minutes looking through the spec and the repo. Is "program" the program that is parsing CVRF? And the idea is that the program should use e.g., the namespace read from the CVRF data, but the command line (or maybe environment variables) overrides?
> 
> If so I'd suggest backing off to something like "Can be used by a CVRF parser" and leave the details to the implementation. Although I'm not sure even that is necessary, clearly a CVRF parser is going to use lots/all of the data it finds in a document. Or is it "name/namespace can be set here for the parsing invocation?"
> 
> Maybe this needs to be a separate issue?

I hope I can clarify that fairly quickly:

This sentence is from section **7.1.5 Conformance Clause 5: CVRF CSAF converter**. As we describe the program which converts a CVRF document into a valid CSAF document we should provide guidance what a tool should do with values that
either a) were not present in CVRF but are required in CSAF or b) where the semantic changed from CVRF to CSAF. Otherwise, each tool would handle them differently. IMHO, this clear guidance is needed for both: the tool developer and the user to know what is/ can be expected.

zmanion:
Sorry @tschmidtb51 I didn't read the changed file name. So, I don't have an opinion on the preference/order, I don't think I would unless I tried to implement a converter.

So, for a CVRF to CSAF conversion, the program needs to be provided the name and namespace values, since they don't exist in CVRF? I'd suggest following some UNIX/Linux convention if there is one, and using that for all cases where a conversion needs additional input.

1. command line
2. config file specified on command line
3. environment variable
4. local config
5. global config

This is a lot to specify, and would we need to name the arguments/variables/config entries?











http://docs.oasis-open.org/csaf/csaf-cvrf/v1.2/csaf-cvrf-v1.2.html#_Toc483138415

tschmidtb51:
> Sorry @tschmidtb51 I didn't read the changed file name. So, I don't have an opinion on the preference/order, I don't think I would unless I tried to implement a converter.

No worries.

> So, for a CVRF to CSAF conversion, the program needs to be provided the name and namespace values, since they don't exist in CVRF? 

Correct.

> I'd suggest following some UNIX/Linux convention if there is one, and using that for all cases where a conversion needs additional input.
> 
>     1. command line
> 
>     2. config file specified on command line
> 
>     3. environment variable
> 
>     4. local config
> 
>     5. global config
> 

I like that list. Is it exactly the UNIX/Linux convention?

> 
> This is a lot to specify, and would we need to name the arguments/variables/config entries?

IMHO, that should be left to the author of the tool. The benefit of doing so would be that it becomes easier to replace the tool. However, we would have to carefully choose those names (which might be existing entries)... Overall, I think that is more work than it is worth.

tschmidtb51:
> I'd suggest following some UNIX/Linux convention if there is one, and using that for all cases where a conversion needs additional input.

I just addressed that and added the `user config`. This should also resolve the "usual expected hierarchical relationships on overriding across sources".

----

I skipped the hard coded defaults as IMHO there is no point to claim conformance for a tool with the standard if only one can use it (esp. when it is so easy to fix as introducing a config or command line argument).

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: The real question raised here is: Do we allow the namespace to be
1. only a URL
2. only a email
3. or both?

I would like to continue this discussion at #275

sthagen (approved):
I think this chamge resolves the feedback nicely.

Still, I sense we are going down a rabbit hole with the configuration topic. And I am not even sure what to do with the rabbit when we find it … :wink:

mprpic (approved):
LGTM!

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -3616,7 +3634,16 @@ Secondly, the program for all items of:
 
 * type `/definitions/version_t`: If any element doesn't match the semantic versioning, replace the all elements of type `/definitions/version_t` with the corresponding integer version. For that, CVRF CSAF converter sorts the items of `/document/tracking/revision_history` by `number` ascending according to the rules of CVRF. Then, it replaces the value of `number` with the index number in the array (starting with 1). The value of `/document/tracking/version` is replaced by value of `number` of the corresponding revision item. The match must be calculated by the original values used in the CVRF document.
```
tolim, line 3635: This comment is for line 3615 (I cannot comment on it in this PR):
```
Secondly, the program for all items of:
```
The meaning is unclear to me. Do you mean `Secondly, the program fulfils all items of:`?

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -3616,7 +3634,16 @@ Secondly, the program for all items of:
...
+* `/document/publisher/name` and `/document/publisher/namespace`: Sets the value as given in
+
+  1. the corresponding argument from the command line,
+  2. the config file specified on the command line,
+  3. the corresponding environment variable,
+  4. the local configuration of the program,
+  5. the user configuration of the program or
+  6. the system configuration of the program.
+
+  If more than one of these value is present, the program should prefer the one with the lowest number in the list above.
```
tolim, line 3646: I do not understand why we explain possible configuration options in so much detail here. Depending on different use cases and requirements, this list might look entirely different, so this section might generate more questions than it will help.
I would suggest to shorten as follows:
```
* `/document/publisher/name` and `/document/publisher/namespace`: Both values are not contained in the right format in the original CVRF document. The converter MUST provide sensible values itself (e.g. by requesting user input).
```

tolim (commented):
Just some minor remarks.

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: It really helps interoperability a lot since you then can change from one CVRF-CSAF converter to a different one fairly quickly.

BTW: `The converter MUST provide sensible values itself (e.g. by requesting user input).` This includes hard coded values which should not be compliant.

sthagen (approved):
LGTM.

csaf_2.0/prose/csaf-v2-editor-draft.md
tolim: > It really helps interoperability a lot since you then can change from one CVRF-CSAF converter to a different one fairly quickly.

I would very much hope to be able to select one from multiple available converters. However, I doubt that we will have any positive effect by going into such implementation details.

tschmidtb51: I change the wording and remove the configuration detail.

tschmidtb51:
> Still, I sense we are going down a rabbit hole with the configuration topic. And I am not even sure what to do with the rabbit when we find it … :wink:

I removed the rabbit hole. So the rabbit is now save :wink:

sthagen (approved):
I like this simplified version better and I think it is more easily implementable in a way that fosters usability of the tools around CSAF.

tolim (approved):

santosomar (approved):
This pull requests was discussed and approved during the June 30, 2021 CSAF TC Monthly meeting. It was suggested to add example documents to support the change.

Issue #195, opened 2021-03-09
Add validation tests


tschmidtb51:
We have some parts which can't be enforced be the schema itself but should be tested programmatically. Therefore, I suggest to add those tests to the standard to ease the pain. Here are some ideas:

- Test whether definition of `product_id` exists
   - Elements of type `"#/definitions/product_id_t"` (implicit: `"#/definitions/products_t"`) which are not inside a Full Product Name (`full_product_name_t`) and therefore reference an element within the `product_tree` => test whether the Full Product Name with the `product_id` exists
   - 
      <details><summary>relevant paths: click to expand</summary><p>

      ```
      /product_tree/product_groups[]/product_ids[]
      /product_tree/relationships[]/product_reference
      /product_tree/relationships[]/relates_to_product_reference
      /vulnerabilities[]/product_status/first_affected[]
      /vulnerabilities[]/product_status/first_fixed[]
      /vulnerabilities[]/product_status/fixed[]
      /vulnerabilities[]/product_status/known_affected[]
      /vulnerabilities[]/product_status/known_not_affected[]
      /vulnerabilities[]/product_status/last_affected[]
      /vulnerabilities[]/product_status/recommended[]
      /vulnerabilities[]/product_status/under_investigation[]
      /vulnerabilities[]/remediations[]/product_ids[]
      /vulnerabilities[]/scores[]/products[]
      /vulnerabilities[]/threats[]/product_ids[]
      ```
      </p>
      </details>

- Test whether the definition of `product_id` ends up in a circle
   Example:
   ```
    "product_tree": {
        "full_product_names": [
          {
    	"product_id": "CSAFPID-0001",
    	"name": "Product A"
          }
        ],
        "relationships": [
          {
    	"full_product_name": {
    	  "name": "Product B",
    	  "product_id": "CSAFPID-0002"
    	},
    	"product_reference": "CSAFPID-0001",
    	"relates_to_product_reference": "CSAFPID-0002",
    	"relationship_type": "installed_on"
          }
        ]
      }
   ```
- Test whether definition of `group_id` exists
   - Elements of type `"#/definitions/product_group_id_t"` (implicit: `"#/definitions/product_groups_t"`) which are not inside a Product Group (`product_groups[]`) and therefore reference an element within the `product_tree` => test whether the Product Group with the `group_id` exists
   - 
      <details><summary>relevant paths: click to expand</summary><p>

      ```
      /vulnerabilities[]/remediations[]/group_ids
[...]

PR #282, opened 2021-07-09
Tests

tschmidtb51:
- addresses part of #195
- add more tests
- switch places of Tests and Profiles to avoid forward referencing
- fix section referencing
- rephrase conformance target "CSAF producer" to use definition of "CSAF document"
- rephrase conformance target "CSAF document" to allow data streams

### DIFF
--- a/csaf_2.0/prose/csaf-v2-editor-draft.md
+++ b/csaf_2.0/prose/csaf-v2-editor-draft.md
@@ -328,6 +328,8 @@ _GTLS/SSL and crypto library_, OpenSSL Software Foundation, https://www.openssl.
 _Package URL (PURL)_, GitHub Project, https://github.com/package-url/purl-spec
 ###### [RFC3552]
 Rescorla, E. and B. Korver, "Guidelines for Writing RFC Text on Security Considerations", BCP 72, RFC 3552, DOI 10.17487/RFC3552, July 2003, https://www.rfc-editor.org/info/rfc3552.
+###### [RFC7231]
+Fielding, R. and J. Reschke, "Hypertext Transfer Protocol (HTTP/1.1): Semantics and Content", RFC 7231, DOI 10.17487/RFC7231, June 2014, <https://www.rfc-editor.org/info/rfc7231>.
 ###### [RFC7464]
 N. Williams., "JavaScript Object Notation (JSON) Text Sequences", RFC 7464, DOI 10.17487/RFC7464, February 2015, http://www.rfc-editor.org/info/rfc7464.
 ###### [SCAP12]
@@ -4069,9 +4071,9 @@ Example which fails the test:
 
 ### 5.3.6 Use of non-self referencing URLs Failing to Resolve
 
-For each URL which is not in the category `self` it must be tested that it resolves with a HTTP status code less than 400.
+For each URL which is not in the category `self` it must be tested that it resolves with a HTTP status code from the 2xx (Successful) or 3xx (Redirection) class.
 
-> This test does not apply for any item in an array of type `references_t` with the category `self`.
+> This test does not apply for any item in an array of type `references_t` with the category `self`. For details about the HTTP status code classes see [RFC7231].
 
 The relevant paths for this test are:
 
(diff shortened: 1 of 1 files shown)


csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -2932,6 +2934,8 @@ Example which fails the test:
 
 > The required element `baseSeverity` is missing.
 
+> A tool MAY add one or more of the missing properties `version`, `baseScore` and `baseSeverity` based on the values given in `vectorString` as quick fix.
```
sthagen, line 2937: The prior wording `implement add one or more of …` read like copy-pasta (which I am sure it was not, but …) 
A sentence with a `MAY` Is often hard to read especially when additional layers of `potential only application`  join that keyword.
My understanding is, that in this and similar phrases about tools (proxy actors) a capability of the proxy is given permission to exist per this spec but that these phrases IMO try to achieve too much.
I propose to rely on the fearsome power of the `MAY` and state the possible actions without any `implements` addition (just removing the word).

sthagen (approved):
Thank you so much for the many added tests and the clarity we win.

My comment on the `MAY` sentences hopefully helps to remove the remaining `implements` Indirections which in turn will ease the reading without impacting the precision.

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: > The prior wording `implement add one or more of …` read like copy-pasta 

You are completely right.

> (which I am sure it was not, but …)

Sorry about that - but it looks like it was :wink: 

> I propose to rely on the fearsome power of the `MAY` and state the possible actions without any `implements` addition (just removing the word).

Thank you for the suggestion. It makes it much easier to read. There was only one (`implement the recommendation`) which I was not sure how to replace that one... Do you have an idea or is that fine as it doesn't have multiple layers?

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -4058,15 +4503,15 @@ The entities ("conformance targets") for which this document defines requirement
...
+* satisfies at least one profile defined in section 4.
+* does not fail any mandatory test defined in section 5.1.
 
 ### 8.1.2 Conformance Clause 2: CSAF producer
 
 A program satisfies the "CSAF producer" conformance profile if the program:
 
-* produces output in the CSAF format, according to the syntax and semantics defined in section 3. The output MUST not fail any mandatory test defined in section 4.1.
-* satisfies those normative requirements in section 3 and 6 that are designated as applying to CSAF producers.
+* produces output in the CSAF format, according to the syntax and semantics defined in section 3. The output MUST satisfies at least one profile defined in section 4 and MUST NOT not fail any mandatory test defined in section 5.1.
```
tschmidtb51, line 4513: @sthagen I was wondering whether we could reference Conformance Clause 1 here. I guess that would be easier to read.

sthagen: Back referencing a conformance clause inside the conformance section itself is very good service to the reader and IMO fully compatible with any OASIS expectation on the conformance section. Please go ahead with that approach. Thanks for identifying. Does that mean in this case that conformance clause 2 extends clause 1?

csaf_2.0/prose/csaf-v2-editor-draft.md
sthagen: W.r.t. an enhancement for `implement the recommendation` I suggest to de-noun-ify and try to form a statement starting with `MAY recommend …` taking the perspective of the tool changing, or if the `recommendation` Is a jungle of interwoven statements, maybe flatten into a list of possible actions.

If the `recommendation` Is not actionable, this will not work, but then how shall the tool implement?

I hope this helps.

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: What about the following:

```suggestion
* produces output in the CSAF format, according to the conformance profile "CSAF document" .
```

However, I guess we should remove the text file part in conformance clause 1.

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -4058,15 +4503,15 @@ The entities ("conformance targets") for which this document defines requirement
 A text file satisfies the "CSAF document" conformance profile if the text file:
```
tschmidtb51, line 4503: Related to ???:

```suggestion
A text file or data stream satisfies the "CSAF document" conformance profile if it:
```

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -4007,7 +4007,7 @@ Example which fails the test:
 
 ### 4.3.7 Use of self referencing URLs Failing to Resolve
 
-For each item in an array of type `references_t` with the category `self` it must be tested that the URL referenced resolves with a HTTP status code less than 400. 
+For each item in an array of type `references_t` with the category `self` it must be tested that the URL referenced resolves with a HTTP status code less than 400.
```
sthagen, line 4115: Hm, I would trade the elegance of the less than operator for the clarity of excluding the informative HTTP response codes of the 100 family. I think we want to say that the URL is in fact dereference-able. Maybe a `Continue`, `Processing`, or similar is not helpful.

sthagen (approved):
LGTM.

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: Sure. Possibly solved in 531e172.

sthagen (approved):

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: Possibly resolved in f67d3b4.

sthagen (approved):

santosomar (approved):
Great job!

tschmidtb51:
@sthagen Thaks for spotting 69c8cef.

PR #287, opened 2021-07-10
Distributing

### LINKED ISSUE #272: Requirement 3: TLS
From the requirements section:
> ### 5.1.3 Requirement 3: TLS
> 
> The CSAF document is retrievable from a website which uses TLS for encryption and server authenticity. The CSAF document MUST not be downloadable from a location which does not encrypt the transport.

I like that TLS is required here. Two points as comments:
- The requirement emphasizes encryption for data protection. However, most of the published advisories are already public so encryption is not really required. We need to ensure authenticity of the server and data integrity for the document download.
- Advisories may be located in local unencrypted network shares within organizations. This implementation may be perfectly adequate from a security point of view. So we should avoid defining that the document MUST NOT be downloadable from a non-encrypting location. 

This would be my suggestion:
`The organization's default website offering CSAF documents MUST provide data protection based on TLS to ensure server authenticity and data integrity.`

tschmidtb51:
- resolves #272
- addresses parts of #152 
- add .well-known as an option
- rearrange order to ease understanding

### DIFF
--- a/csaf_2.0/prose/csaf-v2-editor-draft.md
+++ b/csaf_2.0/prose/csaf-v2-editor-draft.md
@@ -4295,15 +4295,31 @@ CSAF: https://psirt.domain.tld/advisories/csaf/
 CSAF: https://domain.tld/security/csaf/csaf-service.json
 ```
 
-### 6.1.8 Requirement 8: Well-known path
+### 6.1.8 Requirement 8: Well-known URL for Service Document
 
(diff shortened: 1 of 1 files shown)


sthagen (approved):
LGTM - fixed a typo (occurring twice) and amended the rationale for the transport protocol only redirect requirement.

sthagen (approved):

tschmidtb51:
@santosomar Ready for merge.

PR #288, opened 2021-07-10
Additional conventions

### LINKED ISSUE #227: Proposal for acknowledgments attributes to be single values instead of arrays

### LINKED ISSUE #277: Find a better place to specify the filename rules
@tolim commented in #256: 

We specify the filename of our CSAF document in a quite hidden section under Document Property - Tracking - ID. I would not have expected that as a reader. Can we find some other, more adequate place for this specification so that it can also be found in the TOC?

tschmidtb51:
- resolves #277  
- add extra section "Additional conventions"
- move filename rules to new section
- add reference to RFC7464
- add remark about sorting
- add test for sorting

### DIFF
--- a/csaf_2.0/prose/csaf-v2-editor-draft.md
+++ b/csaf_2.0/prose/csaf-v2-editor-draft.md
@@ -3957,6 +3957,30 @@ Example which fails the test:
 
 > The only element where the `category` is `self` has a URL that does not fulfill the requirement of a valid filename for a CSAF document.
 
+### 6.2.12 Sorting
+
+It must be tested that all keys in a CSAF document are sorted alphabetically.
+
+The relevant path for this test is:
+
+```
+  /
+```
+
+Example which fails the test:
+
+```
+  "document": {
+    "csaf_version": "2.0",
+    "category": "generic_csaf",
+    // ...
+  }
+```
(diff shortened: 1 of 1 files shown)


sthagen (approved):
LGTM - I think the addition of the conventions section was long overdue and esp. the filename rules now have a more prominent place. Thanks.

santosomar (approved):
I agree with Stefan. Thank you!

PR #296, opened 2021-07-16
Alternative to JSON Schema 2020-12 #294

sthagen:
Technically I did not find a way to push to the existing JSON Schema 2020-12 #294 PR and thus created a branch on this repo with the changes from @tschmidtb51 `json-2020-12` branch and applied two minimally invasive fixes:

1. provide `git` to the Debian image
2. added the proposed install steps for the intermediate validator - thanks to @nezhar (Harald Nezbeda)

I propose to merge this PR instead of #294 - closes #92

Issue #92, opened 2020-07-04
What jsonschema version to use


sthagen:
Given the fact, that draft 7 of jsonschema did expire in March 2020 we should analyze which version to settle on.

tolim:
The latest version of JSON schema is 2019-09 (https://json-schema.org/draft/2019-09/release-notes.html). This is one release after draft 7, so we were only slightly outdated :-)
I propose to always settle on the latest version until we finalize our spec.

sthagen:
I am in favor of following the components development we rely on - but I will investgate the practical aspect of validator implementations following the draft (or not) and possibly exposing errors in the new "dialects" they offer. If these validators are of equal quality I have no problem in us profiting from the enhancements of new json schema draft verisons.

Another aspect is the support for tool builders and API consumers: OpenAPI/Swagger/BakeMyAppFromThat-Service-providers are also an important aspect to include in the analysis. If draft-7 is good later this year for these interface builders but not the fresh stuff we may profit from staying in the draft-7 regime and offer easily well known tooling helpers for our consumers.

tschmidtb51:
I just had a look into the library we use for our tests: jsonschema has an open issue to support 2019-09: Julian/jsonschema#613

I suggest to revisit the issue before we publish to see the developments on the market. @sthagen mentioned a big point: support for tool builders and API consumers. If the users need to implement the JSON validator themselves instead of including a library - most of them won't do that. This would result in advisories which are incompatible with the schema.

sthagen:
Updates from the JSON schema ecosystem: 
* The JSON Schema community published on 2021-02-01 the new draft version 2020-12:
  * JSON Schema Core (2020-12) at https://json-schema.org/draft/2020-12/json-schema-core.html
  * JSON Schema Validation (2020-12) at https://json-schema.org/draft/2020-12/json-schema-validation.html
  * Relative JSON Pointers (2020-12) at https://json-schema.org/draft/2020-12/relative-json-pointer.html
* OpenAPI published version 3.1 on 2021-02-18 (cf. https://www.openapis.org/blog/2021/02/18/openapi-specification-3-1-released) and the exciting news is (citation from ibid):
  >This new version now supports 100% compatibility with the latest draft (2020-12) of JSON Schema. 
* Tools:
  * python jsonschema package has a ticket requesting support for 2020-12 draft at https://github.com/Julian/jsonschema/issues/782
[...]

PR #297, opened 2021-07-16
JSON Schema 2020-12 (update prose)

tschmidtb51:
- addresses part of oasis-tcs/csaf#92 and oasis-tcs/csaf#296
- adopt prose to match schema


sthagen (approved):
Very good to see the draft 2019 go away and esp. to see how few places need change to be JSON Schema draft 2020-12 compliant. Great work @tschmidtb51 - thanks a lot!

PR #298, opened 2021-07-16
PURL

### LINKED ISSUE #126: Create proposal to add purl, and swid properties to full_product_name_t

tschmidtb51:
- resolves #126 
- remove comment
- change regex
- adopt prose


tschmidtb51:
@pombredanne Please review that change concerning PURL representation in CSAF.

sthagen (approved):
LGTM a partial regex - we should reach out to
@pombredanne (Philippe Ombredanne), @ stevespringett (Steve Springett), or other members of the purl-spec community, if this structural constraint collides with the future plans for purl or if it looks ok.

tschmidtb51:
@stevespringett, @pombredanne: Do you have any additional comments or can we proceed with that PR?

FYI: The PR provides a regex in addition to the URL check:
"^pkg:[A-Za-z\\.\\-\\+][A-Za-z0-9\\.\\-\\+]*/.+" 

I think that captures enough of the structure and ensures the canonical form of purl but leaves plenty of room to be future prove.

stevespringett:
@tschmidtb51 Does the regex take into consideration percent encoded strings?

tschmidtb51:
@stevespringett To be honest: No. But that is not necessary:

- `^pkg:` matches the **scheme**
- `[A-Za-z\\.\\-\\+][A-Za-z0-9\\.\\-\\+]*` matches the **type**. According to the [spec](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#rules-for-each-purl-component):
   - The package type is composed only of ASCII letters and numbers, '.', '+' and '-' (period, plus, and dash)
   - The type cannot start with a number
   - The type cannot contains spaces
   - *The type must NOT be percent-encoded*
- `/.+` matches either the **name** or the **namespace**. According to the [spec](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#rules-for-each-purl-component) each of them *must be a percent-encoded string*. But `/.+` just says that there must be any character after the `/`.

It certainly matches a things that are not valid PURLs (e.g. `pkg:maven/@1.3.4`) but it should give enough guidance to enforce valid URI/URL per [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3). For the rest, dedicated libraries SHOULD be used. That allows CSAF still to be valid if the PURL spec is updated (as long as the rules for **type** and the separator `/` do not change).

tschmidtb51:
@santosomar, @sthagen We did not received feedback from @pombredanne (in #126) and @stevespringett. We replied to both without changes (as it was already covered by the suggested regex). We did not receive further feedback to this. Therefore, I suggest to merge it (to provide a full CSD for review) and point it out in the TC meeting.

@stevespringett, @pombredanne If you want to provide additional comments - please feel free to do so.

stevespringett:
I think this approach is solid and future-proof. No further comments from me.

santosomar:
> I think this approach is solid and future-proof. No further comments from me.

Thank you so much @stevespringett !

santosomar (approved):

PR #300, opened 2021-07-17
Sort examples

### LINKED ISSUE #264: Sort examples
We should make sure that examples are sorted.

sthagen:
This change set closes #264 

In addition, I updated a 2019 year mention in an example to 2021 ...

### DIFF
--- a/csaf_2.0/json_schema/csaf_json_schema.json
+++ b/csaf_2.0/json_schema/csaf_json_schema.json
@@ -27,8 +27,8 @@
               "type": "string",
               "minLength": 1,
               "examples": [
-                "Johann Sebastian Bach",
-                "Albert Einstein"
+                "Albert Einstein",
+                "Johann Sebastian Bach"
               ]
             }
           },
@@ -39,8 +39,8 @@
             "minLength": 1,
             "examples": [
               "CISA",
-              "Talos",
-              "Google Project Zero"
+              "Google Project Zero",
+              "Talos"
             ]
           },
           "summary": {
@@ -110,14 +110,14 @@
             "type": "string",
             "minLength": 1,
             "examples": [
-              "Microsoft",
-              "Siemens",
-              "Windows",
-              "Office",
-              "SIMATIC",
               "10",
               "365",
-              "PCS 7"
+              "Microsoft",
+              "Office",
+              "PCS 7",
+              "SIMATIC",
+              "Siemens",
+              "Windows"
             ]
           },
           "product": {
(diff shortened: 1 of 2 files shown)


csaf_2.0/json_schema/csaf_json_schema.json
```diff
@@ -974,9 +974,9 @@
                 "type": "string",
                 "pattern": "^CWE-[1-9]\\d{0,5}$",
                 "examples": [
-                  "CWE-79",
                   "CWE-22",
-                  "CWE-352"
+                  "CWE-352",
+                  "CWE-79"
```
tschmidtb51, line 979: Shouldn't that be:
```suggestion
                  "CWE-22",
                  "CWE-79",
                  "CWE-352"
```

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -2135,19 +2135,19 @@ and holds the ID for the weakness associated.
 Examples:
 
 ```
-    CWE-79
     CWE-22
     CWE-352
+    CWE-79
```
tschmidtb51, line 2140: I guess, that should also be:
```suggestion
    CWE-22
    CWE-79
    CWE-352
```

tschmidtb51 (approved):
In general, approve. 2 minor comments...

csaf_2.0/json_schema/csaf_json_schema.json
sthagen: Well, we order lexically. The prefix makes the 7 in the CWE-79 appear after the CWE-352 due to the leading 3 of the latter number part. Personally I think CWE numbers have long fallen victim to the _initial plan_ Syndrome, where at start you think you know the terrain and set numbers, and then life goes on and all the balancing is gone as well as some of these well meant advice documents that require deletion as time goes by.

csaf_2.0/prose/csaf-v2-editor-draft.md
sthagen: The user interface of PRs on sandboxed tablets is kind of multiplicative it seems. Per [the „above“ comment](https://github.com/oasis-tcs/csaf/pull/300#discussion_r671796036) I would appreciate, if we do not make a special case for CWEs - esp. because In my experience the lower numbered CWEs are even more unuseful in day to day developer support than the higher numbered ones 🏃🏽‍♂️ 🧯

csaf_2.0/json_schema/csaf_json_schema.json
tschmidtb51: > Well, we order lexically. The prefix makes the 7 in the CWE-79 appear after the CWE-352 due to the leading 3 of the latter number part.

Makes totally sense. I agree. I withdraw the comment ;-)

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: As above: I withdraw the comment ;-)

tschmidtb51 (approved):
IMHO: Ready for merge.

PR #302, opened 2021-07-17
Sort enum values

### LINKED ISSUE #301: Sort enumeration value listings
Some enumeration value listings in the schema (and of course) prose still do not follow a lexical ordering.

I propose we sort the enum values to make clear, that the order of those values has no further semantics.

Later additions of the spec should be empowered to add (possible) or remove (improbable) such catalog values and then simply resort the list.

Any implementation is well advised to not rely on some numerical ordering of an underlying representation like the index number of the JSON array we use to denote the values in the schema.

sthagen:
Small and scoped changes:
* Sorted enum values for three types
* kept the description order for restart category in prose as is because of a natural ordering

Closes #301

**Note**: 
* This change set contains all changes from #300 as I directly branched off of that feature branch.
* For review of only the enum order changes you may review the commit [5a9fb990](https://github.com/oasis-tcs/csaf/pull/302/commits/5a9fb99017b0e2b15b9f7d90637bb122ebce5d63)

No merge conflict expected when merging this PR after #300

### DIFF
--- a/csaf_2.0/json_schema/csaf_json_schema.json
+++ b/csaf_2.0/json_schema/csaf_json_schema.json
@@ -576,9 +576,9 @@
                   "description": "Provides the TLP label of the document.",
                   "type": "string",
                   "enum": [
-                    "RED",
                     "AMBER",
                     "GREEN",
+                    "RED",
                     "WHITE"
                   ]
                 },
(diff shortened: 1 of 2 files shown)


tschmidtb51 (approved):
In general, I approve these changes. The same comments (https://github.com/oasis-tcs/csaf/pull/300#discussion_r671747238, https://github.com/oasis-tcs/csaf/pull/300#discussion_r671747453) apply here.

Really important to me is that the order of the explanation of the values in the prose is not changed as it carried the semantics.

tschmidtb51 (approved):
IMHO: Ready for merge.

santosomar:
> IMHO: Ready for merge.

+1

santosomar (approved):
Thank you again!

Issue #304, opened 2021-07-20
Enumerate examples in prose


sthagen:
To ease publication and support reader comments we should enumerate the examples in the prose document (the blocks).

tschmidtb51:
This was solved with #305. @sthagen Thank you for the addition.

PR #305, opened 2021-07-20
Enumerate examples in prose

sthagen:
Enumerated all 104 examples (numbers 99 and 101 for ROLIE service and category doc are TODO.

This PR solves #304

PS: Updated log date, title date, and fixed a typo.

### DIFF
--- a/csaf_2.0/prose/csaf-v2-editor-draft.md
+++ b/csaf_2.0/prose/csaf-v2-editor-draft.md
@@ -556,7 +556,7 @@ The properties are: `names`, `organization`, `summary`, and `urls`.
 List of acknowledged names (`names`) has value type `array` with 1 or more items holds the names of entities being recognized.
 Every such item of value type `string` with 1 or more characters represents the name of entity being recognized and contains the name of a single person.
 
-*Examples 1*:
+*Examples 1:*
 
 ```
     Albert Einstein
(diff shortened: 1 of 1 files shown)


tschmidtb51 (commented):
Thanks for adding these changes. 

As it is currently inconsistent, I have 2 questions:
1. Should the `:`inside or outside the `*[...]*`? E.g. `*Examples 42:*` or `*Examples 1*:`
2. How do we treat plural? Do we use "Examples" when we have multiple and "Example" when we have one?

sthagen:
I did these 104 changes by hand, so I see no inconsistency but mistakes I made - I suggest the italic format to bracket the whole content so outside the colon - any asterisk preceding the colon is a typo. The singular plural is by choice. I see an enumeration as an isolated tag and do not need a constant prefix. So if there are examples than it would be `*Examples 42 ...` and if only a simple example then `*Example 42 ...` (I left all nouns as they were, so there may be a mismatch between the multiplicity of the example term and the actual number of examples present in the block.

sthagen:
The amended PR should now be consistent within the flexibility we offer for the singular vs. plural depending on the content.

Note, that the source format markdown eases automatic extraction of the examples per start regex token `^\ *\*Example` & end token the second `^\ *```aaa` following a start token.

santosomar (approved):
I like the way you are allowing for these examples and completely agree to have the ability to include multiple (plural). Thank you all!!

Issue #223, opened 2021-04-13
Add missing generic keywords in schema


tschmidtb51:
At the moment, we don't have the generic keywords `title` and `description` at all places where they are needed. For example, they are present at:
- `/document/lang`
- `/document/notes`
- `/document/source_lang`
- `/vulnerabilities[]/product_status/first_affected`
- `/vulnerabilities[]/product_status/first_fixed`
- `/vulnerabilities[]/product_status/fixed`
- `/vulnerabilities[]/product_status/known_affected`
- `/vulnerabilities[]/product_status/known_not_affected`
- `/vulnerabilities[]/product_status/last_affected`

However, they are missing at:
- `/document/acknowledgments`
- `/document/references`
- `/product_tree/relationships[]/product_reference`
- `/product_tree/relationships[]/relates_to_product_reference`
- `/vulnerabilities[]/acknowledgments`
- `/vulnerabilities[]/notes`
- `/vulnerabilities[]/references`

Missing, but IMHO not (necessarily) needed at:
- `/definitions/branches_t[]/branches`
- `/definitions/branches_t[]/product`
-  `/definitions/full_product_name/product_id`
-  `/definitions/product_groups_t[]`
- `/definitions/products_t[]`
- `/document/tracking/revision_history[]/version`
- `/document/tracking/version`
- `/product_tree/branches`
- `/product_tree/full_product_names`
- `/product_tree/product_groups[]/group_id`
- `/product_tree/product_groups[]/product_ids`
- `/product_tree/relationships[]/full_product_name`
- `/vulnerabilities[]/remediations[]/group_ids`
- `/vulnerabilities[]/remediations[]/product_ids`
- `/vulnerabilities[]/scores[]/products`
- `/vulnerabilities[]/threats[]/group_ids`
- `/vulnerabilities[]/threats[]/product_ids`

tschmidtb51:
This is also related to the review comment from @tolim at https://github.com/oasis-tcs/csaf/pull/205#discussion_r611959017.

tschmidtb51:
This was resolved with #308.

PR #308, opened 2021-07-22
Generic keywords

tschmidtb51:
- addresses parts of #223 
- add missing (necessary) generic keywords
- changes for consistency
- minor format changes

### DIFF
--- a/csaf_2.0/json_schema/csaf_json_schema.json
+++ b/csaf_2.0/json_schema/csaf_json_schema.json
@@ -670,6 +670,8 @@
           }
         },
         "references": {
+          "title": "Document references",
+          "description": "Holds a list of references associated with the whole document.",
           "$ref": "#/$defs/references_t"
         },
         "source_lang": {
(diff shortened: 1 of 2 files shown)


csaf_2.0/json_schema/csaf_json_schema.json
```diff
@@ -951,6 +958,8 @@
         "minProperties": 1,
         "properties": {
           "acknowledgments": {
+            "title": "Vulnerability acknowledgments",
+            "description": "Contains a list of acknowledgment elements associated with this vulnerability item.",
```
tschmidtb51, line 962: @sthagen Should we use `associated with the current vulnerability item.` instead?

sthagen: Good question. The entry in question is part of a vulnerability object being an item in the array of all vulnerabilities of the advisory / CSAF document. The term `this` Is even used in some funny programming languages as reference to such an entity, others use `self` Per convention but allow any valid identifier if the programmer insists. I would keep the `this` As the `current` may provide to many notions like an `iterator like concept` Or even a `temporal aspect` Both of which we do not want to convey.

sthagen: 👍🏽

sthagen (approved):

santosomar (approved):

Issue #38, opened 2020-06-02
Add automated tests to repository


tolim:
I suggest to add automated tests into the Github repository using Github actions. Triggered by pushed commits into the repository, the tests would automatically be executed and show the result on the repository's web site. If a test failed, an email would automatically be sent to the committer.

The following tests could be integrated:
- [x] Check syntax of CSAF JSON schema
- [x] Check syntax of CSAF examples (CVE-2018-0171-modified.json and cvrf-rhba-2018-0489-modified.json)
- [x] Check validity of CSAF examples against JSON schema
- [x] Check validity of CSAF examples against modified version of our JSON schema to enable strict checking (add `"additionalProperties": false` in each object)
- [ ] Perform automated conversion of the original CVRF 1.2 examples into CSAF and check their validity

The already implemented tests are available in our forked repository, please see an example here: https://github.com/tschmidtb51/csaf/runs/720620127?check_suite_focus=true)

I have the following questions:
- What do you think? As a next step, I would create a pull request using the current tests for this repository.
- What additional tests would make sense?

Thanks!

santosomar:
That makes a lot of sense. Perhaps others in the TC also can help. Thank you so much for your contribution!

tolim:
I just added a pull request #59 that will include the automated tests in our master branch.

tschmidtb51:
IMHO, the test data for the conversion should live in a separate repo at https://github.com/oasis-open.

tschmidtb51:
Therefore, I suggest to close the issue.

santosomar:
I completely agree. Overall, while there has been significant progress in addressing this issue through a few contributions and commits, it seems that discussions have dwindled, and the issue can be closed.

PR #310, opened 2021-07-22
Provider metadata

tschmidtb51:
- addresses part of #290, #152
- correct mistake in provider metadata schema

PR #311, opened 2021-07-22
CSAF

tschmidtb51:
- addresses part of oasis-tcs/csaf#38
- correct mistake in schema found with ajv

### DIFF
--- a/csaf_2.0/json_schema/csaf_json_schema.json
+++ b/csaf_2.0/json_schema/csaf_json_schema.json
@@ -1250,7 +1250,7 @@
               "title": "Score",
               "description": "specifies information about (at least one) score of the vulnerability and for which products the given value applies.",
               "type": "object",
-              "minItems": 2,
+              "minProperties": 2,
               "required": [
                 "products"
               ],


csaf_2.0/json_schema/csaf_json_schema.json
```diff
@@ -1250,7 +1250,7 @@
               "title": "Score",
               "description": "specifies information about (at least one) score of the vulnerability and for which products the given value applies.",
               "type": "object",
-              "minItems": 2,
+              "minProperties": 2,
```
sthagen, line 1253: 👀

sthagen (approved):
Good catch.

csaf_2.0/json_schema/csaf_json_schema.json
santosomar: Oh wow! Great catch indeed! Thank you so much!

santosomar (approved):

csaf_2.0/json_schema/csaf_json_schema.json
tschmidtb51: It was actually searching for the mistake I (then) corrected in #310... This was more accidentally.

Issue #309, opened 2021-07-22
Test other schemas


tschmidtb51:
We should also test the aggregator schema and the provider metadata schema.

PR #312, opened 2021-07-23
Namespace upgrade

### LINKED ISSUE #68: CSAF $id: link to OASIS namespacse
As discussed during the June 24, 2020 monthly meeting this is a reminder to change the URL to the official https://docs.oasis-open.org/csaf/ns/ namespace once the schema becomes official.

sthagen:
Added some plausible target namespace URLs for all three schema files - final URLs will be set collaborating with TC administration.

Closes #68.


tschmidtb51 (approved):
Hmm. This will likely break my approach to #309 :disappointed: ... But you are totally right!

santosomar (approved):
@sthagen @tschmidtb51  Looks good, but quick question:  since I own the CSAF.io domain, would you like to simplify the namespace URI?

tschmidtb51:
@sthagen We also need that change for example-01-feed-tlp-white.json

tschmidtb51:
> @sthagen @tschmidtb51 Looks good, but quick question: since I own the CSAF.io domain, would you like to simplify the namespace URI?

No. We should stick to the official OASIS namespace.

But I would like to use `https://csaf.io` as `namespace` for the examples we provide in https://github.com/oasis-tcs/csaf/tree/master/csaf_2.0/test/validator/data

santosomar:
Sounds good. Thank you!

PR #315, opened 2021-07-23
Rolie examples

### LINKED ISSUE #313: Add ROLIE examples
We should provide examples for the ROLIE documents as requested in https://github.com/oasis-tcs/csaf/pull/269#issuecomment-871596061.

tschmidtb51:
- resolves #313 
- add examples for ROLIE service document
- add example for ROLIE category document
- update example for ROLIE feed document:
  - update schema (as done with #312)
  - update category to hold more than one (see [RFC8322](https://datatracker.ietf.org/doc/html/rfc8322#section-6.1.1))


tschmidtb51:
@mprpic Please review this PR with ROLIE examples.

sthagen (approved):

tschmidtb51:
@santosomar, @sthagen We did not receive additional feedback to this. Therefore, I suggest to merge it and point it out in the TC meeting.

csaf_2.0/examples/ROLIE/example-01-category.json
```diff
@@ -0,0 +1,10 @@
+{
+  "categories": [
+    {
+      "term": "Example Company Product A"
+    },
+    {
+      "term": "Example Company Product B"
+    }
+  ]
+}
```
mprpic, line 10: I think this also need to specify a scheme to which these categories relate. Though, I'm not sure what the value of that field should be, so:

```json
{
   "categories": {
      "category": [
         {
            "term": "Example Company Product A"
         },
         {
            "term": "Example Company Product B"
         }
      ],
      "scheme": "<reference to itself?>"
   }
}
```

Perhaps it can just be an arbitrary URN that gets then used in the individual documents. I think only terms specific to the `urn:ietf:params:rolie:category:information-type` category don't need to be explicitly noted in the category document.

For now, I'd recommend using a self-reference to the category document itself as the scheme.

mprpic (commented):
Looks great!

csaf_2.0/examples/ROLIE/example-01-category.json
tschmidtb51: > I think this also need to specify a scheme to which these categories relate.

I agree that there SHOULD be a category scheme. However, while reading [RFC 5023 Section 7](https://datatracker.ietf.org/doc/html/rfc5023#section-7) I did not see that it is mandatory
```
    appInlineCategories =
       element app:categories {
           attribute fixed { "yes" | "no" }?,
           attribute scheme { atomURI }?,
           (atomCategory*,
           undefinedContent)
       }
```

Therefore, I skipped it. But I might have overlooked something. Can you please double-check that?

If it is necessary, I second your approach to app the reference to itself in the document.

tschmidtb51: Nevertheless, you are right about mentioning of `category`.

tschmidtb51: I fixed that in f44b6cb.

sthagen (approved):

csaf_2.0/examples/ROLIE/example-01-category.json
mprpic: You're right, it is optional. Let's leave it out. Thanks for the `category` fix!

mprpic (approved):

csaf_2.0/examples/ROLIE/example-01-category.json
tschmidtb51: No worries. I also fixed that in the service document examples and fixed a filename error.

Issue #27, opened 2020-04-30
Update examples for each of the elements


santosomar:
Examples should be included for each objects/elements in the CSAF 2.0 schema documentation.

tschmidtb51:
@sthagen I found two other examples which where not numbered yet. See 35eaea6.

tschmidtb51:
@sthagen I guess the best approach to correct the numbering is to wait until #321 and #320 have been merged. Otherwise, there will be merge conflicts.

PR #320, opened 2021-07-23
Tests

tschmidtb51:
- resolves #195 
- add test for document language
- add tests from prose as separate files (optional and informative)

PR #321, opened 2021-07-23
Example

tschmidtb51:
- addresses part of oasis-tcs/csaf#27
- add example for relationship

@santosomar I hope you are okay with that example. If not I could change it to a generic `Product A` and `Product B`. However, IMHO the current suggestion really support the understanding of the reader.


tschmidtb51:
@sthagen I did not update the Example numbering as you wanted to correct that globally. (Also #320 would affect that.)

santosomar (approved):
You are awesome!! I like it ;-)

sthagen (approved):

PR #322, opened 2021-07-23
Remove todos

tschmidtb51:
- removes addressed TODOs
- correct fingerprint length


santosomar (approved):
Getting closer!!! ;-)

sthagen (approved):
I noticed the minimal fingerprint length for the PGP key enlarged by 8 bytes - I understand this as hardening the hash.

PR #324, opened 2021-07-24
CPE

### LINKED ISSUE #91: CPE regex

tschmidtb51:
- resolves #91 
- add test for CPE regex against dictionary


csaf_2.0/test/cpe/run_tests.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
```
sthagen, line 2: In my own daily work, I tend to have a `# synopsis like sigle line comment always on line 2 ... should we do this here too?

.gitignore
```diff
@@ -2,3 +2,5 @@
 meeting_minutes/.DS_Store
 .DS_Store
 *_strict_schema.json
+official-cpe-dictionary_v2.3.*
```
sthagen, line 5: I can imagine a single `official-cpe-dictionary_v2*` would do but I am `+0` on just keeping your version - after all it is only a git ignore file.

sthagen (approved):
I changed the driver and the validator without testing and hope the testing happens within this PR mechanics ... hope helps 😁 - it worked (I usually do not change that much as reviewer ... esp. without prior verification).

Note:  I have no problem when my commits are reverted (I approve the PR never the less).

Let us merge this fast, it is a very good contribution I think. Thank you so much @tschmidtb51 !

csaf_2.0/test/cpe/run_tests.sh
tschmidtb51: Feel free to do so.

.gitignore
tschmidtb51: I tend to be a bit more specific in `.gitignore` as it helps me to avoid the "works on my machine". But I'm happy with both variants...

santosomar:
Let the tests start 😜 I will merge now.

PR #326, opened 2021-07-24
[prose] Removed unused conventions and updated to current date

### LINKED ISSUE #306: Remove unused elements from typography section
In case we do not refer from conformance section back to the preceding sections on sentence / statement level we can drop the special convention we used in the v1.2 documents.

This operation should be done short before publication when we can guarantee that the conformance section is self-contained or references exclusively to role names and section level elements.

The latter is plausible, as we define tests inside the specification in a dedicated section with "atomic" subsections.

sthagen:
Closes #306


tschmidtb51 (approved):
LGTM.

santosomar (approved):
Thank you!! Looks good.

PR #364, opened 2021-09-29
Cs01 cand incl non material csdpr01 feedback

sthagen:
This pull request provides the opportunity for all TC members to revisit the candidate revision for submitting CSAF CS01 as a complete end result (not for tracing the changes back to feedback from public review to CsD01 which the prior PR towards the [cs01-cand-incl-non-material-csdpr01-feedback  branch](https://github.com/oasis-tcs/csaf/compare/master...cs01-cand-incl-non-material-csdpr01-feedback) focused on.

Please kindly read this version of the document and schema files so we can ensure a best quality CS01 document for submittal.

I hereby move to merge this  PR into master branch including any changes members suggest after one week.

On my calendar this is October 6, 2021 at or after 18:10 UTC

### DIFF
--- a/csaf_2.0/prose/csaf-v2-editor-draft.md
+++ b/csaf_2.0/prose/csaf-v2-editor-draft.md
@@ -1,37 +1,54 @@
 
-![OASIS Logo](http://docs.oasis-open.org/templates/OASISLogo-v2.0.jpg)
+![OASIS Logo](https://docs.oasis-open.org/templates/OASISLogo-v3.0.png)
+
 -------
 
 # Common Security Advisory Framework Version 2.0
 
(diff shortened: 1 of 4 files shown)


santosomar (approved):
Thank you @sthagen 🙏

sthagen:
@santosomar I suggest to not rush with merging as to give all reviewers extra time of say one day to provide feedback the CS01 candidate may profit from.

tschmidtb51:
@sthagen, @santosomar: I stumbled across a questionregarding 6.1.21:
> Is there any reason why /use case where a revision history could start with 2 or 4 or 17? If we don't come across such a case I suggest to make that explicit.

PR #365, opened 2021-10-07
Address unclear start of revision history

tschmidtb51:
- resolves https://github.com/oasis-tcs/csaf/pull/364#issuecomment-937285664
- explicitly state that the first item of revision history sorted by date must be 0 or 1 (or equivalent major version)

### DIFF
--- a/csaf_2.0/prose/csaf-v2-editor-draft.md
+++ b/csaf_2.0/prose/csaf-v2-editor-draft.md
@@ -3510,7 +3510,7 @@ The relevant path for this test is:
 
 ### 6.1.21 Missing Item in Revision History
 
-It must be tested that items of the revision history do not omit a version number when the items are sorted ascending by `date`. In the case of semantic versioning, this applies only to the Major version.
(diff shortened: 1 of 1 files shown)


csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -3510,7 +3510,7 @@ The relevant path for this test is:
 
 ### 6.1.21 Missing Item in Revision History
 
-It must be tested that items of the revision history do not omit a version number when the items are sorted ascending by `date`. In the case of semantic versioning, this applies only to the Major version.
+It must be tested that items of the revision history do not omit a version number when the items are sorted ascending by `date`. In the case of semantic versioning, this applies only to the Major version. It must also be tested that the first item in such a sorted list has either the version number 0 or 1 in the case of integer versioning or a Major version of 0 or 1 in the case of semantic versioning.
```
sthagen, line 3513: I am ˋ-0ˋ on the addition within the document at this stage. Rationale: I encounter revision histories as secondary literature more often than not. My reluctance comes from my hope that the TC unanimously approves this document as CS01 with non-material changes.
Worst case of not implementing the test now: some orderly forces take over authoring of a security advice and there are gaps in the change history - that should not require post-history changes but instead be maintained as historic fact and move on.

Issue #396, opened 2022-01-31
Use "openpgp" to indicate that the open standard is meant


bernhardreiter:
Section 7.1.20 mentions `PGP` which is a brand by Broadcom these days, the open standard meant is `OpenPGP` (e.g. https://datatracker.ietf.org/doc/html/rfc4880)

So at
https://github.com/oasis-tcs/csaf/blob/master/csaf_2.0/prose/csaf-v2-editor-draft.md?plain=1#L5144-L5144
PGP should be replaced with OpenPGP.

bernhardreiter:
Mentioned on csaf-comment today: https://groups.oasis-open.org/discussion/other-suggestions-partly-already-handled-395-396-397-398-469-470-477-508

PR #403, opened 2022-02-04
Replace PGP with OpenPGP

tschmidtb51:
- fix https://github.com/oasis-tcs/csaf/issues/396
- replace PGP to OpenPGP
- refer to RFC4880


sthagen (approved):
LGTM.

PR #404, opened 2022-02-04
Public openpgp key

### LINKED ISSUE #397: field suggestions "pgp_keys" -> "public_openpgp_keys"
Like #52 "pgp" should be renamed to "openpgp" in the `provider-metadata.json` field definitions (which is done by example in csaf 2.0).

The renaming could be done with the next revision of the standard as it a substaintical change.
When done, it should also be taken into considerdation to add a "public_" to the field name for extra clarity.
(Extra clarity helps security and there were cases where a private key was send in error by a human, where only a public key was requested by an outside party. If these 7 bytes help to lower the chances of this happening, they maybe well spend. ;) )

tschmidtb51:
- fix #397 
- rename `pgp_keys` with `public_openpgp_keys`
- adopt titles and descriptions
- adopt example
- suggest a location for OpenPGP keys


sthagen (approved):
LGTM.

Issue #389, opened 2022-01-26
Add test for private languages (warning)


tschmidtb51:
Private languages should not be used in CSAF documents as they have no meaning for customers (which

sthagen:
@tschmidtb51 which … lacks a continuation, I guess.

tschmidtb51:
> @tschmidtb51 which … lacks a continuation, I guess.

Just edited that one as well. You spot all my flaws tonight... :wink:

sthagen:
Only those that act, can produce errors … ❤️

PR #412, opened 2022-02-07
Private languages

tschmidtb51:
- resolves https://github.com/oasis-tcs/csaf/issues/389
- add test for private and default languages
- add test files


sthagen (approved):
LGTM.

Issue #400, opened 2022-02-01
Add optional test


tschmidtb51:
An advisory should provide at least one `product_identification_helper` for each product.

PR #405, opened 2022-02-04
Support for multiple ids in vulnerabilities

tschmidtb51:
- fix #388 
- change `id` into `ids` of type `array`
- adopt specification
- adopt example
- add conversion rule
- adopt Appendix C
- enforce uniqueness

PR #414, opened 2022-02-08
Product identification helper test

tschmidtb51:
- resolves https://github.com/oasis-tcs/csaf/issues/400
- add test for product_identification_helper
- correct numbering
- add test file

### DIFF
--- a/csaf_2.0/prose/csaf-v2-editor-draft.md
+++ b/csaf_2.0/prose/csaf-v2-editor-draft.md
@@ -4577,7 +4577,7 @@ The relevant paths for this test are:
 
 > A tool MAY remove such subtag as a quick fix.
 
-### 6.2.14 Use of Default Language
+### 6.2.15 Use of Default Language
 
 For each element of type `/$defs/language_t` it must be tested that the language code is not `i-default`.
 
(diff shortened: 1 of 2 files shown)


csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -4598,6 +4598,31 @@ The relevant paths for this test are:
 
 > A tool MAY remove such element as a quick fix.
 
+## 6.2.16 Missing Product Identification Helper
+
+For each element of type `/$defs/full_product_name_t` it must be tested that it includes the property `product_identification_helper` and at least one Product Identification Helper is provided.
```
tschmidtb51, line 4603: I guess that is the same question as in https://github.com/oasis-tcs/csaf/pull/405#discussion_r801832289 ...

Can we rely on the schema to enforce non-empty objects?

My first intention even was to write:
```suggestion
For each element of type `/$defs/full_product_name_t` it must be tested that at least one Product Identification Helper is provided.
```

tschmidtb51: Another option would be:
```suggestion
For each element of type `/$defs/full_product_name_t` it must be tested that it includes the property `product_identification_helper`.

tschmidtb51: Personally, I tend to option 2...

sthagen: I think changing that strategy of ours now is a bit late down the road. I remember when we discussed the optional elements (most of the CSAF vocabulary). I was in favor for relying on the „do not include null values“ convention as convention, while the TC wanted a more strict approach to ease the receiver side in automation. so, we should trust the schema for cardinality IMO, because it represents important constraints well within the scope of a schema language.

sthagen (approved):

Issue #423, opened 2022-02-10
Clarify handling of CVSS conversion


tschmidtb51:
There are a few things unclear when converting CVSS from CVRF to CSAF. We should clarify:
1. What should happen, when the `product_status` element is not given? => suggestion: error in conversion
2. What should happen, if the CVSS vector was not given? => suggestion: error in conversion
3. How to reliable determine CVSS 3.x versions?

tschmidtb51:
@sthagen: Does that sound reasonable?

sthagen:
> @sthagen: Does that sound reasonable?

I guess implementers will be able to use these steps for the processing. So, yes.

PR #424, opened 2022-02-11
CVSS conversion

tschmidtb51:
- resolves oasis-tcs/csaf#423
- add conversion rules for CVSS

### DIFF
--- a/csaf_2.0/prose/csaf-v2-editor-draft.md
+++ b/csaf_2.0/prose/csaf-v2-editor-draft.md
@@ -5683,8 +5683,28 @@ Secondly, the program fulfills the following for all items of:
 * `/product_tree/relationships[]`: If more than one `prod:FullProductName` instance is given, the CVRF CSAF converter converts the first one into the `full_product_name`. In addition, the converter outputs a warning that information might be lost during conversion of product relationships.
 * `/vulnerabilities[]/cwe`: If more than one `vuln:CWE` instance is given, the CVRF CSAF converter converts the first one into `cwe`. In addition, the converter outputs a warning that information might be lost during conversion of the CWE.
 * `/vulnerabilities[]/ids`: If a `vuln:ID` element is given, the CVRF CSAF converter converts it into the first item of the `ids` array.
-* `/vulnerabilities[]/scores[]`: If no `product_id` is given, the CVRF CSAF converter appends all Product IDs which are listed under `../product_status` in the arrays `known_affected`, `first_affected` and `last_affected`.
-* `/vulnerabilities[]/scores[]`: If there are CVSS v3.0 and CVSS v3.1 Vectors available for the same product, the CVRF CSAF converter discards the CVSS v3.0 information and provide in CSAF only the CVSS v3.1 information.
+* `/vulnerabilities[]/scores[]`:
+  * If no `product_id` is given, the CVRF CSAF converter appends all Product IDs which are listed under `../product_status` in the arrays `known_affected`, `first_affected` and `last_affected`. If none of these arrays exist, the CVRF CSAF converter outputs an error that no matching Product ID was found for this score element.
+  * If a `vectorString` is missing, the CVRF CSAF converter outputs an error that the CVSS element could not be converted as the CVSS vector was missing. A CVRF CSAF converter MAY implement a config option to delete such elements.
+  * If there are CVSS v3.0 and CVSS v3.1 Vectors available for the same product, the CVRF CSAF converter discards the CVSS v3.0 information and provide in CSAF only the CVSS v3.1 information.
+  * To determine, which minor version of CVSS v3 is used, the CVRF CSAF converter uses the following steps:
+    1. Retrieve the CVSS version from the CVSS vector, if present.
+
+        *Example XYZ:*
+
+        ```
+          CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H => 3.1
+        ```
+
+    2. Retrieve the CVSS version from the CVSS namespace, if present. The CVRF CSAF converter outputs a warning that this value was guessed from the namespace. If more than one CVSS namespace is present and the element is not clearly defined via the namespace, this step MUST be skipped without a decision.
+
+        *Example XYZ:*
+
+        ```
+          xmlns:cvssv3="https://www.first.org/cvss/cvss-v3.0.xsd" => 3.0
+        ```
+
+    3. Retrieve the CVSS version from a config value, which defaults to `3.0`. (As the spec only mentions 3.0.) The CVRF CSAF converter outputs a warning that this value was taken from the config.
 
 ### 9.1.6 Conformance Clause 6: CSAF content management system
 


tschmidtb51:
My first impression was that this could be valid:
```
<CVSSScoreSets>
      <ScoreSetV3 xmlns="https://www.first.org/cvss/cvss-v3.0.xsd>
        <BaseScoreV3>7.5</BaseScoreV3>
         </ScoreSetV3>
      <ScoreSetV3 xmlns="https://www.first.org/cvss/cvss-v3.1.xsd>
        <BaseScoreV3>7.5</BaseScoreV3>
         </ScoreSetV3>
      </CVSSScoreSets>
```

In that case, the element would clearly indicate its version...

@sthagen: Is that valid?

@sthagen

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -5683,8 +5683,28 @@ Secondly, the program fulfills the following for all items of:
...
 * `/vulnerabilities[]/ids`: If a `vuln:ID` element is given, the CVRF CSAF converter converts it into the first item of the `ids` array.
-* `/vulnerabilities[]/scores[]`: If no `product_id` is given, the CVRF CSAF converter appends all Product IDs which are listed under `../product_status` in the arrays `known_affected`, `first_affected` and `last_affected`.
-* `/vulnerabilities[]/scores[]`: If there are CVSS v3.0 and CVSS v3.1 Vectors available for the same product, the CVRF CSAF converter discards the CVSS v3.0 information and provide in CSAF only the CVSS v3.1 information.
+* `/vulnerabilities[]/scores[]`:
+  * If no `product_id` is given, the CVRF CSAF converter appends all Product IDs which are listed under `../product_status` in the arrays `known_affected`, `first_affected` and `last_affected`. If none of these arrays exist, the CVRF CSAF converter outputs an error that no matching Product ID was found for this score element.
+  * If a `vectorString` is missing, the CVRF CSAF converter outputs an error that the CVSS element could not be converted as the CVSS vector was missing. A CVRF CSAF converter MAY implement a config option to delete such elements.
```
sthagen, line 5688: I wonder, if `implement a config option` &rightarrow;  `offer a configuration` would be more to the point here. My understanding is that we state two things in wun sentence:
1. the result of the transform MAY be deletion
2. The requester SHOULD have a say in the matter

The second requirement I would prefer to satisfy more implicit, hence the suggested rewording.

sthagen (approved):
I think this is a good and helpful change that provides implementers of converters an algorithm to optimally automate the CVSS conversion.

sthagen:
> My first impression was that this could be valid:
> 
> ```
> <CVSSScoreSets>
>       <ScoreSetV3 xmlns="https://www.first.org/cvss/cvss-v3.0.xsd>
>         <BaseScoreV3>7.5</BaseScoreV3>
>          </ScoreSetV3>
>       <ScoreSetV3 xmlns="https://www.first.org/cvss/cvss-v3.1.xsd>
>         <BaseScoreV3>7.5</BaseScoreV3>
>          </ScoreSetV3>
>       </CVSSScoreSets>
> ```
> 
> In that case, the element would clearly indicate its version...
> 
> @sthagen: Is that valid?

Revisiting what we wrote in CSAF CVRF 1.2 section 6.12 I conclude that yes. One or more ScoreSetV3 elements are allowed, and the namespace is the only explicitly conforming option to indicate the version splits that only evolved after the schema had been published.

tschmidtb51:
> > My first impression was that this could be valid:
> > ```
> > <CVSSScoreSets>
> >       <ScoreSetV3 xmlns="https://www.first.org/cvss/cvss-v3.0.xsd>
> >         <BaseScoreV3>7.5</BaseScoreV3>
> >          </ScoreSetV3>
> >       <ScoreSetV3 xmlns="https://www.first.org/cvss/cvss-v3.1.xsd>
> >         <BaseScoreV3>7.5</BaseScoreV3>
> >          </ScoreSetV3>
> >       </CVSSScoreSets>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In that case, the element would clearly indicate its version...
> > @sthagen: Is that valid?
> 
> Revisiting what we wrote in CSAF CVRF 1.2 section 6.12 I conclude that yes. One or more `ScoreSetV3` elements are allowed, and the namespace is the only explicitly conforming option to indicate the version splits that only evolved after the `vuln` schema had been published.

Looks like this example fails already in the input validation. 

> Input document not valid: Element '{https://www.first.org/cvss/cvss-v3.0.xsd}ScoreSetV3': This element is not expected. Expected is one of ( {http://docs.oasis-open.org/csaf/ns/csaf-cvrf/v1.2/vuln}ScoreSetV2, {http://docs.oasis-open.org/csaf/ns/csaf-cvrf/v1.2/vuln}ScoreSetV3 ). (<string>, line 0).

We have 2 options here:
1. State that the CVRF CSAF converter must ignore that input error and proceed.
2. Or remove the phrase `and the element is not clearly defined via the namespace`.

@sthagen: Thoughts?

sthagen:
@tschmidtb51  how unfortunate 🤷🏽‍♂️ I would go with option 1. Producers always were creative working around constraints of CRF 1.1 already (tunneling CVSSv3.0 scores per the v2 only CVSS element, so there may be valid CVRf documents out there having stated CVSSv3.1 vectors in CVRf v1.2 regime subversion.

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -5696,15 +5696,34 @@ Secondly, the program fulfills the following for all items of:
...
+    3. Retrieve the CVSS version from the CVSS namespace given in the root element, if present. The CVRF CSAF converter outputs a warning that this value was guessed from the global namespace. If more than one CVSS namespace is present and the element is not clearly defined via the namespace, this step MUST be skipped without a decision.
 
         *Example XYZ:*
 
         ```
           xmlns:cvssv3="https://www.first.org/cvss/cvss-v3.0.xsd" => 3.0
         ```
 
-    3. Retrieve the CVSS version from a config value, which defaults to `3.0`. (As the spec only mentions 3.0.) The CVRF CSAF converter outputs a warning that this value was taken from the config.
+    4. Retrieve the CVSS version from a config value, which defaults to `3.0`. (As the spec only mentions 3.0.) The CVRF CSAF converter outputs a warning that this value was taken from the config.
```
sthagen, line 5726: Maybe replace `As the spec only mentions 3.0. ` with `As CSAF CVRF v1.2 predates CVSS v3.1` Or similar. The itch I try to scratch is the ambiguity of the term `the spec` Inside another specification (Russian doll)

sthagen (approved):

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: Done.

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: Done in [44eab57](https://github.com/oasis-tcs/csaf/pull/424/commits/44eab5729e15727cd089a785f8c01427920ba02e)

Issue #436, opened 2022-02-17
Add test: use only one versioning


tschmidtb51:
We should add a test to check that the version scheme isn't changed during the revision history.

PR #437, opened 2022-02-17
One version system

tschmidtb51:
- resolves oasis-tcs/csaf#436
- add test for only one versioning system
- add test file

### DIFF
--- a/csaf_2.0/prose/csaf-v2-editor-draft.md
+++ b/csaf_2.0/prose/csaf-v2-editor-draft.md
@@ -4214,7 +4214,7 @@ The relevant path for this test is:
 
 #### 6.1.30 Integer and Semantic Versioning
 
-It must be tested that all elements of type `/$defs/version_t` follow the same versioning system. Integer versioning and semantic versioning SHALL NOT occur within the same document.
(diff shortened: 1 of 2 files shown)


csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -4212,6 +4212,43 @@ The relevant path for this test is:
 
 > A tool MAY add all products of the affected group of this vulnerability to the remediation as quick fix.
 
+#### 6.1.30 Integer and Semantic Versioning
+
+It must be tested that all elements of type `/$defs/version_t` follow the same versioning system. Integer versioning and semantic versioning SHALL NOT occur within the same document.
```
tschmidtb51, line 4217: @sthagen Should we use `SHALL NOT` or `MUST NOT`in the second sentence? The original requirement from section [3.1.11](https://docs.oasis-open.org/csaf/csaf/v2.0/cs01/csaf-v2.0-cs01.html#3111-version-type) states:

> A CSAF document MUST use only one versioning system.

sthagen: Here we may profit from using a single sentence like so:

It must be tested that all elements of type `/$defs/version_t` follow either integer versioning or semantic versioning homogeneously within the same document.

Issue #440, opened 2022-02-18
Strip leading and trailing whitespaces before generating filename


tschmidtb51:
Currently, leading and trailing whitespaces are converted to `_` which seems to be awkward. The `/document/tracking/id` SHOULD NOT contain any whitespaces.

PR #448, opened 2022-02-18
Prohibit trailing white spaces 

tschmidtb51:
- resolves https://github.com/oasis-tcs/csaf/issues/440
- add restriction that Document Tracking ID can't start or end with white space nor have a line break inside
- add informative comment about regex
- correct order in JSON schema for pattern


sthagen (approved):
The shouting words in the informative text part seem to be OK here, because these only illustrate the normative regex in the normative schema artifact.

PR #449, opened 2022-02-18
Spelling & Formatting & Other

tschmidtb51:
- fix spelling mistakes
- correct format issues
- add OpenPGP key strength comment
- adopt review comments

PR #450, opened 2022-02-18
Editor revision 2022 02 23

### LINKED ISSUE #368: Integrity Files / Signatures to be mentioned in Rolie Feed?
Hello together,

is there any requirement (as CSAF "trusted provider" role) to mention the integrity files / signature files (.asc) in the rolie feed? I have not found a dedicated field for that.

Or is it enough to place them just with the correct naming next to the document file (as described in the specification draft)? It is not clear for me.

Thank you
Klaus

### LINKED ISSUE #371: Relation of categories in the category document and feed / service document
As far as I understood the [documentation ](https://datatracker.ietf.org/doc/html/draft-ietf-mile-rolie)

the category field within the feed and the service document have to be identical. If not correct please let me know.

```json
              "category": [
                {
                  "scheme": "urn:ietf:params:rolie:category:information-type",
                  "term": "csaf"
                }
              ]
```

But how does the listed term relate to the category-term listed in the category document? Has the term to be identical?  ​

Related to this, the question comes up, if - in the case of having multiple category entries in the category-document (as shown in the [example category.json](https://docs.oasis-open.org/csaf/csaf/v2.0/csd01/csaf-v2.0-csd01.html#7117-requirement-17-rolie-category-document)) - all of them have to be included in the feed / service document? Or can the feed / service document include just a subset?

If the categories mentioned in the category document do not relate to the feed / service document, how can they be chosen?

Thanks a lot!

### LINKED ISSUE #377: Section 2 large block of informative text carries normative information in disguise
The current section 2 states:

# 2 Design Considerations
The Common Security Advisory Framework (CSAF) is a language to exchange Security Advisories formulated in JSON.

Informative Comment:

>The term Security Advisory as used in this document describes any notification of security issues in products of and by providers. Anyone providing a product is considered in this document as a vendor, i.e. developers or maintainers of information system products or services. This includes all authoritative product vendors, Product Security Incident Response Teams (PSIRTs), and product resellers and distributors, including authoritative vendor partners. A security issue is not necessarily constraint to a problem statement, the focus of the term is on the security aspect impacting (or not impacting) specific product-platform-version combinations. Information on presence or absence of work-arounds is also considered part of the security issue. This document is the definitive reference for the language elements of CSAF version 2.0. The encompassing JSON schema file noted in the Additional Artifacts section of the title page shall be taken as normative in the case a gap or an inconsistency in this explanatory document becomes evident. The following presentation in this section is grouped by topical area, and is not simply derivative documentation from the schema document itself. The information contained aims to be more descriptive and complete. Where applicable, common conventions are stated and known common issues in usage are pointed out informatively to support implementers of document producers and consumers alike. The section 3 Schema elements derives from the JSON schema itself as a service for the reader

# Proposal

Revisit the wording of this „massive“ portion of section 2 and esp. try to move text into normative parts.

Example: Inside an informative only text the sentence (cf. above) seems misplaced:

> The encompassing JSON schema file noted in the Additional Artifacts section of the title page shall be taken as normative in the case a gap or an inconsistency in this explanatory document becomes evident.

### LINKED ISSUE #379: Para 8 in section 2.1 - leftover?
In the current [section 2.1](https://docs.oasis-open.org/csaf/csaf/v2.0/cs01/csaf-v2.0-cs01.html#21-construction-principles) eighth para (counting the list as para) seems to be leftover from the 1.2 XML version and placeholder like term:

> A dictionary like presentation of all defined schema elements is given in the section SCHEMASECTIONNUMBER Schema. Any expected relations to other elements (linkage) is described there. This linking relies on setting attribute values accordingly (mostly guided by industry best practice and conventions) and thus implies, that any deep validation on a semantic level is to be ensured by the producer and consumer of CSAF documents. It is out of scope for this specification.

### LINKED ISSUE #382: Clarify source language
The standard should clarify that the source language should not be set if the document is not translated. A test for that is suggested.

### LINKED ISSUE #383: Bind remediation to product
A remediation should enforce that it is bound to at least one `product_id` or `group_id`. Currently, a remediation without that can exist.

### LINKED ISSUE #384: Add model as new Product identification helper
Many products (esp. hardware) have model "numbers" that can be used to identify them. We should add that as a `product_identification_helper`.

Here is the implementation suggestion:
```
            "models": {
              "title": "List of models",
              "description": "Contains a list of parts, or full models.",
              "type": "array",
              "minItems": 1,
              "items": {
                "title": "Model number",
                "description": "Contains a part, or a full model number of the component to identify.",
                "type": "string",
                "minLength": 1
              }
            },
```

### LINKED ISSUE #385: Clarify conversion of more than one CWE
> In the [specification of CVRF 1.2](http://docs.oasis-open.org/csaf/csaf-cvrf/v1.2/cs01/csaf-cvrf-v1.2-cs01.html#_Toc493508923) is stated that CWE _element MUST be present zero or one time_, while in [xsd scheme of CVRF 1.2](http://docs.oasis-open.org/csaf/csaf-cvrf/v1.2/cs01/schemas/vuln.xsd) the occurrence is not bounded `<xs:element name="CWE" minOccurs="0" maxOccurs="unbounded">`.

For the CVRF CSAF converter, we should clarify:
- If more than one vuln:CWE instance is given, the CVRF CSAF converter converts the first one into the cwe. In addition, the converter outputs a warning that information might be lost during conversion of the cwe.

### LINKED ISSUE #388: Support for multiple id properties in vulnerability
Currently, a vulnerability may have at most one vulnerability id.  This is a bit limiting, especially if the id used is something along the lines of an internal bug id.  I'm proposing an adjustment to the id property such that it is an array of id properties.  For example:

```
"vulnerabilities": [
  {
    "id": [
               {
                 "system_name": "Bug ID",
                 "text": "1234"
               },
               {
                 "system_name": "Bug ID",
                 "text": "5678"
               },
    ],
   ...
  }
]
```

### LINKED ISSUE #389: Add test for private languages (warning)

### LINKED ISSUE #390: Add VEX flags
The TC should consider to support VEX flags.

### LINKED ISSUE #392: Clarify usage of version ranges
I strongly believe that the easiest way for a customer is, not having to worry about version ranges. Nevertheless, I also understand the PSIRTs working hard to compile advisories to the best of their knowledge. And sometimes PSIRTs don't know exactly which versions are affected.

Therefore, I suggest the following approach:

1. Introduce a new category `product_version_range`. This one holds all version ranges while the existing `product_version` only holds single versions. Write a mandatory test to enforce it. (Could be something like "check whether `all versions`, `<`, `<=`, `prior`, `later` exist in the value.)
2. Clarify, how version ranges should be represented in CSAF. An approach could be to specify:
  - if possible: use [vers](https://github.com/package-url/purl-spec/blob/version-range-spec/VERSION-RANGE-SPEC.rst)
  - if vers is not possible: use a specifier that follows the rules of vers as closely as possible (then the programs can make a best effort guess whether it is in range)

Thoughts?

Flagging @santosomar, @tolim, @mprpic, @wrideout for attention.

### LINKED ISSUE #394: Add field for legacy number in revision history
CSAF enforces a specific set of versioning. However, current advisories might use a different version scheme. We should have a place to aid in mapping versions in the `revision_history`.

### LINKED ISSUE #396: Use "openpgp" to indicate that the open standard is meant

### LINKED ISSUE #397: field suggestions "pgp_keys" -> "public_openpgp_keys"

### LINKED ISSUE #398: Indicate order of places to look for the provider-metadata.json file (to ease implementation)
csaf 2.0 requires "CSAF providers" to check three places for the `provider-meta-data.json` file. 

https://github.com/oasis-tcs/csaf/blob/5ac8645b9da054634bc5e2810bcacce06dbf0c8e/csaf_2.0/prose/csaf-v2-editor-draft.md?plain=1#L5289

To ease  and streamline implementations, we suggest to give an explicit order by which implementations reading CSAF conforming publishing sites should check for the file. This order should also correspond to what is prefered where providers (and aggregators) shall publish the `provider-meta-data.json`. 

This also affects the writing in requirements 7 to 10 as 7 gives suggestions without order and thus has an avoidable overlap with the different locations in 8,9 and 10.

(The suggestion from implementing https://github.com/csaf-poc/csaf_distribution is to check the .well-known location first, then security and then DNS.)

### LINKED ISSUE #399: Clarify conversion of language
CVRF CSAF 1.2 had some information about the language which was hidden in the notes and applied only for specific elements. We can either:
1. define that if such elements exist, the `lang` should be set to the value that was found (if multiple throw an warning that the document might contain multiple languages),
2. define that it should be set by config
3. define to ignore it.

I'm current in favor of option 1.

Thoughts?

### LINKED ISSUE #400: Add optional test

### LINKED ISSUE #406: Add test for CVE in `ids` field (warning)
We should add a test, that fails if a CVE-ID is given in one of the items of `/vulnerabilities[]/ids`. This is based on #388.

### LINKED ISSUE #408: Add `csaf_version` into the profiles
`csaf_version` is currently not mandatory in the profiles, but according to the schema, it is. Therefore, we should add that in the profile `Generic CSAF`.

### LINKED ISSUE #413: Consider guidance to use `product_tree`
@tlimmer commented in https://github.com/oasis-tcs/csaf/commit/c2ff5a517c253ad1224319d84dfa992a4bc53e09#r49169485:

>Suggestion: add further explanation on the quite complex nature of `branches_t`, such as:
>
>> `branches_t` supports building a hierarchical structure of products that allows to indicate the relationship of products to each other and enables grouping for simpler referencing. As an example, the structure may use the following levels: `vendor` -> `product_family` -> `product_name` -> `product_version`.

We should consider to add this to the standard as an informational note.

### LINKED ISSUE #416: Clarify serial numbers' uniqueness in `product_identification_helper`
We should clarify that a serial number should be unique in the list of serial_numbers. There is no value in having a single serial number in there multiple times...

### LINKED ISSUE #423: Clarify handling of CVSS conversion

### LINKED ISSUE #425: Clarify conversion of `Realm` and `Resource`
The [XML schema](https://docs.oasis-open.org/csaf/csaf-cvrf/v1.2/cs01/schemas/prod.xsd) defines `Realm` and `Resource` as possible  `Type` for a `Branch` - the [specification](http://docs.oasis-open.org/csaf/csaf-cvrf/v1.2/cs01/csaf-cvrf-v1.2-cs01.html#_Toc493508797) does not. We need to clarify, what to do, if such values occur.

Suggestion: Convert them into a `product_name` and add a warning that those types do not exist in CSAF.

@sthagen: Do you agree?

### LINKED ISSUE #426: Enhance category semantics
# Work in Progress ...
Context: [CSAF v2.0 CS01 section 3.2.1.3 "Document Property - Category"](https://docs.oasis-open.org/csaf/csaf/v2.0/cs01/csaf-v2.0-cs01.html#3213-document-property---category) 

In CS01 we open pandoras box with category content rules on both sides.

1. We offer a text field but aim for enumeration (type mismatch)
2. There is no reserved space for future addition from the TC (profile names)

Adding insult to injury, 

3. we do not fully separate even the existing enumeration values from everything else (visual distinction).

>I created an issue in my browser already with detailed links and verbiage explaining the rationale on February, 6 but did not click the insole green button below the form before giving all bits away ... so, I choose to not explain again.

In short:

1. We seem to not be able to change, as we want one attribute to allow two things:
  a. Prescribe CSAF TC approved profiles (per community input, enabling in spec description and validation)
  b. Allow every community to make up their specific profiles "in not time"
2. I think we have to change and below is proposal (ii) that describes my vision on how to fix
3. Inspecting the existing implementation of [Secvisogram](https://github.com/secvisogram/secvisogram) indicates additional verbiage is needed to guide implementers as given per proposal (iii)

# Proposals

## Ad (1) - consider separation of concerns (i)

Let us consider to come up with a new name, say "profile" as an added field and sibling to the "category" field that allows us
to specify the content as "real" enumeration providing only the values within the specification at any version.
We could than leave the category to whatever parties like to use it for (can be private community profile enumerations can be extended lyrics like "Vendor Security Notification", you name it.

> I would love to see some discussion on this from the members working in the field (vendor and consumer side).

## Ad (2) - rename existing CSAF profiles and reserve the name space (ii)

I hereby suggest that we perform two types of change:

1. Introduce a namespace prefix for profiles defined within the spec (or any future version thereof and simply name that `csaf_`.
2. Rename the existing profiles to use this `csaf_` prefix and concat all profile specific words in lowercase per underscore (`_`).

Change 2 would map:

* Profile 1: "Generic CSAF": `generic_csaf` &rightarrow; `csaf_generic` (or maybe better `csaf_base`) - for discussion, please see below this list
* Profile 2: "Security incident response": `security_incident_response` &rightarrow; `csaf_security_incident_response`
* Profile 3: "Informational Advisory": `informational_advisory` &rightarrow; `csaf_informational_advisory`
* Profile 4: "Security Advisory": `security_advisory` &rightarrow; `csaf_security_advisory`
* Profile 5: "VEX": `vex` &rightarrow; `csaf_vex` (but as fashions come and go maybe better more explicitly like `csaf_vulnerability_exploitability_exchange`, esp. in case the acronym vexes us :wink:)

Discussion of Profile 1:

The current use of the "Generic CSAF" profile may send us all to the world of "Alice in Wonderland" one day or another.
We essentially specify (the describing section together with our mandatory rules) that this "profile" in reality states what we expect when no profile applies. That would all be easy to implement (and automatically produce and consume) if and only if we did not expect communities to use the exact field value to make up their own profiles.

I hereby suggest that for validation we make clear the following rules (we already have):

* Every document must comply to the rules of `csaf_base` (the old generic in new clothes) 
* Every other profile can only add to the `csaf_base` (i.e. to maintain the base validation claim)

are more clearly stated and the name that any category value not "matching" profiles 2 bis implies `csaf_base` or `csaf_core` which both better describe the expectations on a document matching.

"Matching" in the previous sentence implies "by the rules of the specification" (cf. iii)

## Ad (3) - clarify expectations on profile value separation (iii)

We should disallow padding and leading characters considered irrelevant for profile names by this spec. That we can simply do by language change in the prose of the category section:

Suggested is to write instead of the current [6.1.26 "Prohibited Document Category Name"](https://docs.oasis-open.org/csaf/csaf/v2.0/cs01/csaf-v2.0-cs01.html#6126-prohibited-document-category-name) first paragraph:

> It must be tested that the document category is not equal to the (case insensitive) name of any other profile than "Generic CSAF". This does not differentiate between underscore, dash or whitespace. This test does only apply for CSAF documents with the profile "Generic CSAF". Therefore, it must be skipped if the document category matches one of the values defined for the profile other than "Generic CSAF".

the following (or similar):

> It must be tested that the document category is not equal to the (case insensitive) name of any other profile than "Generic CSAF". This does not differentiate between underscore, dash or space and includes the. start and the end of the field value. This test does only apply for CSAF documents with the profile "Generic CSAF". Therefore, it must be skipped if the document category matches one of the values defined for the profile other than "Generic CSAF".

### LINKED ISSUE #431: Rules about usage of `*_on_CSAF_aggregators` should be formal
Currently, we state these rules as informal:

> Issuing parties MUST indicate through the value `false` in `list_on_CSAF_aggregators` if they do not want to be listed. Issuing parties MUST indicate through the value `false` in `mirror_on_CSAF_aggregators` if they do not want to be mirrored.

However, that should be formal language to be binding for issuing parties. Otherwise, nobody could rely on it...

### LINKED ISSUE #434: Add current practice on invalid files
It is a current practice to append `_invalid` to the filename to indicate that a CSAF document is invalid. We should add an informational comment on that.

### LINKED ISSUE #436: Add test: use only one versioning

### LINKED ISSUE #440: Strip leading and trailing whitespaces before generating filename

### LINKED ISSUE #441: Clarify filename computation
Currently, it seems to be unclear whether consecutive characters which should be replaced by an `_` get one or multiple `_`.

Example: `Hans Peter_Hase 010-10==?/1,+-`

1. Multiple: `hans_peter_hase_010-10____1_+-`
2. Single: `hans_peter_hase_010-10_1_+-`

Personally, I prefer option 2.

@sthagen: Thoughts?

### LINKED ISSUE #453: Add vulnerability to terminology
In CVRF CSAF 1.2, we defined vulnerability according to ISO 29147. Currently, we do not have any definition in the document. Therefore, I suggest to either use that from ISO 29174 (which is currently not available to me; @zmanion (or anyone) are you able to provide that?) 

or use the one from the [The CERT Guide to Coordinated Vulnerability Disclosure](https://vuls.cert.org/confluence/display/CVD/1.2.+CVD+Context+and+Terminology+Notes):

> A vulnerability is a set of conditions or behaviors that allows the violation of an explicit or implicit security policy.

Thoughts?

CC: @tolim, @mprpic, @santosomar, @wrideout

tschmidtb51:
Dear colleagues,
we have worked on the new editor revision. It addresses various issues that have been discovered by TC members and programmers working on the implementation of the standard. This is a PR to merge it into the master branch.

Please review this PR before our next TC meeting on 2022-02-23. If you have comments, please mark them in your review or open a new issue. You can review the changes via the [Files changed tab](https://github.com/oasis-tcs/csaf/pull/450/files) or read the [whole editor revision as rendered markdown](https://github.com/oasis-tcs/csaf/blob/editor-revision-2022-02-23/csaf_2.0/prose/csaf-v2-editor-draft.md).

Your feedback is very much welcome.

Thank you!

---

Please find below a list of addressed issues:
- fixes #453
- fixes #441 
- fixes #440
- fixes #436 
- fixes #434 
- fixes #431 
- fixes #425 
- fixes #423 
- fixes #416
- fixes #413
- fixes #408 
- fixes #406 
- fixes #400
- fixes #399 
- fixes #398 
- fixes #397 
- fixes #396 
- fixes #394 
- fixes #389 
- fixes #388 
- fixes #385 
- fixes #384 
- fixes #383 
- fixes #382 
- fixes #379 
- fixes #377 
- fixes #371 
- fixes #368
- spelling, formatting and other (#449)

---

The following issues were discussed and approved during the monthly TC meeting on 2022-02-23:

- fixes #392
- fixes #426
- fixes #390 

---

**Note**: to be consistent we use lower case for MUST/SHOULD/SHALL/MAY in informative statements. Please flag, if we missed one.

### DIFF
--- a/csaf_2.0/prose/csaf-v2-editor-draft.md
+++ b/csaf_2.0/prose/csaf-v2-editor-draft.md
@@ -331,7 +331,9 @@ https://www.iso.org/standard/72311.html.
 ###### [OPENSSL]
 _GTLS/SSL and crypto library_, OpenSSL Software Foundation, https://www.openssl.org/.
 ###### [PURL]
-_Package URL (PURL)_, GitHub Project, https://github.com/package-url/purl-spec
+_Package URL (PURL)_, GitHub Project, https://github.com/package-url/purl-spec.
+###### [RFC3339]
+Klyne, G. and C. Newman, "Date and Time on the Internet: Timestamps", RFC 3339, DOI 10.17487/RFC3339, July 2002, https://www.rfc-editor.org/info/rfc3339.
 ###### [RFC3552]
 Rescorla, E. and B. Korver, "Guidelines for Writing RFC Text on Security Considerations", BCP 72, RFC 3552, DOI 10.17487/RFC3552, July 2003, https://www.rfc-editor.org/info/rfc3552.
 ###### [RFC3986]


tschmidtb51:
Flagging @mprpic, @tolim, @wrideout to review it.

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -1888,7 +1888,16 @@ Engine version (`version`) of value type `string` with 1 or more characters cont
...
-Unique identifier for the document (`id`) of value type `string` with 1 or more characters holds the Identifier.
+Unique identifier for the document (`id`) of value type `string` with 1 or more characters with `pattern` (regular expression):
+
+```
+    ^[\\S](.*[\\S])?$
+```
+
+Unique identifier for the document holds the Identifier.
+
+> It SHALL NOT start or end with a white space and SHALL NOT contain a line break.
```
sthagen, line 1899: Let us remove any space coloring :wink: ... end with space characters ...

Why anybody would want to use identifiers with spaces anywhere is a miracle to me.

sthagen (approved):

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: One of our examples does: `Example Company - 2019-YH3234`

I'm happy to change it if the TC wants to prohibit space characters in the `/document/tracking/id`.

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -5,9 +5,9 @@
 
 # Common Security Advisory Framework Version 2.0
 
-## Committee Specification 01
+## Committee Specification Draft 02
```
sthagen, line 8: I intended to create a PR from some ephemeral branch to the editor branch but my fingers chose differently.

Personally, I do not think that this is no problem as the change set is very small and clear in scope. So, I hope the TC members see it the same way ... 🙈

sthagen (approved):

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -298,8 +299,7 @@ _Common Vulnerability and Exposures (CVE) – The Standard for Information Secur
 ###### [CVE-NF]
 _Common Vulnerability and Exposures (CVE) – The Standard for Information Security Vulnerability Names - CVE ID Syntax Change_, MITRE, January 01, 2014, https://cve.mitre.org/cve/identifiers/syntaxchange.html.
 ###### [CVRF-1-1]
-_The Common Vulnerability Reporting Framework (CVRF) Version 1.1_, M. Schiffman, Editor, May 2012, Internet Consortium for Advancement of Security on the Internet (ICASI), 
-http://www.icasi.org/the-common-vulnerability-reporting-framework-cvrf-v1-1/.
+_The Common Vulnerability Reporting Framework (CVRF) Version 1.1_, M. Schiffman, Editor, May 2012, Internet Consortium for Advancement of Security on the Internet (ICASI), http://www.icasi.org/the-common-vulnerability-reporting-framework-cvrf-v1-1/.
```
mprpic:
```suggestion
_The Common Vulnerability Reporting Framework (CVRF) Version 1.1_, M. Schiffman, Editor, May 2012, Internet Consortium for Advancement of Security on the Internet (ICASI), https://www.icasi.org/the-common-vulnerability-reporting-framework-cvrf-v1-1/.
```

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -392,12 +392,10 @@ Informative Comment:
...
+The term Security Advisory as used in this document describes any notification of security issues in products of and by providers. Anyone providing a product is considered in this document as a vendor, i.e. developers or maintainers of information system products or services. This includes all authoritative product vendors, Product Security Incident Response Teams (PSIRTs), and product resellers and distributors, including authoritative vendor partners.
+A security issue is not necessarily constraint to a problem statement, the focus of the term is on the security aspect impacting (or not impacting) specific product-platform-version combinations. Information on presence or absence of workarounds is also considered part of the security issue.
```
mprpic:
```suggestion
A security issue is not necessarily constrained to a problem statement, the focus of the term is on the security aspect impacting (or not impacting) specific product-platform-version combinations. Information on presence or absence of workarounds is also considered to be a part of the security issue.
```

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -440,10 +438,12 @@ Delegation to industry best practices technologies is used in referencing schema
 * Vulnerability Classification
   * Common Weakness Enumeration (CWE) [CWE]
     * CWE List: http://cwe.mitre.org/data/index.html
-* Classfication for Document Distribution
+* Classification for Document Distribution
   * Traffic Light Protocol (TLP)
     * Default Definition: https://www.first.org/tlp/
 
+Section 4 defined profiles which are used to ensure a common understanding which fields are required in a use case. Additional conventions are stated in section 5. The tests given in section 6 support CSAF producers and consumers to verify rules from the specification which can not be tested by the schema. Section 7 states how to distribute and where to find CSAF documents. Safety, Security and Data Protection are considered in section 8. Finally, a set of conformance targets describes tools in the ecosystem.
```
mprpic:
```suggestion
Section 4 defined profiles that are used to ensure a common understanding of which fields are required in a given use case. Additional conventions are stated in section 5. The tests given in section 6 support CSAF producers and consumers to verify rules from the specification which can not be tested by the schema. Section 7 states how to distribute and where to find CSAF documents. Safety, Security, and Data Protection are considered in section 8. Finally, a set of conformance targets describes tools in the ecosystem.
```

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -774,7 +777,7 @@ Product ID (`product_id`) holds a value of type Product ID (`product_id_t`).
 #### 3.1.3.3 Full Product Name Type - Product Identification Helper
 
 Helper to identify the product (`product_identification_helper`) of value type `object` provides in its properties at least one method which aids in identifying the product in an asset database.
-Of the given seven properties `cpe`, `hashes`, `purl`, `sbom_urls`, `serial_numbers`, `skus` and `x_generic_uris`, one is mandatory.
+Of the given eight properties `cpe`, `hashes`, `model_numbers`, `purl`, `sbom_urls`, `serial_numbers`, `skus` and `x_generic_uris`, one is mandatory.
```
mprpic:
```suggestion
Of the given eight properties `cpe`, `hashes`, `model_numbers`, `purl`, `sbom_urls`, `serial_numbers`, `skus`, and `x_generic_uris`, one is mandatory.
```

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -4825,11 +5088,52 @@ The relevant paths for this test are:
 
 > There is a spelling mistake in `Secruity`.
 
+## 6.3.9 Branch Categories
+
+For each element of type `/$defs/full_product_name_t` in `/product_tree/branches` it MUST be tested that ancestor nodes along the path exist which use the following branch categories `vendor` -> `product_name` -> `product_version` in that order starting with the Product tree node.
+
+> Other branch categories can be upfront, in between or after the aforementioned without making the test invalid.
```
mprpic:
```suggestion
> Other branch categories can be used before, after, or between the aforementioned branch categories without making the test invalid.
```

mprpic (commented):
Only minor nitpicks, otherwise looks good!

santosomar (approved):

sthagen (approved):

mprpic (approved):

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: Resolved through #458

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: Resolved through #458

sthagen (approved):

sthagen (approved):

santosomar:
Thank you to @sthagen and @tschmidtb51 for the significant amount of work and comprehensive coverage of all the editorial issues addressed here! As discussed on the TC meeting on 2022-02-23, we will give the TC time to review the changes until 2022-03-02 at midnight (UTC). If no further edits are suggested, I will merge this pull request shortly after.

sthagen (approved):

sthagen (approved):

sthagen (approved):

tschmidtb51:
Added with #466 a test do avoid contradiction when using `flags`.

sthagen (approved):

sthagen (approved):

sthagen (approved):

tschmidtb51:
We also added some last-minute clarifications:
- fixes #470 
- fixes #475

sthagen (approved):

sthagen (approved):

santosomar:
The review period ended. Merging the pull request as discussed during the TC monthly meeting.

Issue #429, opened 2022-02-16
Add conformance target CSAF SBOM matching system


tschmidtb51:
When I give a presentation about "How CSAF helps vendors / the supply chain" I also refer to a system that matches incoming CSAF documents against SBOMs. We should also have such system defined in the standard as counterpart to the CSAF asset matching system.

@sparrell: Are you aware of a system/tool that vendors use to manage (create, store, ...) their SBOMs?

rjb4standards:
@tschmidtb51 I recommend checking the Tools pages for SPDX and CycloneDX:
https://cyclonedx.org/tool-center/
https://spdx.dev/resources/tools/

tschmidtb51:
@rjb4standards Thanks for the hint.

sparrell:
responding to @tschmidtb51 question to me: "Are you aware of a system/tool that vendors use to manage (create, store, ...) their SBOMs?": multi-part answer. 

First, on the creation side: I think it is important to recognize SBOMs can be created at 3 very different points in lifecycle. We don't have a good name to distinguish these "types" of SBOMs but I'll call them types and say there are 3 types: 
- source-code, 
- build, 
- executable-analysis. 

They all have their uses and in ideal world you would have all 3, and potentially multiple of the same type even from different sources (eg different tools yield different executable analysis). Note this is totally independent of the "format issue" (eg spdx vs cdx vs swid). To answer your question for "create" - Yes there are tools for all 3 of these creation types. For source and build, they mostly come from the SDLC of the programming languages used and Dick already gave you the links to a long list of them. For the executable analysis, they mostly come from a plethora of vendors, but I think there are also some open source tools (not an area I look at much - I mainly pick up info on them at hacker cons).

Second dimension: previous paragraph is all in the "create" space. The "store" space is a related, with overlap, but totally independent space. Yes there both vendor (at least 5) and open source (at least 4) tools/projects I am aware of. Some vendors have been around for years and participate in NTIA/CISA work. E.g. Cybeats (who is a recent client of mine but I speak for sFractal, not Cybeats, at anything OASIS or SBOM), Adolus, Ion Channel. Others are newer entrants (at least from my having heard of them). 

[...]

PR #452, opened 2022-02-19
CSAF SBOM matching system

tschmidtb51:
- resolves oasis-tcs/csaf#429
- add conformance target CSAF SBOM matching system
- add definitions


sthagen (approved):

santosomar:

PR #454, opened 2022-02-19
Version ranges

### LINKED ISSUE #392: Clarify usage of version ranges

tschmidtb51:
- resolves #392 
- add new field `product_version_range`
- add prose
- add tests
- add test files

**Note**: This is based on the editor revision 2022-02-23 branch!

### DIFF
--- a/csaf_2.0/json_schema/csaf_json_schema.json
+++ b/csaf_2.0/json_schema/csaf_json_schema.json
@@ -99,6 +99,7 @@
               "product_family",
               "product_name",
               "product_version",
+              "product_version_range",
               "service_pack",
               "specification",
               "vendor"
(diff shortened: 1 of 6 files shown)


sthagen (approved):

santosomar:

santosomar (approved):

Issue #386, opened 2022-01-26
Add flags for product lifecycle


tschmidtb51:
Currently, we can't efficiently transport the information that a product is End of Life or End of Support. I suggest to remove the `lecagy` option from the `branches_t` and create a lifecycle flag in the `full_product_name`. 

The definition could be as follows:

```
"lifecycle_state": {
            "title": "State of the lifecycle",
            "description": "Describes in which state of the lifecycle the product is.",
            "type": "string",
            "enum": [
              "specification",
              "design",
              "implementation",
              "verification_validation",
              "feature_maintenance",
              "security_maintenance",
              "de-commission"
            ]
```

If anyone has better generic lifecycle states, happy to hear them.

sthagen:
Could we spare the dash. Would decommission also work? This way the enumeration would be valid identifiers in the programming languages I know …

tschmidtb51:
> Could we spare the dash? in other words: would decommission also work? This way, all enumeration values would be valid identifiers in the programming languages I know …

My bad - I updated the suggestion.

tschmidtb51:
Here are some examples:

_Example 1:_ (not longer supported, aka End of Support / End of Life)
```
      {
        "product_id": "CSAFPID-0001",
        "name": "Very Old Product A",
        "lifecycle_state": "decommission"
      }
```

_Example 2:_ (only security updates, but no feature updates anymore)
```
      {
        "product_id": "CSAFPID-0002",
        "name": "Old Product B",
        "lifecycle_state": "security_maintenance"
      }
```


_Example 3:_ (**Standard**, fully operational product)
```
      {
        "product_id": "CSAFPID-0003",
        "name": "Product C",
        "lifecycle_state": "feature_maintenance"
      }
```

_Example 4:_ (product is new, currently tested for production)
```
      {
        "product_id": "CSAFPID-0004",
        "name": "Product D beta",
        "lifecycle_state": "verification_validation"
      }
```

_Example 5:_ (product is currently being implemented)
```
      {
        "product_id": "CSAFPID-0005",
        "name": "Product E (test)",
        "lifecycle_state": "implementation"
      }
```

tschmidtb51:
Here is a todo list:
[ ] remove `legacy`
[ ] add new structure `lifecycle_state`
[ ] adopt prose
[ ] add conversion rules
[ ] adopt committee note draft
[ ] add test

sparrell:
[...]

PR #457, opened 2022-02-21
Product lifecycle

tschmidtb51:
- resolves https://github.com/oasis-tcs/csaf/issues/386
- remove branch category `legacy`
- add new lifecycle_state
- add description in prose
- adopt numbering
- add examples
- add conversion rule
- adopt committee note


sthagen (approved):

santosomar:
We discussed this in detail during the monthly TC meeting on 2022-02-23. It was decided that we will address this at a later time and include this in the considerations for CSAF 2.1. This is a very complicated issue to implement by vendors and consumers.

PR #460, opened 2022-02-22
VEX flags

tschmidtb51:
- resolves https://github.com/oasis-tcs/csaf/issues/390
- add flags for VEX and other purposes
- adopt prose to reflect schema changes
- add flags into vex profile
- add test to bind flags to product
- add test file (assumed to be 32)

**Note**: I chose the generic term `flags`. This will allow us in the future also to add machine readable flags for other purposes. At the moment, we only have the justifications from the VEX NOT_AFFECTED group.


csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -2278,6 +2283,69 @@ The Weakness name (`name`) has value type `string` with 1 or more characters and
...
+      "items": {
+        // ...
+      }
+    },
+```
+
+Every Flag item of value type `object` with the mandatory property Label (`label`) contains product specific information in regard to this vulnerability as a single machine readable flag.
+For example, this could be a machine readable justification code why a product is not affected.
+
+> These flags MAY be used to decide in an automated fashion which actions to take.
```
sthagen, line 2302: The shouting `may` normally alerts the `other` partner in sender/receiver, producer/consumer, or server/client protocols to deal with complications arising from an alternate reality. Here, we intend to suggest a use outside of the `parsing` arena.

Something like the following wording might be closer to the intended message:
> These flags enable the receiving party to automate the selection of actions to take.

tschmidtb51: Sorry - did that again. :see_no_evil:
Thank you for the suggestion. Fixed in [103ff4b](https://github.com/oasis-tcs/csaf/pull/460/commits/103ff4bba8f934d5b151f6e06f3b870078f688ba).

csaf_2.0/prose/csaf-v2-editor-draft.md
```diff
@@ -2278,6 +2283,69 @@ The Weakness name (`name`) has value type `string` with 1 or more characters and
...
+```
+
+The given values reflect the VEX not affected justifications. See [VEX-Just] for more details. The values MUST be used as follows:
+
+* `component_not_present`: The software is not affected because the vulnerable component is not in the product.
+* `vulnerable_code_not_present`: The product is not affected because the code underlying the vulnerability is not present in the product.
+  > Unlike `component_not_present`, the component in question is present, but for whatever reason (e.g. compiler options) the specific code causing the vulnerability is not present in the component.
+* `vulnerable_code_cannot_be_controlled_by_adversary`: The vulnerable component is present, and the component contains the vulnerable code. However, vulnerable code is used in such a way that an attacker cannot mount any anticipated attack.
+* `vulnerable_code_not_in_execute_path`: The affected code is not reachable through the execution of the code, including non-anticipated states of the product.
+  > Many products contain additional components that are not used or executed by the application.
```
sthagen, line 2344: Maybe replace `Many products contain additional components` with `Components`?

`Many` Reads a bit like sales lingo to me. Also, no need for product, might be a tool, package, module, … that has a name, and dead code within.

sthagen (approved):

csaf_2.0/prose/csaf-v2-editor-draft.md
tschmidtb51: Sorry - I'm currently stuck:
Your suggestion was:
```suggestion
  > Components that are not used or executed by the application.
```
Is that correct?

tschmidtb51: BTW: I would that exchange `application` against `product` as everything is a product in CSAF :wink:

sthagen (approved):

santosomar:
@tschmidtb51  - please address a couple of conflicts (highlighted above) with csaf_2.0/prose/csaf-v2-editor-draft.md and I will merge it immediately after.

tschmidtb51:
> Discussed and approved during the monthly TC meeting on 2022-02-23. @tschmidtb51 - please address a couple of conflicts (highlighted above) with csaf_2.0/prose/csaf-v2-editor-draft.md and I will merge it immediately after.

@santosomar already on it :smile:

santosomar:
Thank you so much! Cheers! 🍻🍻🍻

Issue #469, opened 2022-03-02
Simply datetime values (timestamps) by moving to RFC3339 


bernhardreiter:
CSAF-2.0 has ISO 8601 as normative reference.
Allowing a subset of this, specially RFC3339 with the further restriction of only using UTC,
makes the standard easier to parse and write.

Because CSAF is an exchange format for automation in the security context,
there does not seem to be a case for timezones or all iso 8601 variants.

https://datatracker.ietf.org/doc/html/rfc3339 says
```
The following profile of ISO 8601 [ISO8601] dates SHOULD be used in
   new protocols on the Internet.  
```
and 
```
some sites may wish to emit times in UTC only.
```

so the ABNF should be used with this line changed to:
```
 time-offset     = "Z"
```

bernhardreiter:
Another argument for using RFC3339 is ROLIE, as RFC3339 is the normative reference there.

sthagen:
I am for now not in favor of this proposal.

There are no easy choices here as different but not less  problems await. 
I learned from discussions in the converter project that RFC3339 cannot accept `20220302T123456Z` (needing colons and dashes.

That alone renders the that RFC as normative reference "replacing" ISO-8601 quite useless.

bernhardreiter:
Did more reading: the JSON schemas already refer to `date-time`, which is RFC 3339 according to
https://json-schema.org/draft/2020-12/json-schema-validation.html
so most CSAF-2.0 will already be using RFC3339. So this issue is about

 * doing this more clearly in the document and remove ISO 8601 as normative reference and replace it with RFC 3339
 * doing a further simplification by only allowing UTC as indicated by an ending "Z" with out `time-numoffset`

sthagen:
In case we later ask OASIS to submit CSAF v2.0 to ISO it is not convincing for me to remove ISO-8601. The converter arguments seem weak, as we could just use milliseconds since epoch and trust converters to do it right. The usecase of colon free timestamps is always the slugness when using as parts of filenames and URL parts. 

In summary I think we are too late in the process for v2.0 to preform this change.

bernhardreiter:
> The usecase of colon free timestamps is always the slugness when using as parts of filenames and URL parts.

As filename parts a special format maybe preferable, but then ISO 8601 is still having way to many possibilities.
If a format suitable for good filenames is needed, then it IMO probably should be one that has only one variant, very close to RFC3339 with UTC only.

sthagen:
[...]

PR #471, opened 2022-03-02
Change to RFC3339

tschmidtb51:
- resolves oasis-tcs/csaf#469
- remove reference to ISO 8601
- add reference to RFC3339 (which was previously only indirect given through [7.3.1. Dates, Times, and Duration](https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.7.3.1) of JSON Schema validation


sthagen (changes requested):
As per https://github.com/oasis-tcs/csaf/issues/469#issuecomment-1057086703 I propose to not include this change. We may add the RFC reference as informative reference, that I would approve now, as it enhances the documentation.

tschmidtb51:
Then, I close this PR and provide a new one adding RFC 3339 as informative reference.

Issue #480, opened 2022-03-17
JSON schema for ROLIE feed


tschmidtb51:
We should provide an auxiliary JSON schema for CSAF ROLIE feeds. That would aid in the creation of correct ROLIE feeds.

tschmidtb51:
We also should clean up the directory structure.

sthagen:
For now I suggest to add as `optional_json_schema` (or else) because we did not register that schema as additional artifact with the CSAF v2.0 OASIS Standard.

I would use optional for v2.0 as this is the case, and refrain from complicated words many people have spelling problems with :smile:

PR #481, opened 2022-03-17
ROLIE feed schema

tschmidtb51:
- addresses parts of https://github.com/oasis-tcs/csaf/issues/480
- draft first JSON schema for ROLIE feeds
- update timestamps of example to conform format "date-time"
- add test for ROLIE feed schema
- integrate into GitHub actions

### DIFF
--- a/csaf_2.0/json_schema/ROLIE_feed_json_schema.json
+++ b/csaf_2.0/json_schema/ROLIE_feed_json_schema.json
@@ -1,0 +1,292 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://raw.githubusercontent.com/oasis-tcs/csaf/master/csaf_2.0/json_schema/ROLIE_feed_json_schema.json",
+  "title": "ROLIE Feed auxiliary Schema",
+  "description": "Representation of CSAF ROLIE feed as a JSON document.",
+  "$defs": {
+    "json_link_t": {
+      "title": "JSON Link",
+      "description": "Contains the URL of the JSON file.",
+      "type": "string",
+      "format": "uri",
+      "pattern": "^https://.+\\.json$"
+    },
+    "link_t": {
+      "title": "List of Links",
+      "description": "Contains a list of links related to the current context.",
+      "type": "array",
+      "prefixItems": [
+        {
+          "title": "Link",
+          "description": "Specifies the JSON link.",
+          "type": "object",
+          "required": [
+            "rel",
+            "href"
+          ],
+          "properties": {
+            "href": {
+              "title": "Hyper reference",
+              "description": "Contains the URL of the JSON file.",
+              "$ref": "#/$defs/json_link_t"
+            },
+            "rel": {
+              "title": "Relationship",
+              "description": "Contains the relationship value of the link.",
+              "type": "string",
+              "enum": [
+                "self"
+              ]
+            }
+          }
+        }
+      ],
+      "minItems": 1,
+      "uniqueItems": true,
+      "items": {
+        "title": "Link",
+        "description": "Specifies a single link.",
+        "type": "object",
+        "required": [
+          "rel",
+          "href"
+        ],
+        "properties": {
+          "href": {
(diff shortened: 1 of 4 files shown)


sthagen (approved):

santosomar (approved):

santosomar (commented):
WOW! This is amazing! Great work on this. The [ROLIE RFC mentions the support for pagination](https://www.rfc-editor.org/rfc/rfc8322.html#:~:text=A%20Feed%20can%20contain%20an%20arbitrary,immediately%20following%20page%20of%0A%20%20%20%20%20%20the%20Feed.):
```
   o  "first" - Indicates that the "href" attribute value of the link
      identifies a resource URI for the furthest preceding page of
      the Feed.

   o  "last" - Indicates that the "href" attribute value of the link
      identifies a resource URI for the furthest following page of
      the Feed.

   o  "previous" - Indicates that the "href" attribute value of the link
      identifies a resource URI for the immediately preceding page of
      the Feed.

   o  "next" - Indicates that the "href" attribute value of the link
      identifies a resource URI for the immediately following page of
      the Feed.
```
      
We probably don't have to edit the CSAF ROLIE schema, but probably we can mention it in the documentation? If a ROLIE feed has pagination, will it fail validation of this schema?

santosomar:
Since pagination **could** be a problem (probably?), we may need to think about it and document this... The JSON Schema provided doesn't include any specific provisions for pagination. To implement pagination, you would need to provide additional information in the JSON data that represents each feed, such as the number of entries per page and the current page number, etc.

tschmidtb51:
> We probably don't have to edit the CSAF ROLIE schema, but probably we can mention it in the documentation? If a ROLIE feed has pagination, will it fail validation of this schema?

The validation won't fail in the "simple" version but in the "strict" one. The "simple" one will just ignore any property that is not in the schema.

I didn't implement all the fields from [section 6.1](https://www.rfc-editor.org/rfc/rfc8322.html#section-6.1). So the question is: Should we do that? I guess it could be beneficial...

@sthagen, @santosomar: Thoughts?

csaf_2.0/json_schema/ROLIE_feed_json_schema.json
```diff
@@ -0,0 +1,291 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://raw.githubusercontent.com/oasis-tcs/csaf/master/csaf_2.0/json_schema/ROLIE_feed_json_schema.json",
```
sthagen, line 3: Can we try to maintain lowercase file names?

tschmidtb51:
Converting it back to draft to implement missing fields and address changes from reviewers.

oxisto:
Are there any plans to continue this? I could volunteer to pick this up again. I think this would be very helpful to implement ROLIE in kotlin-csaf.

tschmidtb51:
> Are there any plans to continue this? I could volunteer to pick this up again. I think this would be very helpful to implement ROLIE in kotlin-csaf.

Plans: yes, but it has been a low priority for us.
So, if you want to volunteer - go ahead.

oxisto:
> > Are there any plans to continue this? I could volunteer to pick this up again. I think this would be very helpful to implement ROLIE in kotlin-csaf.
> 
> Plans: yes, but it has been a low priority for the editors. So, if you want to volunteer - please go ahead.

Got it. We got it to work in our initial implementation and I will test this with different providers now, if we encounter any issues, I will update the schema accordingly.

tschmidtb51:
Please also have a look at the reviewers' comments and address them in you approach.

oxisto:
> Please also have a look at the reviewers' comments and address them in you approach.
> 
> Also, please move the JSON Schema into the `referenced_schemas/rfc8322`

Got it, I probably need to create a new pull request for this then, since I edit this one.

tschmidtb51:
> Got it. We got it to work in our initial implementation and I will test this with different providers now, if we encounter any issues, I will update the schema accordingly.

Note, that some features might not be used yet but publicly available providers (e.g. pagination).

PR #482, opened 2022-03-18
BSI advisory

tschmidtb51:
add a "real world" example from BSI


santosomar (approved):
This is really good 👍 thank you for providing this perfect example!

sthagen (approved):

PR #504, opened 2022-03-28
Clarify the association of a product to a score

mprpic:
A CVSS score listed for a product should reflect that product's status.
For example, a fixed product version should in most cases have a CVSS score
of 0 since the vulnerability in it no longer exists. Previous version
ranges of that same product can list the CVSS scores as it applies to them.


mprpic:
@tschmidtb51 Review appreciated.

sthagen (approved):

santosomar:
This makes complete sense. Thank you for clarifying this!

tschmidtb51 (commented):
LGTM.

Issue #508, opened 2022-04-07
Example filenames should be lower case?!


bernhardreiter:
The csaf 2.0 filenames for the examples (as of this moment) are:
```
BSI-2022-0001.json
CVE-2018-0171-modified.json
cvrf-rhba-2018-0489-modified.json
```
(https://github.com/oasis-tcs/csaf/tree/f5be5c83d34c24b732f01c054f6d65285922d8a0/csaf_2.0/examples)

According to https://docs.oasis-open.org/csaf/csaf/v2.0/cs01/csaf-v2.0-cs01.html#51-filename , shouldn't they all be lower case?

bernhardreiter:
https://github.com/oasis-tcs/csaf/tree/master/csaf_2.0/test/validator/data has many upper case filenames as well. (Somehow I feel like I am missing somethinge here.)

sthagen:
@bernhardreiter I think we were missing to simplify the names as a coordinated task before producing editor revisions in PRs for adoption by the TC. @tschmidtb51 I suggest we use this feedback (issue) as feedback from public  review and can then be confident that changes we do based on this feedback are considered non-material by the TC so we can ensure the CS02 (presumably following the CSD02/PRSD02) will receive these enhancements (and the repository for the parts not delivered to docs.oasis-open.org)

tschmidtb51:
This issue has been resolved and filename test has been added to check filenames an prevent mistakes.

bernhardreiter:
Mentioned on csaf-comment today: https://groups.oasis-open.org/discussion/other-suggestions-partly-already-handled-395-396-397-398-469-470-477-508

PR #509, opened 2022-04-11
Check filenames

tschmidtb51:
- addresses parts of https://github.com/oasis-tcs/csaf/issues/508
- add script to test filenames of CSAF files
- add Github Action to run test script
- rename CSAF examples
- fix filenames of test file


tschmidtb51:
@mprpic, @juspence: Should we delete `rhba-2018_0489.json` as you provided fresh examples?

mprpic:
> @mprpic, @juspence: Should we delete `rhba-2018_0489.json` as you provided fresh examples?

Yes, please! Would you mind making that change in this MR?

tschmidtb51:
> > @mprpic, @juspence: Should we delete `rhba-2018_0489.json` as you provided fresh examples?
> 
> Yes, please! Would you mind making that change in this MR?

Done in [e6ec249](https://github.com/oasis-tcs/csaf/pull/509/commits/e6ec2494987769e264d70956a6d06a0a300eb38e).

sthagen (approved):
I have another proposal for the python script but we can merge this first to profit from the new functionality already.

sthagen:
Not important, but what about implementing the verification/correction tool like so:
```python
#! /usr/bin/env python3
"""Verify or modify filenames of security advisories against CSAF 2.0 rules."""
import argparse
import json
import pathlib
import re
import sys

ENCODING = 'utf-8'
INVALID_ID = '_invalid'
VALID_NAME_PAT = r'([^+\-_a-z0-9]+)'
SUCC = 'TRUE'
FAIL = 'FALSE'


def load(csaf_path):
    """Load the CSAF data from the path to the JSON file or fail miserably."""
    with open(csaf_path, 'rt', encoding=ENCODING) as handle:
        return json.load(handle)


def dump(csaf_data, csaf_path):
    """Dump the CSAF data formated to 2 space indent to the JSON file given by the path."""
    with open(csaf_path, 'wt', encoding=ENCODING) as f:
        json.dump(csaf_data, f, indent=2)


def compute_filename(csaf_data):
    """Derive the filename from document/tracking/id if exists else return the conventional invalid json name."""
    document_tracking_id = INVALID_ID
    if (doc := csaf_data.get('document')) and (track := doc.get('tracking')) and (the_id := track.get('id')):  # noqa
        document_tracking_id = the_id

    return f'{re.sub(VALID_NAME_PAT, "_", document_tracking_id.lower())}.json'


def api():
    """Implementation of command line API returning parser."""
    parser = argparse.ArgumentParser(description='Verifies or modifies the name of a CSAF 2.0 advisory file')
    parser.add_argument('input_file', type=str, help='CSAF advisory file to verify or modify the filename of')
    parser.add_argument('-p', '--print', dest='echo', action='store_true', help='Prints the correct filename')
    parser.add_argument(
        '-v',
        '--verbose',
        dest='verbose',
        action='store_true',
        help=f'Prints the logic result as either {SUCC} or {FAIL}',
    )
    parser.add_argument(
        '-w',
        '--write',
        dest='add',
        action='store_true',
        help=(
            'Writes the CSAF advisory file to the correct filename if different'
            ' - will overrule -r/--rename if given in addition'
        ),
    )
    parser.add_argument(
        '-r',
        '--rename',
        dest='update',
        action='store_true',
        help=(
            'Renames the CSAF advisory file to the correct filename if necessary'
            ' - will be overruled by -w/--write if given in addition'
        ),
    )
    return parser


def main(argv=None):
    """Drive the verification or modification of advisory file(name)s per CSAF 2.0 rules."""
    argv = sys.argv[1:] if argv is None else argv
    job = api().parse_args(argv)
    data = load(job.input_file)
    current_path = pathlib.Path(job.input_file)
    correct_path = current_path.parent / compute_filename(data)
    ok = current_path == correct_path

    if job.echo:
        print(correct_path.name)
    if job.verbose:
        print(SUCC if ok else FAIL)

    if ok:
        return 0

    if job.add:
        dump(data, correct_path)
    elif job.update:
        _ = current_path.rename(correct_path)

    return 1


if __name__ == '__main__':
    sys.exit(main(sys.argv[1:]))
```

Mostly refactored to avoid overwrites of builtins, streamlined the flow, separated the API,  and added documentation.

Made explicit what happens when given the `-r` and `-w` options for a single call:
When receiving both -r and -w args, the `check.py` of this change set first creates a new file with the corrected name and then tries to rename the input file from the old name to the new name - no problem, but users may not like to have files moved when their fingers added -w in the first place.

Maybe `add` (or `create`) and `update` are better terms for these actions?

Issue #522, opened 2022-04-28
Reference security.txt as RFC9116


tschmidtb51:
# Situation

It looks like the security.txt got a RFC number assigned. If that is stable, we should refer to that as well.

# Proposal

Add a reference to RFC9116.

tschmidtb51:
@sthagen Could you please double check on the RFC status?

sthagen:
Published: https://www.rfc-editor.org/rfc/rfc9116 I think about two days ago (depending on the timezone one is living in): https://datatracker.ietf.org/doc/rfc9116/

PR #532, opened 2022-05-13
RFC 9116

tschmidtb51:
- resolves https://github.com/oasis-tcs/csaf/issues/522
- add RFC 9116 as informative reference
- adopt reference and process status


tschmidtb51:
@sthagen What do you think about [1adcf8f](https://github.com/oasis-tcs/csaf/pull/532/commits/1adcf8f7f65a9b206da30fb9f75deeafd020e459)? Is that ok?

sthagen (approved):

PR #553, opened 2022-05-18
RFC 9116

tschmidtb51:
- addresses parts of oasis-tcs/csaf#522
- rephrase informative comment about RFC 9116 and status of registration (as discussed during the TC meeting)


santosomar (approved):
Looks good and thank you for changing this as discussed during the May 18, 2022 meeting.

Issue #517, opened 2022-04-27
Add missing reference [VEX-Just]


tschmidtb51:
# Situation

The Reference to `[VEX-Just]` is listed in [1.4 Informative References](https://docs.oasis-open.org/csaf/csaf/v2.0/csd02/csaf-v2.0-csd02.html#14-informative-references) as `TBA`.

# Proposal

Either remove the reference or link to a draft version of the document.

Maybe, we should also improve the name of the reference...

tschmidtb51:
Flagging @sparrell, @allan-ntia for attention.

sthagen:
I think the name is OK (VEX and VEX-Just side by side.

PR #556, opened 2022-05-27
VEX Status Justification

tschmidtb51:
- resolves oasis-tcs/csaf#517
- add link to publication (will be available as soon as document is online)


tschmidtb51:
As discussed in the last TC meeting, I added the link.

PR #582, opened 2022-08-29
Committee Specification 03

tschmidtb51:
- add Committee Specification 03 as published to OASIS repo


sthagen (approved):
Does this change require an entry in the change history?

tschmidtb51:
Well, we never did that before - so I don't know... 
But given that this version might be published as OASIS standard, I'm a bit reluctant change that now...

sthagen:
Ah, ok, I was on the wrong train of thought: Thinking this is for publication when instead this is to create a commit with a tag in this repo so people ca find the exact matching version also in our "history" :+1:

PR #585, opened 2022-09-12
Organizing meeting minutes in folders per year

santosomar:
Organizing the meeting minutes directory:
- Created separate directories for each year.
- Moved meeting minute files to the respective directory (i.e., all meeting notes from 2021 to the /2021/ directory, etc.)


sthagen (approved):
lgtm (in case we have added the links to the kavi we might have to update the entries in the documents on the OASIS site.

PR #587, opened 2022-09-26
Proposed three registration content forms for upstreams

sthagen:
- [x] Added .well-known/csaf/ request form content
- [x] Added .well-known/csaf-aggregator/ form content
- [x] Added CSAF as dedicated field in security.txt from content

### DIFF
--- a/csaf_2.0/register/security-dot-txt/csaf.txt
+++ b/csaf_2.0/register/security-dot-txt/csaf.txt
@@ -3,4 +3,4 @@ Description: link to a provider-metadata.json resource of the Common Security Ad
 Multiple Appearances: yes
 Status: current
 Change controller: OASIS-OPEN
-Reference: RFC 9116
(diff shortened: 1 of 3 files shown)


csaf_2.0/register/security-dot-txt/csaf.txt
```diff
@@ -0,0 +1,6 @@
+Field Name: Canonical
```
tschmidtb51, line 1: I didn't check the other ones yet - but shouldn't that be `CSAF` instead of `Canonical` as the field in the security.txt should be called `CSAF`?

csaf_2.0/register/security-dot-txt/csaf.txt
```diff
@@ -0,0 +1,6 @@
...
+Description: link to a provider-metadata.json resource of the Common Security Advisory Framework (CSAF)
+Multiple Appearances: yes
+Status: current
+Change controller: OASIS-OPEN
+Reference: RFC 9116
```
tschmidtb51, line 6: Also not checked, but I guess we need to link to our spec here, correct?

sthagen:
Relates to issues #317 and #318

tschmidtb51 (approved):
LGTM. Thank you for putting those together.

Issue #594, opened 2022-11-29
Bug in the check script


tschmidtb51:
The current script to filenames doesn't conform with the [standard](https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#51-filename) as it allows for multiple consecutive `_`. This should be fixed.

PR #595, opened 2022-11-29
Fix check filenames script

tschmidtb51:
- resolves oasis-tcs/csaf#594
- remove `_` from the regex to fix bug


sthagen (approved):
LGTM. Thanks.

PR #602, opened 2023-01-25
Fix GitHub Actions

### LINKED ISSUE #601: Update GH-Actions
Currently, we use 2 things in GH-Actions that are deprecated:
```
csaf_2.0/test/validator.py:33
DeprecationWarning: Accessing jsonschema.draft202012_format_checker is deprecated and will be removed in a future release. Instead, use the FORMAT_CHECKER attribute on the corresponding Validator.
```
as well as
```
cpe-test
Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2, actions/setup-node@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.
```

tschmidtb51:
- resolves #601 
- adopt validator script to use public API of jsonschema
- update Node Version to current LTS (18)
- update job name

### DIFF
--- a/.github/workflows/cpe.yml
+++ b/.github/workflows/cpe.yml
@@ -11,6 +11,6 @@ jobs:
     - name: Setup Node
       uses: actions/setup-node@v2
       with:
-        node-version: '14'
+        node-version: '18'
     - name: Perform CPE Dictionary Test
       run: ./csaf_2.0/test/cpe/run_tests.sh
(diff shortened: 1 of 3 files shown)


sthagen (approved):
Thanks, bumps and parameter changes look good to me.

Issue #611, opened 2023-02-23
Suggestion to replace `/csaf_2.0/test/filenames/check.py` with  pypi package `paikalta` (filename test)


sthagen:
We might replace [`/csaf_2.0/test/filenames/check.py`](https://github.com/oasis-tcs/csaf/blob/master/csaf_2.0/test/filenames/check.py) with the PyPI package [`paikalta`](https://pypi.org/project/paikalta/) for CSAF filename testing.

@tschmidtb51 ... if I remember correctly what we planned to provide.

PR #612, opened 2023-02-23
Update filename test

tschmidtb51:
- resolves https://github.com/oasis-tcs/csaf/issues/611
- update filename action to use pypi paikalta


csaf_2.0/test/filenames/run_tests.sh
```diff
@@ -10,7 +9,7 @@ cd `dirname $0`/../../..
 
 check() {
   printf "%s" "Testing filename of $1 ... "
-  if python3 ${CHECK} $1; then
+  if paikalta $1; then
```
sthagen, line 12: We could run with `-v` and drop the print statements (for now it would replace `SUCCESS` with `TRUE` and `FAILED` with `FALSE`.

Later we may add user based names per environment or option to the pypi package. I took the choice of these "outcome labels" from check.py ...

sthagen: Anyway, I created "[Feature: Offer user defined SUCC and FAIL labels #2](https://todo.sr.ht/~sthagen/paikalta/2)" to track that "feature"

sthagen: Implemented in upstream tool per version 2023.2.23 which would work here as:

```bash
paikalta --labels SUCCESS,FAILED $1 || FAIL=1
```

replacing the complete if-them-else-fi construct.

tschmidtb51: Thank you - I somehow overlooked that comment... Anyway #613 has been merged.

PR #680, opened 2024-01-30
Seeding CSAF v2.1

tschmidtb51:
@sthagen and @tschmidtb51 prepared a base for the new CSAF 2.1 version. The changes include:
- setting up a CSAF 2.1 directory
- converting schemas to CSAF 2.1
- converting the prose to CSAF 2.1
- splitting the CSAF 2.1 source files into one per section
- adding scripts for generation of the Markdown and HTML version as one file (located at the `csaf_2.1/prose/share` directory)
- adding examples and test from CSAF 2.0 and converting them to CSAF 2.1
- reorganizing test scripts and GitHub actions

PR #681, opened 2024-01-30
Seeding CSAF 2.1

tschmidtb51:
- addresses parts of oasis-tcs/csaf#680
- update citing of CSAF 2.1
- update README with minimum version of tools
- add updated example that was forgotten to save (6.1.22)
- regenerated new CSAF drafts in `share` folder


sthagen (approved):
LGTM, thanks. I think we should always use valid dates. Instead of "?? month 2024" just the date of the PR or so. But no need to fix now.

Issue #687, opened 2024-02-09
Correct `namespace` in example 17


tschmidtb51:
Currently, the `namespace` in example 17 is not valid. It should point to `https://spdx.github.io/spdx-spec/latest/document-creation-information/#65-spdx-document-namespace-field` instead of `https://spdx.github.io/spdx-spec/document-creation-information/#65-spdx-document-namespace-field`

Issue #684, opened 2024-02-06
Add reference to RFC8322


tschmidtb51:
Currently, we have references to RFC8322 but do not mention it in the reference section...

Issue #626, opened 2023-05-23
Appendix C: Raise file size softlimit


tschmidtb51:
The practice has shown that the current file size soft limit, given through [Appendix C.1](https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#c1-file-size) does not meet the needs in practice. At least Red Hat is generating files that are bigger than 15 MB.
Therefore, I suggest to raise the file size limit to 50 MB.

sthagen:
During the 2023-09-27 meeting of the TC the members approved to raise the limit to 50 Megabytes as proposed in CSAF v2.1.

PR #691, opened 2024-02-16
Editorial 2024 02 28

tschmidtb51:
- add RFC 8322 as informative reference (#684)
- correct namespace in example 17 (#687)
- update date
- update revision history
- update citation format
- update size limit and unit (#626)
- generate artifacts

### DIFF
--- a/csaf_2.1/prose/edit/src/guidance-on-size.md
+++ b/csaf_2.1/prose/edit/src/guidance-on-size.md
@@ -36,7 +36,7 @@ All _CSAF producers_ SHOULD NOT produce CSAF documents which exceed those limits
 ## File size
 
 A CSAF document in the specified JSON format encoded in UTF-8 SHOULD conform to known size limits of current technologies parsing JSON content,
-e.g.: 15 MB.
+e.g.: 50 MiB.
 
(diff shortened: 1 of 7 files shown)


tschmidtb51:
@sthagen: It looks like, the end of the HTML file is missing. Could be the script or my computer :see_no_evil:

csaf_2.1/prose/edit/src/guidance-on-size.md
```diff
@@ -36,7 +36,7 @@ All _CSAF producers_ SHOULD NOT produce CSAF documents which exceed those limits
 ## File size
 
 A CSAF document in the specified JSON format encoded in UTF-8 SHOULD conform to known size limits of current technologies parsing JSON content,
-e.g.: 15 MB.
+e.g.: 50 MiB.
 
 > At least one database technology in wide use for storing CSAF documents rejects insert attempts when
```
sthagen, line 41: I will amend the following note, as the relation between the 50 Megabytes and the 16 Megabytes is not as clear, as the former reasoning of less than 16 Megabytes leading to a 15 Megabytes soft limit.

sthagen (approved):
I think this draft is ready for review by the TC.
Great work, @tschmidtb51! Thanks a lot.

Issue #629, opened 2023-05-26
Clarify Markdown


tschmidtb51:
We should clarify that we are thinking about Github-flavored Markdown and which fields are allowed. As a starting point, I suggest the following fields:

```
  /document/acknowledgments[]/summary
  /document/distribution/text
  /document/notes[]/text
  /document/publisher/issuing_authority
  /document/references[]/summary
  /document/tracking/revision_history[]/summary
  /product_tree/product_groups[]/summary
  /vulnerabilities[]/acknowledgments[]/summary
  /vulnerabilities[]/involvements[]/summary
  /vulnerabilities[]/notes[]/text
  /vulnerabilities[]/references[]/summary
  /vulnerabilities[]/remediations[]/details
  /vulnerabilities[]/remediations[]/entitlements[]
  /vulnerabilities[]/remediations[]/restart_required/details
  /vulnerabilities[]/threats[]/details
```

sthagen:
During the 2023-09-27 meeting of the TC the members approved to implement the proposal on CSAF v2.1.

PR #697, opened 2024-02-24
Clarify markdown

tschmidtb51:
- addresses parts of https://github.com/oasis-tcs/csaf/issues/629
- clearly state that GitHub-flavoured Markdown is allowed
- explicitly mention fields that may contain Markdown


sthagen (approved):
LGTM.

Issue #579, opened 2022-07-29
Write purl instead of PURL


tschmidtb51:
The seems to prefer `purl` over PURL: https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst

sthagen:
I am not sure if we should follow inconsequential case "wishes" (just look at the filename 🙈 `PURL-SPECIFICATION.rst`

PR #701, opened 2024-02-24
purl

tschmidtb51:
- resolves oasis-tcs/csaf#579
- use purl as lowercase (but not the references)

### DIFF
--- a/csaf_2.1/prose/edit/src/tests-01-mndtr-13-purl.md
+++ b/csaf_2.1/prose/edit/src/tests-01-mndtr-13-purl.md
@@ -1,6 +1,6 @@
-### PURL
+### purl
 
-It MUST be tested that given PURL is valid.
+It MUST be tested that given purl is valid.
 
 The relevant paths for this test are:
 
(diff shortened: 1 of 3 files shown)


csaf_2.1/prose/edit/src/tests-01-mndtr-13-purl.md
```diff
@@ -1,6 +1,6 @@
-### PURL
+### purl
```
sthagen, line 1: Nit: I think, in titles we should add Title Text before any artificial term, as this looks wrong to me (a single word title in non-title case).

sthagen (approved):
I am +0 on this one.

csaf_2.1/prose/edit/src/tests-01-mndtr-13-purl.md
tschmidtb51: I was also wondering but wanted to flag it so that we don't accidentally forget it - let's revert it:
```suggestion
### PURL

tschmidtb51: If you are okay with that, feel free to commit the change.

PR #703, opened 2024-02-26
fix broken link

wurstbrot:


tschmidtb51:
@mprpic Could you please double-check?

tschmidtb51:
@wurstbrot Thank you for your contribution. Due to the rules of OASIS, we might not be able to accept your suggestion via a Pull Request (even though we known it is often easier to create a PR than trying to describe where a file is). Knowing that you can't use the formal way via csaf-comment mailing list right now, we will discuss the issue.

I'll set the PR to draft for now. Due to the formal rules, a TC member might need to create the PR. Nevertheless, again: Thank you for your contribution.

tschmidtb51:
@sthagen, @santosomar, @mprpic: I would suggest to add a disclaimer to the file instead.

mprpic:
Red Hat no longer publishes CVRF files as of Sep of 2023, and the API endpoint this PR points to will also be removed at some point: https://access.redhat.com/articles/124913.

PR #710, opened 2024-03-19
fix: escape slashes in regular expressions

ctron:
It looks like some regexp engines do have an issue with slashes not being escaped. This change fixes this by also escaping slashes.

Issue #693, opened 2024-02-23
CPE pattern


tschmidtb51:
The current CPE regex seems to have some issues:

1. The CPE version are not restricted correctly. Instead of `^(CPE2.3|CPE2.2)$`, the pattern uses `^(CPE2.3)|(CPE2.2)$`. This allows values like `NOTAVALIDCPEcpe:/o:example:a:42::anyother` resp. (appended ones in CPE2.3).
2. [Some implementations (e.g. deno)](https://github.com/secvisogram/csaf-validator-lib/issues/156) don't like the `/` as un-escaped character in JSON patterns.

Regarding 1:
- That is definitely something that we need to address in CSAF 2.1.

Regarding 2:
- Our test scripts (using `python` and `nodejs`) haven't complained so far. So we need to find out whether an un-escaped `/` is valid in JSON patterns or just accepted. (This means reading the standard.)
- We need to consider improving the situation in CSAF 2.1 by adding the `\\` to escape `/`.

tschmidtb51:
To the best of our knowledge: This has been resolved.

PR #711, opened 2024-03-27
CPE and purl regex

tschmidtb51:
- supersedes #710  
- addresses parts of https://github.com/oasis-tcs/csaf/issues/693, https://github.com/oasis-tcs/csaf/pull/710
- correct pattern to `^((CPE2.3)|(CPE2.2))$`
- add additional `\\` to escape previously unescaped `/` (for CPE and purl)
- correct parsing of CPE 2.3 Dictionary (to also capture endings `">` instead of just `"/>`)
- add new local test cases
- adopt test script


tschmidtb51:
@ctron Please review

santosomar (approved):
Thank you for incorporating these and for following up on the other PR.

santosomar:
Since this is correcting a bug, there is no need to bring it into voting. I will merge the PR now.

PR #712, opened 2024-03-27
CPE tests

tschmidtb51:
- addresses parts of oasis-tcs/csaf#710
- backport of 61e78c8
- correct parsing of CPE 2.3 Dictionary (to also capture endings `">` instead of just `"/>`)


sthagen (approved):
LGTM.

PR #720, opened 2024-04-22
Mandatory TLP 2.0

tschmidtb51:
### Addresses parts of https://github.com/oasis-tcs/csaf/issues/591
- change schema to use TLP 2.0
- adapt prose to reflect schema
- add TLP:AMBER+STRICT in appropriate places
- add conversion rule
- update PMD schema
- update example files
### Addresses parts of https://github.com/oasis-tcs/csaf/issues/633
- adapt prose to reflect schema
- add `/document/distribution/tlp/label` in CSAF base profile
- mark test 6.2.10 as deprecated
- add conversion rule
- correct examples
- add TLP label
- add `/document/distribution/tlp` to validator test data
- add `/document/distribution/tlp` to filename test data
- adapt prose to reflect current test files
- exclude deprecated test 6.2.10

PR #714, opened 2024-03-27
Editor revision for TC meeting 2024-03-27

tschmidtb51:
- Editor revision for CSAF 2.1 (based on #699)
- addresses #693 
- fixes #710 
- fixes #708 
- fixes #579 
- addresses parts of #652

Issue #678, opened 2024-01-08
Warning/Error for signature expirations


ctron:
When documents are signed with a key, that key/certificate may expire, making the validation of such signatures problematic.

I believe there should be a grace period in the specification which requires that there is "enough" time left. And while one can argue about the exact number, I think it would make sense to have two requirements in the spec:

* Signatures MUST be valid for 30 more days
* Signatures SHOULD be valid for 90 more days

santosomar:
Comment from the TC (Denny) : "Signatures MUST be valid for at least 30"

santosomar:
As discussed in the CSAF TC monthly meeting on 2024-02-28, this will be added to the guidance documentation for CSAF 2.0.

tschmidtb51:
As discussed in the CSAF TC monthly meeting on 2024-04-24, the wording from #724 will be added to CSAF 2.1.

tschmidtb51:
@ctron  The [comments mailing](https://groups.oasis-open.org/communities/community-home?communitykey=3720cda2-3056-4183-a759-018f5aa7b5f2) list is now back online. Please formally announce your suggestion there, e.g. through "Please see our suggest in Github Issue XYZ (https://github.com/oasis-tcs/csaf/issues/XYZ)."

Thank you!

tschmidtb51:
The issue was suggested on the mailing list: https://groups.oasis-open.org/discussion/warningerror-for-signature-expirations

tschmidtb51:
The issue was resolved with #723 and #724

Issue #673, opened 2023-12-21
Add conformance target "CSAF-2.0-CSAF-2.1 converter"


tschmidtb51:
We need to add a conformance target for a converter that takes CSAF 2.0 as input and converts it into CSAF 2.1. That also requires that we formulate all conversion rules for each change.

tschmidtb51:
- [ ] mandate run of extended validator

PR #723, opened 2024-04-23
Editor revision for TC meeting 2024-04-24

### LINKED ISSUE #591: Update CSAF to use TLP 2.0
I'm not sure of all of the differences between TLP 1.0 and 2.0, but the one that immediately jumps out at me when reading the CSAF 2.0 spec and the TLP 2.0 spec is that TLP 2.0 has added TLP:AMBER+STRICT.  I also see that TLP:CLEAR replaces TLP:WHITE.  

TLP:AMBER+STRICT is the kind of thing I believe many commercial orgs would want to use instead of TPL:AMBER (my company certainly would).

### LINKED ISSUE #706: Add "Preconditions" item
There are different note categories in the Vulnerabilities -> Vulnerability Notes section. We noticed that the “Preconditions” item is missing. Preconditions can be local privileges, executions rights, physical access, etc. 

Of course you can solve this with the “Details” item, but we would highlight this as a separate category

### LINKED ISSUE #718: Add a new category "Platform" to the Product Branch
Add a new category "platform" to the product branch. The use case is to convey whether the platform is an Intel/AMD/ARM etc.  While ARM can differentiate via the architecture, the architecture for Intel and AMD would both be AMD64. This makes a difference whether an AMD or Intel processor/mother board is used, it will have a differentiate BIOS versions. Especially, if only one is affected by a vulnerability. 

cc: @jdstefaniak

tschmidtb51:
- Editor revision for CSAF 2.1 (based on #714)
- addresses #720 
- fixes #591
- addresses #678 
- fixes #706 
- fixes #718
- addresses parts of #673


sthagen (approved):
LGTM, all 305 files, 😅

santosomar:
The motion to accept the Editor Revision 2024-04-24 as suggested in this PR has passed. This motion is independent of the potential generation of a rendered version of the document which will be located in the csaf_2.1/prose/share folder.
The email archive for this motion can be accessed at:
https://groups.oasis-open.org/discussion/motion-for-723#bm30ffc20c-d0b1-4368-878b-1c12cf6aaa37

Issue #721, opened 2024-04-22
Set `TLP:CLEAR` as default 


tschmidtb51:
We are expecting that most CSAF documents will be freely accessible. So, I guess we should make `TLP:CLEAR` a default value in the schema.

santosomar:
A motion will be put in place via email after the CSAF TC meeting on April 24, 2024.

tschmidtb51:
Note: Setting the default value in the schema does not change the ability to publish CSAF documents with other TLP labels. It also does not change existing CSAF documents with a given TLP label. However, it will provide a suggested default for anyone writing CSAF documents (where as the majority of those will be public.) This enables better adoption as a newcomer does not have to figure out what TLP is, but can use the default value.

santosomar:
The motion has passed. 
https://groups.oasis-open.org/discussion/motion-for-721

PR #734, opened 2024-05-23
`TLP:CLEAR` as default

tschmidtb51:
- resolves https://github.com/oasis-tcs/csaf/issues/721
- set TLP label `CLEAR` as default
- add reasoning


sthagen (approved):
LGTM. Thank You.

Issue #647, opened 2023-08-23
Clarify 6.1.14: same timestamps


tschmidtb51:
Currently, the standard does not explicit state what should happen, if the exact same timestamp occurs in two entries. We need to clarify that the `number` needs to be the second level sorting field.

This is related to #628.

Alternative: Prohibit same timestamps.

Issue #628, opened 2023-05-26
Clarify 6.1.16: same timestamps


tschmidtb51:
Currently, the standard does not explicit state what should happen, if the exact same timestamp occurs in two entries. We need to clarify that the `number` needs to be the second level sorting field.

tschmidtb51:
During the 2023-08-30 meeting of the TC the members approved the motion to add `number` as a second level sorting field.

PR #736, opened 2024-05-25
Same timestamps

tschmidtb51:
- resolves https://github.com/oasis-tcs/csaf/issues/628
- add `number` as second level sorting field in 6.1.16
- unify phrasing to match 6.1.14
- resolves https://github.com/oasis-tcs/csaf/issues/647
- add `number` as second level sorting field in 6.1.14


sthagen (approved):
LGTM.

Issue #635, opened 2023-06-28
Clarify that blocking on user-agents is not allowed


tschmidtb51:
We came across [a situation](https://github.com/csaf-poc/csaf_distribution/issues/376) where a Web Application Firewall blocked the automatic retrieval of the PMD and CSAF files. Given the reasoning in https://github.com/csaf-poc/csaf_distribution/issues/376#issuecomment-1611571389 IMHO CSAF 2.0 already prohibits restricting the user-agent as it contradicts the intended usage.

Nevertheless, we should clearly state, that restricting the user-agent to a specific value (or set of values) is not allowed as it hinders the implementation of tools.

sthagen:
During the 2023-09-27 meeting of the TC the members approved the motion to clearly state in the v2.1 CSAF with an explicit statement that such blocking is not allowed and also add that statement to the FAQs.

tschmidtb51:
Todo:
- [ ] add requirement to CSAF 2.1
- [ ] add security consideration to CSAF 2.1
- [ ] add FAQ to CSAF 2.0

PR #742, opened 2024-05-27
HTTP User-Agents

tschmidtb51:
- addresses parts of oasis-tcs/csaf#635
- add new requirement explicitly stating that no blocking is allowed

### DIFF
--- a/csaf_2.1/prose/edit/src/distributing.md
+++ b/csaf_2.1/prose/edit/src/distributing.md
@@ -560,14 +560,16 @@ Each such folder MUST at least:
 
 ### Requirement 24: HTTP User-Agent
 
-Access to the files provided, both CSAF metadata and documents, MUST NOT be restricted by the value of HTTP User-Agent.
+Access to the CSAF related files and directories provided, for both metadata and documents, MUST be allowed independent of the
+value of HTTP User-Agent.
 
 > Limit the value of HTTP User-Agents to a certain set would hinder adoption of tools retrieving the files.
 
 The only exception is that the temporary blocking of certain HTTP User-Agents is allowed to mitigate an ongoing security incident
-(e.g. a DoS attack on the web server serving the CSAF files). However, it MUST be checked if a less severe measure with a similar
-effect could be used. White-listing CSAF related files and directories MUST be considered. The temporary blocking MUST be removed
-as soon as possible, at latest two weeks after the security incident process was completed.
+(e.g. a DoS attack on the web server serving the CSAF files).
+However, it MUST be checked if a less severe measure with a similar effect could be used.
+It MUST be considered to exempt CSAF related files and directories from such a the temporary blocking.
+The temporary blocking MUST be removed as soon as possible, at latest two weeks after the security incident process was completed.
 
 > Also confer to the TC's guidance on content delivery networks and caching.
 
--- a/csaf_2.1/prose/edit/src/distributing.md
+++ b/csaf_2.1/prose/edit/src/distributing.md
@@ -567,9 +567,9 @@ value of HTTP User-Agent.
 
 The only exception is that the temporary blocking of certain HTTP User-Agents is allowed to mitigate an ongoing security incident
 (e.g. a DoS attack on the web server serving the CSAF files).
-However, it MUST be checked if a less severe measure with a similar effect could be used.
-It MUST be considered to exempt CSAF related files and directories from such a the temporary blocking.
-The temporary blocking MUST be removed as soon as possible, at latest two weeks after the security incident process was completed.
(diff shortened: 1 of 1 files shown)


csaf_2.1/prose/edit/src/distributing.md
```diff
@@ -558,6 +558,19 @@ Each such folder MUST at least:
   }
 ```
 
+### Requirement 24: HTTP User-Agent
+
+Access to the files provided, both CSAF metadata and documents, MUST NOT be restricted by the value of HTTP User-Agent.
```
sthagen, line 563: I will try to provide an alernative version of this section soon (but not now) trying to mitigate my two concerns:

1. Use of `MUST NOT` instead of phrasing a `MUST` sentence.
2. High `MUST` density in sentences with "escape clauses" like
   > MUST be checked if ...

   and:

   > MUST be considered

In addition, we might follow the fashion (as with TLP) to use an alternative term to white-listing.

tschmidtb51: Thank you for your comments. I addressed 1 and your addition in [df33caa](https://github.com/oasis-tcs/csaf/pull/742/commits/df33caa90cafe4eea5d306040a9c64c0236ea430).

tschmidtb51: Unfortunately, I have no idea how to address 2 without loosing clarity.

csaf_2.1/prose/edit/src/distributing.md
```diff
@@ -560,14 +560,16 @@ Each such folder MUST at least:
...
+value of HTTP User-Agent.
 
 > Limit the value of HTTP User-Agents to a certain set would hinder adoption of tools retrieving the files.
 
 The only exception is that the temporary blocking of certain HTTP User-Agents is allowed to mitigate an ongoing security incident
-(e.g. a DoS attack on the web server serving the CSAF files). However, it MUST be checked if a less severe measure with a similar
-effect could be used. White-listing CSAF related files and directories MUST be considered. The temporary blocking MUST be removed
-as soon as possible, at latest two weeks after the security incident process was completed.
+(e.g. a DoS attack on the web server serving the CSAF files).
+However, it MUST be checked if a less severe measure with a similar effect could be used.
```
sthagen, line 570: The rewrites look good to me, but I think these three MUSTs shall become SHOULDs:

>However, a less severe measure with a similar effect SHOULD be used.
>CSAF related files and directories SHOULD be exempted from temporary blocking.
>The temporary blocking SHOULD be removed as soon as possible, at latest two weeks after the security incident process was completed.

I reordered and simplified the sentences as well as implicitly fixed a nit (`a the`).

sthagen (approved):
LGTM - when considering the MUST to SHOULD including the simplifications (or similarly).
Maybe we should ask the native English speakers for support, as also the version I came up with to me reads still between clumsy, lamenting, and wishful thinking 😁

csaf_2.1/prose/edit/src/distributing.md
tschmidtb51: Hmm, I think I like the MUST better (especially thinking of my service provider) as it enforces (potentially with a documentation part, at least that is what I would require) to:
1. See whether less severe measures could be used.
2. Consider exempts
3. Remove the blocking at some point

I guess that is something the TC needs to decide what the best option is.

tschmidtb51: @santosomar, @justmurphy, @mprpic: Looking for input from native speakers to find the best wording. Please also state your opinion whether the "SHOULD" option or the "MUST consider" (which allows to easily request reasoning as consumer) would be preferred.

sthagen: In my understanding SHOULD already means "there have to be compelling reasons not to do as generally this is the way to go).

tschmidtb51: The wording has been changed to SHOULD as discussed in today's TC meeting.

csaf_2.1/prose/edit/src/distributing.md
tschmidtb51: Addressed in [65dea4f](https://github.com/oasis-tcs/csaf/pull/742/commits/65dea4f650f19f4e5bd6599038c82e3e9fdf87b7)

sthagen (approved):
LGTM.

santosomar (approved):

tschmidtb51:
The [motion](https://groups.oasis-open.org/discussion/motion-for-742) to accept the pull request as suggested in https://github.com/oasis-tcs/csaf/pull/742 and include it into CSAF 2.1, has passed. It automatically carried on 2024-06-05 23:00 UTC.

Issue #627, opened 2023-05-26
Add test: same timestamps in revision history


tschmidtb51:
We should add an (optional?) test that warns, if two timestamps are discovered in the revision history that are the same.

PR #744, opened 2024-05-27
Optional Test: Same Timestamps in Revision History

tschmidtb51:
- resolves oasis-tcs/csaf#627
- add optional test that warns if two timestamps in the revision history are the same
- add invalid examples for 6.2.21
- add valid examples for 6.2.21
- update testcase and schema for testcases


csaf_2.1/prose/edit/src/tests-02-optional.md
```diff
@@ -645,3 +645,32 @@ The relevant path for this test is:
 > The key `custom_property` is not defined in the JSON schema.
 
 > A tool MAY remove such keys as a quick fix.
+
+### Same Timestamps in Revision History
+
+It MUST be tested that the timestamps of all items in the revision history are pairwise disjoint, when converted to UTC.
```
sthagen, line 651: Could we write instead:

>When converted to UTC, the timestamps of all items in the revision history MUST be pairwise disjoint.

sthagen (approved):
LGTM (only one rewording maybe).

csaf_2.1/prose/edit/src/tests-02-optional.md
tschmidtb51: Usually, we include the phrase "it MUST be tested". Could we write:

```suggestion
It MUST be tested that the timestamps converted into UTC of all items in the revision history are pairwise disjoint.
``

Comment (tschmidtb51) [reply csaf_2.1/prose/edit/src/tests-02-optional.md:651]: Or:
```suggestion
It MUST be tested that the timestamps, when converted into UTC, of all items in the revision history are pairwise disjoint.
```

tschmidtb51: Or:
```suggestion
It MUST be tested that, when converted into UTC, the timestamps of all items in the revision history are pairwise disjoint.
```

sthagen: I like to keep the focus on the compliance fact it self and not the on out of some means of compliance: My understanding is here that we provide the test actor a list of claims they can validate and not so much a MUST on theses actors to test. The motivation to test I see as external to our specifications. But, I may be wrong.

tschmidtb51: @sthagen: The reason, why I would like to add the "It MUST be tested that" is that we use it in almost all tests in CSAF 2.0 (cf. https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#618-invalid-cvss and following). I think removing the phrase might raise questions if the implementers have to do different things here (which is not the case).

tschmidtb51: We could put that on the table for our native speakers today.

sthagen: Thank you for adding to discussion. I think we should remove that MUST on the test activity to the MUST on the properties of the document (or where applicable to the behaviors of the roles under validation) instead

tschmidtb51: @sthagen: As we discussed in today's call (TC meeting) regarding the phrase "it MUST be tested that" that it affects all test. Therefore, IMHO it should be handled in a separate issue. This would allow us to process this PR. Do you agree?

sthagen: The remaining loophole for inconsistencies sneaking into implementations now is IMO the "when converted to", what if no conversion is done and consumers use these values as mere textual identifiers which can be have as many colliding values as timezones are available?
But, in case this is also spread across other sections we can easily also split dealing with that concern into another task (PR).

tschmidtb51:
@sthagen, @santosomar: I added a sentence for implementers regarding the UTC conversion. I think that should resolve the discussion.

sthagen (approved):
👍🏽

csaf_2.1/prose/edit/src/tests-02-optional.md
tschmidtb51: Thanks. This is addressed in [f0d17e9](https://github.com/oasis-tcs/csaf/pull/744/commits/f0d17e9dda3b2a92eec60793752c9861dd0dc5e7)

tschmidtb51:
After discussing this with @sthagen again, I realized that it is not needed to enforce the conversion to UTC. It is enough, if the comparison takes timestamps into account (see: [ab3c9b2](https://github.com/oasis-tcs/csaf/pull/744/commits/ab3c9b205aa88a3155bc85e0d6f74f0ce4209640))

sthagen (approved):
Ship it! 😁

santosomar (approved):
Thank you 🙏

santosomar:
The motion to accept the pull request as suggested in https://github.com/oasis-tcs/csaf/pull/744 and include it into CSAF 2.1, has passed. It automatically carried on 2024-06-07 23:00 UTC. https://groups.oasis-open.org/discussion/motion-for-744

santosomar (approved):

Issue #580, opened 2022-08-15
Add new party `multiplier`


tschmidtb51:
There is a use case where an ISAC / National CERT / ... could act as multiplier: Taking the original CSAF document (adding information relevant to the context) and republishing the information. Currently, those parties use the `other` status. We should create a separate value `multiplier` (or `republisher`)  for them.

tschmidtb51:
The TC needs to decide if that is in scope for CSAF 2.1.

tschmidtb51:
Thomas Schmidt proposed a motion, as detailed in this [OASIS mailing list archive](https://groups.oasis-open.org/discussion/motion-580), to add a party `multiplier` to CSAF v2.1. Omar Santos seconded the motion.

tschmidtb51:
@mprpic: is that a comment or an objection (regarding motion https://groups.oasis-open.org/discussion/motion-580)?

tschmidtb51:
> I personally would prefer the wording `republisher` since "multiplier" implies the artifact is replicated into multiple (different) copies and re-distributed. Otherwise, lgtm!

Personally, I prefer `multiplier` as it is closer to the multiplicator and describes IMHO clearly that the goal is to bring it to more people/groups / provide it to a wider (or also other) community.

And yes, that might involve multiple copies that might differ in the information (e.g. the E-ISAC might add different information than the Water-ISAC).

mprpic:
@tschmidtb51 It's just a comment, I'm fine with either. I was just voicing my preference but if everyone else is good with `multiplier` then so be it :-)

santosomar:
The motion to add a new party "multiplier" as suggested in https://github.com/oasis-tcs/csaf/issues/580 to CSAF 2.1 has passed. No objection was received and the motion automatically passed on 2024-04-03 20:00 UTC.

The motion and results are available at: https://groups.oasis-open.org/discussion/motion-580

tschmidtb51:
Solved with #741

PR #747, opened 2024-05-29
Add party `multiplier`

tschmidtb51:
- addresses parts of https://github.com/oasis-tcs/csaf/issues/580
- add new party "multiplier" to JSON schema
- adopt prose to reflect schema
- add conversion rule

### DIFF
--- a/csaf_2.1/prose/edit/src/conformance.md
+++ b/csaf_2.1/prose/edit/src/conformance.md
@@ -520,6 +520,9 @@ Secondly, the program fulfills the following for all items of:
   > This is a common case for CSAF 2.0 documents labeled as TLP:RED but actually intended to be TLP:AMBER+STRICT.
 
   If no TLP label was given, the CSAF 2.0 to CSAF 2.1 converter SHOULD assign `TLP:CLEAR` and output a warning that the default TLP has been set.
+* `/document/publisher/category`: If the value is `other`, the CSAF 2.0 to CSAF 2.1 converter SHOULD output a warning that some parties have
+  been separated into the new value `multiplier`. An option to suppress this warning MUST exist. In addition, an option SHOULD be provided to
+  set the value to `multiplier`.
 
 > A tool MAY implement options to convert other Markdown formats to GitHub-flavoured Markdown.
 
--- a/csaf_2.1/prose/edit/src/conformance.md
+++ b/csaf_2.1/prose/edit/src/conformance.md
@@ -521,7 +521,7 @@ Secondly, the program fulfills the following for all items of:
 
   If no TLP label was given, the CSAF 2.0 to CSAF 2.1 converter SHOULD assign `TLP:CLEAR` and output a warning that the default TLP has been set.
 * `/document/publisher/category`: If the value is `other`, the CSAF 2.0 to CSAF 2.1 converter SHOULD output a warning that some parties have
-  been separated into the new value `multiplier`. An option to suppress this warning MUST exist. In addition, an option SHOULD be provided to
+  been regrouped into the new value `multiplier`. An option to suppress this warning MUST exist. In addition, an option SHOULD be provided to
   set the value to `multiplier`.
 
 > A tool MAY implement options to convert other Markdown formats to GitHub-flavoured Markdown.
--- a/csaf_2.1/prose/edit/src/schema-elements-02-props-01-document.md
+++ b/csaf_2.1/prose/edit/src/schema-elements-02-props-01-document.md
@@ -296,7 +296,7 @@ The value `discoverer` indicates individuals or organizations that find vulnerab
 This includes all manner of researchers.
 
 The value `multiplier` indicates individuals or organizations that use existing CSAF documents or information that could
-be represented in CSAF and create their own CSAF documents for distribution to specific target audience.
+be represented in CSAF, and create their own CSAF documents for distribution to a specific target audience.
 A single multiplier might have target audiences.
 > For example, a National CSIRT might create different CSAF documents for the same vulnerability for critical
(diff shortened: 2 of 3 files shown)


tschmidtb51:
@mprpic Please review.

csaf_2.1/prose/edit/src/schema-elements-02-props-01-document.md
```diff
@@ -294,10 +295,20 @@ This includes all Computer Emergency/Incident Response Teams (CERTs/CIRTs) or ag
 The value `discoverer` indicates individuals or organizations that find vulnerabilities or security weaknesses.
 This includes all manner of researchers.
 
+The value `multiplier` indicates individuals or organizations that use existing CSAF documents or information that could
+be represented in CSAF and create their own CSAF documents for distribution to specific target audience.
```
mprpic, line 299:
```suggestion
be represented in CSAF, and create their own CSAF documents for distribution to a specific target audience.
```

csaf_2.1/prose/edit/src/schema-elements-02-props-01-document.md
```diff
@@ -294,10 +295,20 @@ This includes all Computer Emergency/Incident Response Teams (CERTs/CIRTs) or ag
...
+A single multiplier might have target audiences.
+> For example, a National CSIRT might create different CSAF documents for the same vulnerability for critical
+  infrastructure companies in different sectors, government agencies, non-critical industry and the public based on
```
mprpic, line 302:
```suggestion
  infrastructure companies in different sectors, government agencies, non-critical industry, and the public based on
```

mprpic (approved):
Just a couple minor nitpicks. LGTM!

csaf_2.1/prose/edit/src/conformance.md
```diff
@@ -520,6 +520,9 @@ Secondly, the program fulfills the following for all items of:
   > This is a common case for CSAF 2.0 documents labeled as TLP:RED but actually intended to be TLP:AMBER+STRICT.
 
   If no TLP label was given, the CSAF 2.0 to CSAF 2.1 converter SHOULD assign `TLP:CLEAR` and output a warning that the default TLP has been set.
+* `/document/publisher/category`: If the value is `other`, the CSAF 2.0 to CSAF 2.1 converter SHOULD output a warning that some parties have
+  been separated into the new value `multiplier`. An option to suppress this warning MUST exist. In addition, an option SHOULD be provided to
```
sthagen, line 524: Question: I try to figure out what it means to separate multiple parties into the new value `multiplier`. Maybe an example would help?

csaf_2.1/prose/edit/src/conformance.md
```diff
@@ -520,6 +520,9 @@ Secondly, the program fulfills the following for all items of:
...
+  set the value to `multiplier`.
 
 > A tool MAY implement options to convert other Markdown formats to GitHub-flavoured Markdown.
```
sthagen, line 527: Suggest to use flavored instead (American spelling to stay close to simplified English).

sthagen (approved):
LGTM (plus minus some nits and maybe an illustrating example). Thanks for providing the text.

csaf_2.1/prose/edit/src/conformance.md
tschmidtb51: IMHO, those three parties are doing more or less the same:
- **forwarder**: just replace the publisher details but basically keep the rest of the original CSAF file
- **republisher**: forwarder that also might convert a "non-CSAF document" of a 3rd-party into a CSAF document
- **multiplier**: republisher that also adds additional context for a specific target group and might publish multiple versions for different audiences

IMHO, it is hard in practice to differentiate them as the borders are fluent:
- Sometimes, a **multiplier** might take a CSAF document as input and has no additional insights (e.g. the original CSAF document already covers the case very well), so he would technically speaking be a **forwarder**.
-  Or, a **multiplier** might take a non-CSAF document as input and has no additional insights (e.g. the original document already covers the case very well), so he would technically speaking be a **republisher**.

Determining that automatically, whether someone changed something (especially when converting from human-readable documents) or whether just the fact that it gets republished by a certain party is already adding context seems to be very complex. Moreover, this might change with every CSAF document (or even between versions of the same CSAF document). Therefore, I do not see a benefit in making 3 separate roles. (Also, for the PMD one has to decide for one category anyway.)

@sthagen Does that clarify the reasoning?

sthagen: @tschmidtb51 Thank you for providing the explanation of the concept and rationale.
That all reads reasonable to me.

My problem with the sentence is not specific to CSAF roles, sorry for not being clear in the first place.

Question put differently: Should a converter state that for the category source value `other` has been mapped in the target to `multiplier`? 

Because separating many into one reads wrong (the action/verb).

sthagen (approved):
LGTM.

csaf_2.1/prose/edit/src/conformance.md
tschmidtb51: @sthagen Thanks for pointing that out. This has been addressed by rephrasing in [4049a91](https://github.com/oasis-tcs/csaf/pull/747/commits/4049a916526ceb3108cd59f1ea71c8d8db8cd7b0)

sthagen (approved):
Nice. Thanks.

santosomar:
The motion to accept the pull request as suggested in https://github.com/oasis-tcs/csaf/pull/747 and include it into CSAF 2.1, has passed. It automatically carried on 2024-06-07 23:00 UTC.
https://groups.oasis-open.org/discussion/motion-for-747

santosomar (approved):

PR #748, opened 2024-05-29
Editorial

tschmidtb51:
- addresses review comment from oasis-tcs/csaf#747
- use American English for "flavored"


sthagen (approved):
LGTM - although two sentences in a row - both exposing a MAY - do make me wonder if these guide or confuse implementers. Maybe we should scan the complete specification (not in this PR) and verify our use of MAY is as intended and minimal.

Issue #624, opened 2023-05-22
Feature request: Add source (reference) to CVSS


jaccoNCSCNL:
As a national CERT we often bundle multiple advisories into one. This also means that we sometimes have conflicting CVSS scores from different sources for the same vulnerability.
We think that it would make sense to add an optional source (or reference) to a CVSS score which would indicate who made this score.

tschmidtb51:
Thank you for the suggestion. The TC will discuss the suggestion.

sthagen:
During the 2023-09-27 meeting of the TC the members agreed to implement the feature in CSAF v2.1.

tschmidtb51:
@jaccoNCSCNL To follow the OASIS process, please also announced the request on the [comment mailing list](https://groups.oasis-open.org/communities/community-home?communitykey=3720cda2-3056-4183-a759-018f5aa7b5f2). A simple email pointing to the issue should be sufficient.

jaccoNCSCNL:
This has been announced at https://groups.oasis-open.org/discussion/feature-request-add-source-reference-to-cvss

PR #761, opened 2024-07-31
`scores` to `metrics` and add `source` as optional property

tschmidtb51:
- resolves https://github.com/oasis-tcs/csaf/issues/624
- rename `scores` to `metrics`
- add new level `content` to group scores (and metrics)
- add `source` as URI
- adapt prose to reflect schema
- adapt examples to reflect changed schema
- adapt test files to reflect current schema


csaf_2.1/json_schema/csaf_json_schema.json
```diff
@@ -1201,6 +1201,57 @@
...
+                  }
+                },
+                "products": {
+                  "$ref": "#/$defs/products_t"
+                },
+                "source": {
+                  "title": "Source",
+                  "description": "Contains the URL of the source that originally determined the metric.",
+                  "type": "string",
+                  "format": "uri"
```
mprpic, line 1250: Does it make sense to restrict this to URIs? The initial feature request notes "...add an optional source (or reference)", which makes me believe that setting this to an arbitrary text value is also desired.

mprpic (commented):
Could the motivation behind the change from `scores` to `metrics` be documented in this PR or the associated issue? I understand why it's being renamed but not everyone may be familiar with the context leading to this change.

csaf_2.1/json_schema/csaf_json_schema.json
sthagen: I am not using this currently, but URLs might be more meaningful for consumers than general text, no? Something more like an identifier that one could build hard relational anlaysis on, kind of.

mprpic: @jaccoNCSCNL, any preference from you since you originally requested this feature?

tschmidtb51:
> Could the motivation behind the change from `scores` to `metrics` be documented in this PR or the associated issue? I understand why it's being renamed but not everyone may be familiar with the context leading to this change.

Sure. I should have done that right away - there are mainly 2 reasons:

1. It aligns with other schemas in the space (e.g. CVE).
2. `Metrics` seems to be a more generic term and allows for later addition of other metrics that are not necessarily scores (e.g. SSVC).
3. The keyword change allows CSAF viewers to support CSAF 2.1 with small additions of code, without having to branch on the CSAF version used.

csaf_2.1/json_schema/csaf_json_schema.json
tschmidtb51: As our goal should be automation, I'm quite hesitant to add generic text... As a source does not fall out of the sky, I think a URL is a reasonable choice. (It would have also been a URL if we choose to implement it via `references`).

santosomar:
A motion has been started and seconded at: https://groups.oasis-open.org/discussion/motion-for-761

If no objection received before 2024-08-07 23:00 UTC this motion will automatically carry.

santosomar:
The motion to accept the pull request as suggested in https://github.com/oasis-tcs/csaf/pull/761 and include it into CSAF 2.1, has passed.  https://groups.oasis-open.org/discussion/motion-for-761

Issue #633, opened 2023-06-16
Make TLP mandatory


tschmidtb51:
A central point of the distribution model of CSAF is the TLP label. We should make it mandatory for all CSAF documents.

tschmidtb51:
Motion was set in https://groups.oasis-open.org/discussion/motion-to-add-that-the-tlp-label-is-mandatory-in-every-csaf-document-as-suggested-in-issue-633

tschmidtb51:
- [ ] Check whether it is save to remove `UNLABELED` from PMD schema
- [ ] Deal with Test 6.2.10

PR #775, opened 2024-08-26
TLP 2.0 in PMD

tschmidtb51:
- addresses parts of https://github.com/oasis-tcs/csaf/issues/633
- adapt provider-metadata.json to use only TLP 2.0
- add TLP also to directory-based distribution


sthagen (approved):
LGTM.

Issue #796, opened 2024-10-16
Clarify CPE version 


sonnyvanlingen:
As can be read in `3.1.3.3.1 Full Product Name Type - Product Identification Helper - CPE`, the text includes a reference to the CPE Naming Specification Version 2.3 for more details.

It is not documented explicitly that CPE version 2.3 is required, although one could deduce this from:

1) the regular expression provided
2) the reference to the CPE 2.3 Naming specification.

For usability reasons. I suggest to clarify in the text which version of CPE is used

tschmidtb51:
Even though we mention only CPE 2.3 as reference both, CPE 2.2 and CPE 2.3, are allowed.

sonnyvanlingen:
Right, indeed the regular expression seems to allow both.
My suggestion still stands, let's clarify textually :)

tschmidtb51:
I added a suggestion to resolve it with #800

sonnyvanlingen:
LGTM

PR #800, opened 2024-10-17
CPE

tschmidtb51:
- resolves oasis-tcs/csaf#796
- add note that CPE 2.2 and 2.3 is supported

@sthagen: Do you think that this is the appropriate form?


sthagen (approved):
LGTM that we at least state it (albeit informative only).

sthagen:
Given the fuzziness of CPE matching an inventory I am not sure we do need a normative support statement about the version range of our CPE support. If we need that to me the question is how one can test if we do in fact support all corner cases of these two upstream CPE versions. Maybe some practitioners can speak up? Don't we even have CPE authors on our TC as members?

tschmidtb51:
I don't think we need a mandatory statement - the informative should be sufficient.

Issue #625, opened 2023-05-23
Specify recursion depth for `branches`


tschmidtb51:
I remember that we discussed a recursion depth of 25 nested `branches` within a `/product_tree`. However, I didn't see it in the standard (or I overlooked it).
I think that might be a valuable addition to the guidance - especially for a future API.

tschmidtb51:
The TC should decide whether this should be a formal or informal requirement as well as whether it is a "SHOULD" or a "MUST" requirement.

tschmidtb51:
As we have currently 13 branch categories, I suggest 30 as limit and the following wording in the formal part:

## 5.4 Branch recursion

The `/product_tree` uses a nested structure for `branches`. Along a single path to a leaf, the recursion of `branches` is limited to 30 repetitions. Therefore, the longest path to a leaf is:

```
/product_tree/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/branches[]/product
```

PR #806, opened 2024-10-24
Branches recursion depth

tschmidtb51:
- fixes oasis-tcs/csaf#625
- add original recursion depth of 25 to Guidance of CSAF 2.0


sthagen (approved):
LGTM.

PR #788, opened 2024-09-25
Code block syntax

tschmidtb51:
- addresses parts of oasis-tcs/csaf#787
- correct syntax for `product_identification_helper`
- extracted from #781 => thanks to @mprpic

Issue #787, opened 2024-09-25
Check code blocks for correct syntax


tschmidtb51:
@mprpic found in #781 some incorrect brackets. We need to check all code blocks for the correct syntax and inconsistency.

sthagen:
As discussed during editor meeting today, the brackets were correct before (schema not instance), right?
My current position on verifying schema code blocks (with the comment ellipsis hacks to provide focus) is to visually inspect every time there is a change and to compare with the JSON Schema artifact source.

>When verifying code blocks we must take the intended perspective, as there is JSON instance code and JSON Schema code.

tschmidtb51:
@sthagen We should discuss (and document) in our next meeting how we want to deal with indent and commas.

PR #809, opened 2024-10-25
Revert 788 "Code block syntax"

tschmidtb51:
- addresses parts of #787 
- reverts #788 as that incorrectly changed '{` (used in the JSON schema) to `[` (used in the JSON document)

The code blocks show excerpts of the [***schema***](https://github.com/oasis-tcs/csaf/blob/master/csaf_2.1/json_schema/csaf_json_schema.json), not a single document.


sthagen (approved):
LGTM. Switching between JSON Schema "things" and instances of these "things" confuses me every second day ...

PR #810, opened 2024-10-25
Date-time

tschmidtb51:
- addresses parts of oasis-tcs/csaf#469
- add section on Date and Time to explain the rules

### DIFF
--- a/csaf_2.1/prose/edit/src/conformance.md
+++ b/csaf_2.1/prose/edit/src/conformance.md
@@ -60,6 +60,7 @@ The entities ("conformance targets") for which this document defines requirement
 
 A text file or data stream satisfies the "CSAF document" conformance profile if it:
 
+* satisfies to the syntax and semantics defined in section [sec](#date-time)
 * conforms to the syntax and semantics defined in section [sec](#schema-elements).
 * satisfies at least one profile defined in section [sec](#profiles).
 * does not fail any mandatory test defined in section [sec](#mandatory-tests).
--- a/csaf_2.1/prose/edit/src/conformance.md
+++ b/csaf_2.1/prose/edit/src/conformance.md
@@ -60,7 +60,7 @@ The entities ("conformance targets") for which this document defines requirement
 
 A text file or data stream satisfies the "CSAF document" conformance profile if it:
 
-* satisfies the syntax and semantics defined in section [sec](#date-time)
+* conforms to the syntax and semantics defined in section [sec](#date-time)
 * conforms to the syntax and semantics defined in section [sec](#schema-elements).
 * satisfies at least one profile defined in section [sec](#profiles).
 * does not fail any mandatory test defined in section [sec](#mandatory-tests).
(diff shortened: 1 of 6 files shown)


tschmidtb51:
@sthagen I have a couple questions:
1. The references for ISO 8601 and RFC 3339 are currently in the informative reference. Should they change to normative?
2. Should we add test for that?
3. Do I need to add that file somewhere or does it get picked up automatically?

sthagen:
Now that we want to enforce this, I assume we should move both to normative.
Test cases as regression tests, right? Use and abuse tests that ensure changes to the schema do not change the semantics. Correct? I would like that.

sthagen (approved):
LGTM.

tschmidtb51:
I'll add the test later (or in a different PR).

csaf_2.1/prose/edit/src/conformance.md
```diff
@@ -60,6 +60,7 @@ The entities ("conformance targets") for which this document defines requirement
 
 A text file or data stream satisfies the "CSAF document" conformance profile if it:
 
+* satisfies to the syntax and semantics defined in section [sec](#date-time)
```
sthagen, line 63: Nit: either `satisfies the` or `conforms to the` - and maybe not vary. Why not conforms to all over the place (instead of adding satisfaction)?

sthagen (approved):
SLGTM.

csaf_2.1/prose/edit/src/conformance.md
tschmidtb51: I just copied from ~above~ below...

tschmidtb51: I guess we used the `satisfies` as we were talking about requirements (i.e. for the "tool" conformance targets) so that when we added the CSAF document target it felt still okay-ish

csaf_2.1/prose/edit/src/conformance.md
```diff
@@ -60,7 +60,7 @@ The entities ("conformance targets") for which this document defines requirement
 
 A text file or data stream satisfies the "CSAF document" conformance profile if it:
 
-* satisfies to the syntax and semantics defined in section [sec](#date-time)
+* satisfies the syntax and semantics defined in section [sec](#date-time)
```
tschmidtb51, line 63: @sthagen: My bad - you are right, that should be:
```suggestion
* conforms to the syntax and semantics defined in section [sec](#date-time)

tschmidtb51 (commented):
@sthagen: My bad - you were right.

csaf_2.1/prose/edit/src/conformance.md
tschmidtb51: My bad - you are right.

santosomar:
A motion was moved by Omar to include the changes suggested in this pull request, during the CSAF TC monthly meeting on 2024-10-30. The motion was seconded by Denny. The motion passed.

Issue #824, opened 2024-11-06
Add `$schema` to `testcases_json_schema.json`


tschmidtb51:
Currently, we are missing the `$schema` in `testcases_json_schema.json` which results in errors as we specify it in `testcases.json`.

milux:
Isn't it supposed to be exactly the other way around?
Isn't the schema violated by adding `$schema` to your JSON payload, as this doesn't seem to be specified?

tschmidtb51:
I agree that my wording was bad: True is that `$schema` is not defined in `testcases_json_schema.json` and therefore its usage in `testcases.json` results in an error.

Issue #616, opened 2023-03-12
Add a schema identifier to CSAF v2.1 and later data files 


sthagen:
## Proposal

> Add a schema identifier to CSAF v2.1 (and later) data files with a `MAY` (to minimize backward incompatibility for strict CSAF v2.0 files which are not allowed to carry extra keys)

Ideally this should be simply something like [SARIF does](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317481) with a `$schema` key and a value of type URL. Example of such a SARIF file: 

```json
{
  "version": "2.1.0",
  "$schema": "https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/schemas/sarif-schema-2.1.0.json",
...
```

The canonical value will be the matching eternal schema URL hosted at docs.oasis-open.org 

## Rationale

Currently the consumer of CSAF files has to know what schema they relate to.

santosomar:
Thomas Schmidt proposed a motion, as detailed in this [OASIS mailing list archive](https://lists.oasis-open.org/archives/csaf/202310/msg00019.html), to incorporate a schema identifier into CSAF v2.1 and subsequent data files, in line with the suggestion made in this GitHub issue. Stefan Haggen seconded the motion. There were no discussions or objections raised, and consequently, the motion was automatically passed on November 1, 2023, at 20:00 UTC.

tschmidtb51:
There was some discussion in https://github.com/oasis-tcs/csaf/pull/828#discussion_r1847554756 whether that would be allowed. For documentation purposes: @milux collected information and came to the result that it would be okay to use this method, see https://github.com/oasis-tcs/csaf/pull/828#issuecomment-2501283070 for details.

PR #828, opened 2024-11-18
Tests

tschmidtb51:
- fixes https://github.com/oasis-tcs/csaf/issues/824
- add missing `$schema` to testdata schema for CSAF 2.1
- disallow additional properties for CSAF 2.1
- add missing `$schema` to testdata schema for CSAF 2.0
- disallow additional properties for CSAF 2.0

### DIFF
--- a/csaf_2.0/test/validator/testcases_json_schema.json
+++ b/csaf_2.0/test/validator/testcases_json_schema.json
@@ -79,10 +79,20 @@
     }
   },
   "required": [
+    "$schema",
     "tests",
     "testschema_version"
   ],
   "properties": {
+    "$schema": {
+      "title": "JSON schema",
+      "description": "Contains the URL of the JSON schema for test data which the document promises to be valid for.",
+      "type": "string",
+      "enum": [
+        "https://raw.githubusercontent.com/oasis-tcs/csaf/master/csaf_2.0/test/validator/testcases_json_schema.json"
+      ],
+      "format": "uri"
+    },
     "tests": {
       "title": "List of tests",
       "description": "Contains a list of test data.",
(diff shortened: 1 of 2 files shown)


csaf_2.0/test/validator/testcases_json_schema.json
```diff
@@ -79,10 +79,20 @@
     }
   },
   "required": [
+    "$schema",
```
oxisto, line 82: In general this is ok for me, since this is also a test file, one just needs to be aware of that specifying the `$schema` as a property is not really "recommended" or specified by the standard. 

Although it is used by many editors (see VSCode https://code.visualstudio.com/docs/languages/json#_mapping-in-the-json) there is no official statement from the specification on that (see https://github.com/orgs/json-schema-org/discussions/473). In the fact the discussion there did not end with any conclusion at all.

I will conduct a small test, whether this actually works at least with the 2 JSON schema generator libraries that we use

tschmidtb51: @oxisto Could you please double check that the JSON schema spec does not prohibit the use of `$schema` in a JSON instance and as defined in the JSON schema. (This might have an impact on #616.)

At the moment, I don't see another way to make sure the location of the schema is known other than to declare which schema should be used to validate ...

sthagen (approved):
LGTM.

csaf_2.0/test/validator/testcases_json_schema.json
tschmidtb51: Done in https://github.com/oasis-tcs/csaf/pull/828#issuecomment-2501283070

Issue #705, opened 2024-02-27
Add Sharing Groups


boschpsirt:
At the document level there is a Publisher sub-item. There are some use cases where a publisher does not write an advisory for all customers, but rather publishes it for each one with appropriate additional information. In human-readable format, this is in the title and in the watermark.

Would it be possible to expand this for this or the next release? I could imagine placing Publisher and Document References directly between the points in the Secvisogram and adapting them accordingly to CSAF.

tschmidtb51:
I could imagine to extend the `distribution` by an item called `sharing_group`. A required `id` would be a `uuid` and there could be an optional human-readable `name`. This should allow for such situations.

Note: CSAF documents that differ in content **MUST** use different `/document/tracking/id`. Given your suggestion that can easily be done by appending the `/document/distribution/sharing_group/id` to the `/document/tracking/id`.

tschmidtb51:
@boschpsirt Please also formally notify about the suggestion by writing to the csaf-comment mailing list (as soon as its back online).

boschpsirt:
Hi,
we could imagine to extend the distribution by an item called **sharing_group**. A required id would be a uuid and there could be an optional human-readable name. This should allow for such situations.

Note: CSAF documents that differ in content MUST use different /document/tracking/id. Given your suggestion that can easily be done by appending the /document/distribution/sharing_group/id to the /document/tracking/id.

boschpsirt:
> I could imagine to extend the `distribution` by an item called `sharing_group`. A required `id` would be a `uuid` and there could be an optional human-readable `name`. This should allow for such situations.
> 
> Note: CSAF documents that differ in content **MUST** use different `/document/tracking/id`. Given your suggestion that can easily be done by appending the `/document/distribution/sharing_group/id` to the `/document/tracking/id`.

Hi Thomas,

we agree with this proposal.

tschmidtb51:
Another use case for the suggested field is access restricted advisories. The addition of this field will allow automatic decision which CSAF documents show be displayed/accessible by which group based on the `sharing_group/id`. Therefore, it can also be used during the CVD process to provide data or an advisory stub automatically.

santosomar:
Call to action sent by @tschmidtb51 
Requests that all TC members review the suggested object `sharing_groups` within the next two weeks (2024-05-08) and comment in the issue directly.

https://groups.oasis-open.org/discussion/call-to-action-for-705

[...]

PR #839, opened 2024-12-02
Editor revision for TC meeting 2024-11-27

### LINKED ISSUE #768: Clarify quotes in changes.csv
Currently, it is not clear whether quotes are mandatory in the changes.csv or not. We should have a clear statement about that.

Currently, I tend to say: "All fields SHOULD be quoted." However, we need to check whether that is allowed by the csv-RFC.

### LINKED ISSUE #790: Clarify the inclusion of open-source projects for the value vendor in 3.2.1.8.1 Document Property - Publisher - Category
Section `3.2.1.8.1 Document Property - Publisher - Category` documents that:

> the value vendor indicates developers or maintainers of information system products or services. This includes all authoritative product vendors, Product Security Incident Response Teams (PSIRTs), and product resellers and distributors, including authoritative vendor partners.

For usability reasons, I suggest that in this definition we clarify that this definition also includes open source software providers, also aligning better with the definition in `1.2 Terminology`.

### LINKED ISSUE #797: Clarify requirement 19: ASCII vs. Binary signature
We should clarify whether we expect signatures to be in ASCII or binary format or both should be accepted. Therefore, we need to figure out whether it make a difference for implementers.

### LINKED ISSUE #825: Clarification on why test case 6-1-31-12 in CSAF2.0 is supposed to be valid
I am struggling to understand why https://github.com/oasis-tcs/csaf/blob/master/csaf_2.0/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_0-2021-6-1-31-12.json is supposed to be valid. It contains the product version string `after-eight`, which is (in my opinion) a clear violation of the test as its written, since the test specifies that the following strings should not be contained in the name:
```
<
<=
>
>=
after
all
before
earlier
later
prior
versions
```

The exact wording is "To implement this test it is deemed sufficient that, when converted to lower case, the value of name **does not contain any of the following strings**". `after-eight` clearly contains the string `after`.

I suspect that what is meant that they should not be present as **individual tokens separated by whitespace**, e.g. "after-eight" is valid and "after eight" is not. However, that is not what the test specifies and it is also treacherous ground because words as well as symbols are included in the "exclude list", so if one chooses to tokenize according to words, e.g. with whitespaces "> 4.2" would be invalid, but ">4.2" would not be. And even then if you consider "word boundaries" in terms of Regex, "after-eight" would still be considered two words and not one.

tschmidtb51:
- fixes https://github.com/oasis-tcs/csaf/issues/705
- fixes #825 
- fixes #797 
- fixes #768
- addresses parts of #469
- fixes #790


tschmidtb51:
Note: Some commits appear "doubled" due to a rebase...

sthagen (approved):
LGTM.

Note: The structural changes to align with the latest fashion will be in a dedicated change set / peer review as to not mix material with presentational changes.

santosomar (approved):

tschmidtb51:
@santosomar / @sthagen: I'll update the testfiles to sync them with the new presentational changes before the merge (after the event).

tschmidtb51:
> @santosomar / @sthagen: I'll update the testfiles to sync them with the new presentational changes before the merge (after the event).

Done in #846

tschmidtb51:
@santosomar There was a motion out for this that passed, right?

tschmidtb51:
> @santosomar There was a motion out for this that passed, right?

Yes: https://groups.oasis-open.org/discussion/motion-for-839

Issue #831, opened 2024-11-20
Clarify csaf.data.security.domain.ltd in Requirement 10: DNS path


sonnyvanlingen:
Suggestion to improve the usability. 

Section 7.1.10 Requirement 10: DNS path documents the following:

```The DNS record csaf.data.security.domain.tld SHALL resolve as a web server which serves directly the provider-metadata.json according to requirement 7.```

It is not clarified to the reader that ```domain.tld``` is simply a placeholder which has to be made specific to the issuing party's own organisation and context. 

Options to resolve:

1. Add a sentence clarifying the situation such as ```Replace 'domain.tld' with the appropriate domain specific to your organization of use case.```

2. Simply change ```domain.tld``` to ```example.com```

PR #868, opened 2025-02-19
DNS Domain

tschmidtb51:
- resolves oasis-tcs/csaf#831
- clarify text
- add informative explanation
Flagging @sonnyvanlingen

### DIFF
--- a/csaf_2.1/prose/edit/src/distributing.md
+++ b/csaf_2.1/prose/edit/src/distributing.md
@@ -163,7 +163,7 @@ The use of the scheme "HTTPS" is required. See [cite](#RFC8615) for more details
 ### Requirement 10: DNS path
 
 Assuming that the organization's main domain is `domain.tld`, the DNS record `csaf.data.security.domain.tld` SHALL resolve
-as a web server which serves directly the `provider-metadata.json` according to requirement 7.
+to the IP address of a web server which serves directly the `provider-metadata.json` according to requirement 7.
 
 > The `domain.tld` is just a placeholder for the organization's main domain.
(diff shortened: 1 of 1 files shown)


csaf_2.1/prose/edit/src/distributing.md
```diff
@@ -162,8 +162,13 @@ The use of the scheme "HTTPS" is required. See [cite](#RFC8615) for more details
 
 ### Requirement 10: DNS path
 
-The DNS record `csaf.data.security.domain.tld` SHALL resolve as a web server which serves directly
-the `provider-metadata.json` according to requirement 7. That implies that redirects SHALL NOT be used.
+Assuming that the organization's main domain is `domain.tld`, the DNS record `csaf.data.security.domain.tld` SHALL resolve
+as a web server which serves directly the `provider-metadata.json` according to requirement 7.
```
sthagen, line 166: Nit: Maybe more specifically instead of:

> as a web server which serves directly the

write something like:

> to the IP address of a web server directly serving the

(rationale: DNS resolves to IP addresses not to "web servers")

sthagen (approved):
LGTM (nit only, and that because is specifying details in a SHALL level requirement)

csaf_2.1/prose/edit/src/distributing.md
tschmidtb51: Done in [7a9fae0](https://github.com/oasis-tcs/csaf/pull/868/commits/7a9fae02692ae1c040083ee8c98e7ab3ec36b999)

PR #872, opened 2025-02-21
Model and Serial numbers

### LINKED ISSUE #847: Correct default open `serial_numbers` and `model_numbers`
During the CSAF workshops, it was brought up that the current wording regarding [Model numbers](https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#31333-full-product-name-type---product-identification-helper---model-numbers) and [Serial numbers](https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#31336-full-product-name-type---product-identification-helper---serial-numbers) is unfavorably:

> If a part of a [model/serial] number of the component to identify is given, it SHOULD begin with the first character of the [model/serial] number and stop at any point. Characters which SHOULD NOT be matched MUST be replaced by either ? (for a single character) or * (for zero or more characters).
Two * MUST NOT follow each other.

However, this is bad if the format is changed later on, e.g. by appending a character. For example, there is not way to differentiate between `12345` and `123456`. Both would match, if the value `12345` was given.

Therefore, we should change the text to not allow stopping at any point without appending the `*`.

tschmidtb51:
- fixes #847 
- correct default open behavior of model and serial number
- introduce escaping
- add conversion rules
- tighten definition: prohibit multiple stars
- add comment that backslashes need to be escaped themselves in JSON strings
- add mandatory test 6.1.43 for stars in model numbers
- add mandatory test 6.1.44 for stars in serial numbers
- add invalid examples
- add valid examples


sthagen:
@tschmidtb51 please take a look at  https://github.com/oasis-tcs/csaf/pull/872/commits/8772a187a4a9b7f5f60367fbc59c05871c3d27b3

Note: I was not able to create a branch plus pull request for that, which is OK, but makes me point out this attempt at guessing what we "want" and how "that" should be written here per this comment for clarity.

csaf_2.1/prose/edit/src/conformance.md
```diff
@@ -538,8 +538,38 @@ Secondly, the program fulfills the following for all items of:
 
 * type `/$defs/full_product_name_t/product_identification_helper/cpe`: If a CPE is invalid, the CSAF 2.0 to CSAF 2.1 converter SHOULD removed the
   invalid value and output a warning that an invalid CPE was detected and removed. Such a warning MUST include the invalid CPE.
+* type `/$defs/full_product_name_t/model_number`:
+  * If a model number is given that does not end on a star, the CSAF 2.0 to CSAF 2.1 converter SHOULD add a `*` to the end and output a
+    warning that a partial model number was detected and a star has been added.
+    Such a warning MUST include the model number.
+  * If the model number contains a `\`, the CSAF 2.0 to CSAF 2.1 converter MUST escape it by inserting an additional `\` before the character.
```
sthagen, line 545: Question also applies to serial identifier (a.k.a. number):
> Do we have a problem with UREL-encoding here if producers look up and insert such identifiers per web based systems that URL encode their payload? E.g. is "%5C%5c%5C*"
matching identifiers like "\\\*" (that is only this 4 character text) or like "%5C%5c%5-this-wun" and "%5C%5c%5-that-wun-too" and even "%5C%5c%5"?

I do not expect such corner cases to be significant or even advisable to use, but still maybe we should say something about this to avoid finger pointing between püarticipants of the eco system in the future?

tschmidtb51:
@sthagen Thank you for your feedback.

> - The text was very similar across model and serial number 
   matching and also exposed (in my reading) "conformance" 
   bleeding (e.g. that SHOULD NO be matched is not necessarily a 
   concern of the spec).

Thank you. I do still have a concern with one suggested phrase but we can work on that.

> - Fixed also some failed CPSR-coding where serial number term 
   occurred in the model number description of rules.

Thank you.

> - Added an example, where the first part of the model number
  "Pattern" is replaced with an asterisk wildcard

I'll add a few example for the serial number as well.

> - Some semantic clarification of phrasing attempted

Thanks.

tschmidtb51:
> Question still remains: Why not allow e.g. *-2024-*? 
I understand from our anchoring requirements (pattern describes 
the full string from the set of matching model/serial identifiers)
that we would want to allow shell glob like matching on such texts
from inventory management systems.
>
> Use case in point: In the past model numbers had the year
identifier in the middle. Now in the present there is a problem
with all components from models of the year 2024.
Problem: We force the producer to list all such identifiers 
explicitly because we prohibit patterns like *-2024-* 

I thought about that and decide that the advantages do not outweigh the disadvantages. Here are my reasons:
- Matching something like `*-2024-*` is tricky as this could be anywhere in the string. We want PIH to be unique enough to be a good fit for direct identification of products without a second round of matching - and without a high number of false positives.
- Serial and model numbers are usually finite. So the number of characters they consist of is finite. This also implies, that the lower number of characters could be replaced with `?` and the star is used for the other group, e.g. `???????-2024-*` or `*-2024-???????`.
- Serial and model numbers usually have a fixed format. That implies, a case, where `*-2024-*` is really true, seems seldom. IMHO, it is more likely to have the part of interest at a dedicated point. Such point than can be explicitly stated by using the `?` as position indicator.

csaf_2.1/prose/edit/src/conformance.md
tschmidtb51: I'm not 100% sure whether I understood the case. If someone writes `%5C%5c%5C*` as a serial number
- serial numbers that match would be, for example:
```
%5C%5c%5C
%5C%5c%5C*
%5C%5c%5C1
%5C%5c%5CAVeryAnnoying Long Serial Number - With even Space and 10 whitespaces 
```

- serial numbers that don't match would be, for example:
```
%5C%5c%5
%5C%5c%5-this-wun
*1234
\*4711 
```

What am I missing?

csaf_2.1/prose/edit/src/schema-elements-01-defs-03-full-product-name.md
```diff
@@ -325,9 +331,23 @@ Any given serial number of value type `string` with at least 1 character represe
 abbreviated (partial) serial number of the component to identify.
 
 If a part of a serial number of the component to identify is given,
-it SHOULD begin with the first character of the serial number and stop at any point.
-Characters which SHOULD NOT be matched MUST be replaced by either `?` (for a single character) or `*` (for zero or more characters).  
-Two `*` MUST NOT follow each other.
+it MUST begin at the first and end at the last character position of the string representing the targeted component.
+The wildcard characters `?` (for a single character) and `*` (for zero or more characters) signal exclusion of characters at these positions from matching.
```
tschmidtb51, line 335: I'm not sure, whether the sentence is quite clear. What we actually want is to tell that for a partial number you have to give the full structure but may replace irrelevant parts with the wildcard characters. What do you think about this?
```suggestion
To provide a part of a serial number of the component to identify, the full structure of the serial number MUST be given.
However, characters that are irrelevant to the matching MAY be replaced by the wildcard characters `?` (for a single character) and `*` (for zero or more characters) that signal exclusion of characters at these positions from matching.

File: csaf_2.1/prose/edit/src/schema-elements-01-defs-03-full-product-name.md [hunk 3]
```diff
@@ -236,13 +236,19 @@ the component to identify.
 > Often it is abbreviated as "MN", M/N" or "model no.".
 
 If a part of a model number of the component to identify is given,
-it SHOULD begin with the first character of the model number and stop at any point.
-Characters which SHOULD NOT be matched MUST be replaced by either `?` (for a single character) or `*` (for zero or more characters).  
-Two `*` MUST NOT follow each other.
+it MUST begin at the first and end at the last character position of the string representing the targeted component.
+The wildcard characters `?` (for a single character) and `*` (for zero or more characters) signal exclusion of characters at these positions from matching.
```
tschmidtb51, line 240: Same here.
```suggestion
To provide a part of a model number of the component to identify, the full structure of the model number MUST be given.
However, characters that are irrelevant to the matching MAY be replaced by the wildcard characters `?` (for a single character) and `*` (for zero or more characters) that signal exclusion of characters at these positions from matching.

tschmidtb51 (commented):
@sthagen Here is some new text to improve the partial stuff. Should we remove the term `abbreviated` from the initial field description?

csaf_2.1/prose/edit/src/conformance.md
sthagen: My corner case was that in browser editing environments (as secvisogram and others) the authors may be confused by copy-paste effects where the browser rewrites their data to make it conformant with RFCs and now verbatim backslashes intended as escaping meta characters end up becoming %5C or %5c and thus lose their escaping super power.

tschmidtb51: Ah. I would suggest to leave that to the implementers for now (as I'm out of good ideas how to phrase that). But if you have a suggested wording, please open a PR.

sthagen (approved):
SLGTM

sthagen:
> @sthagen Here is some new text to improve the partial stuff. Should we remove the term `abbreviated` from the initial field description?

let me think about that some more, please.

PR #885, opened 2025-03-04
Editor revision for TC meeting 2025-02-26

### LINKED ISSUE #782: The field name 'release_date' is confusing, consider changing the name to 'public_date'
While the documentation is clear, the field name 'release_date' is confusing. Could it be changed to 'public_date' or 'date_public'?

I'd suggest a minor documentation update also: "...the date and time the vulnerability was originally publicly disclosed" (or "published" or "made public").

https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html

3.2.3.11 Vulnerabilities Property - Release Date
Release date (release_date) with value type string of format date-time holds the date and time the vulnerability was originally released into the wild.

3.2.1.12.2 Document Property - Tracking - Current Release Date
Current release date (current_release_date) with value type string with format date-time holds the date when the current revision of this document was released.

3.2.1.12.5 Document Property - Tracking - Initial Release Date
Initial release date (initial_release_date) with value type string with format date-time holds the date when this document was first published.

### LINKED ISSUE #791: Clarify terminology of initial release of document version in 3.1.11.1 Version Type - Integer versioning
Section `3.1.11.1 Version Type - Integer versioning` documents that:

> Version 1 defines the initial **public** release. Each new version where /document/tracking/status is final has a version number incremented by one.

This may be misleading as it implies a release to the general public, while it could also refer to controlled disclosure labeled TLP:GREEN or higher.

My suggestion is to simply remove the word _public_ altogether changing the text to:

> Version 1 defines the initial release. Each new version where /document/tracking/status is final has a version number incremented by one.

### LINKED ISSUE #831: Clarify csaf.data.security.domain.ltd in Requirement 10: DNS path

### LINKED ISSUE #847: Correct default open `serial_numbers` and `model_numbers`

### LINKED ISSUE #870: CWE link with multiple targets
If you use the rendered version, the [ 6.1.11 CWE link](https://github.com/oasis-tcs/csaf/blob/master/csaf_2.1/prose/share/csaf-v2.1-draft.md#cwe) jumps to the reference instead of the text.

@sthagen Can we detect that automatically (to prevent future mistakes)? Can you please fix that?

tschmidtb51:
- addresses parts of #803 
- fixes #831
- fixes #870 
- fixes #847 
- fixes #791
- fixes #782 
- addresses parts of #469 
- add invalid examples
- add valid examples


csaf_2.1/prose/edit/src/conformance.md
```diff
@@ -604,7 +605,7 @@ Secondly, the program fulfills the following for all items of:
   > This is done to create a deterministic conversion.
 
   The tool SHOULD implement an option to use the latest available CWE version at the time of the conversion that still matches.
-
+* `/vulnerabilities[]/disclosure_date`: If a `release_date` was given, the CSAF 2.0 to CSAF 2.1 converter MUST convert the key as `disclosure_date`.
```
tschmidtb51, line 609: @sthagen: What do you think of the following wording?
```suggestion
* `/vulnerabilities[]/disclosure_date`: If a `release_date` was given, the CSAF 2.0 to CSAF 2.1 converter MUST convert its value as value into the `disclosure_date` element.

sthagen (approved):
LGTM - Ship it!

tschmidtb51:
A motion has been set by Thomas Schmidt: https://groups.oasis-open.org/discussion/motion-for-885

tschmidtb51:
> A motion has been set by Thomas Schmidt: https://groups.oasis-open.org/discussion/motion-for-885

 Stefan and Omar seconded. No discussion, no objections, the motion carries.

PR #896, opened 2025-03-12
Fix a few document/lang mentions

bernhardreiter:
* change `document/language` to `document/lang`, because only the latter is defined in the schema.
 * similar change `language_t` to `lang_t`.

### DIFF
--- a/csaf_2.1/prose/edit/src/conformance.md
+++ b/csaf_2.1/prose/edit/src/conformance.md
@@ -292,7 +292,7 @@ A CSAF content management system satisfies the "CSAF content management system" 
 
   * `/$schema` with the value prescribed by the schema
   * `/document/csaf_version` with the value prescribed by the schema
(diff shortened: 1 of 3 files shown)


bernhardreiter:
Raised at csaf-comment in https://groups.oasis-open.org/discussion/documentlang-mentions-fix .

(A pull-request seems to be the way to submit this suggestion that causes less work on the editors' side.)

tschmidtb51:
> Raised at csaf-comment in https://groups.oasis-open.org/discussion/documentlang-mentions-fix .
> 
> (A pull-request seems to be the way to submit this suggestion that causes less work on the editors' side.)

Thank you for your suggestion and spotting this mistake @bernhardreiter. Due to the OASIS rules, I can't accept Pull Requests from Non-TC members and therefore have to close this. It will be tracked as separate issue #897

Issue #897, opened 2025-03-12
Fix a few document/lang mentions


tschmidtb51:
As @bernhardreiter reported in https://groups.oasis-open.org/discussion/documentlang-mentions-fix, there are a few places, where we made a mistake in the naming:

- change `document/language` to `document/lang`, because only the latter is defined in the schema.
- similar change `language_t` to `lang_t`.

tschmidtb51:
@sthagen What do we do with this for CSAF 2.0? Should we keep a list of known issues somewhere?

sthagen:
> [@sthagen](https://github.com/sthagen) What do we do with this for CSAF 2.0? Should we keep a list of known issues somewhere?

Yes I think we should keep such a list and NOT push another errata of the 2.0 version to focus our efforts on finishing and providing version 2.1.

This way, we can help implementers and operators within v2.0 infrastructures but do not slow down the much requested v2.1 publication / progression.

PR #898, opened 2025-03-12
Language

tschmidtb51:
- addresses parts of oasis-tcs/csaf#897
- correct mistake where `language` was used instead of `lang`

spotted by @bernhardreiter


sthagen (approved):
LGTM - thanks for fixing.

Issue #771, opened 2024-08-22
Add new profile: "Withdrawn"


tschmidtb51:
We should add a special profile that is used for withdrawn CSAF documents.

Reasoning: Sometimes, it is necessary to withdraw a document. Currently, those documents are mostly set into the document category `csaf_base` and the somewhere (e.g. in the title or a note) a comment is added that the document was withdrawn. However, that is not really automatable.

Here is a suggested definition:

## Profile W: Withdrawn

This profile MUST be used for any CSAF document that is withdrawn. It MUST NOT be used for any superseded document.

A CSAF document SHALL fulfill the following requirements to satisfy the profile "Withdrawn":

* The following elements MUST exist and be valid:
  * all elements required by the profile "CSAF Base".
  * `/document[]/notes` with exactly one item using the `category` `description` and the `title` `Reasoning for Withdrawal` describing the original content and the reasons for the withdrawal
  * `/document/tracking/revision_history` with at least 2 entries. Any previous items MUST NOT be removed.
* The value of `/document/category` SHALL be `csaf_withdrawn`.
* The elements `/product_tree` and `/vulnerabilities` SHALL NOT exist.

The CSAF document MAY link to additional information through `/document/references`.

tschmidtb51:
Checklist:
- [ ] Add profile
- [ ] Adapt schema
- [ ] Adapt prose
  - [ ] Document category
  - [ ] Guidance on Size
  - [ ] Conversion rule 
- [ ] Add tests
  - [ ] 6.1.27.X
  - [ ] 6.1.27.3
  - [ ] Notes test
  - [ ] Revision History test 
- [ ] Maybe: add special modifier as conformance target?

santosomar:
Motion proposed [via this email to the TC](https://groups.oasis-open.org/discussion/motion-to-create-a-new-withdrawn-profile-in-csaf-21-issue-771)

Issue #672, opened 2023-12-15
Enforce use of affected in `csaf_security_advisory` 


tschmidtb51:
Currently, we only enforce that a `csaf_security_advisory` has at least one `product_status`. That allows the production of CSAF security advisories with all products in status `fixed`. One might implicit assume that previous versions are affected. However, that is terrible for automated matching:
If the advisory tells that Product A in version 4.2 is fixed and I have Product A but version 4.1 - the matcher can't tell me that I have an issue. Also, fixed products can't have remediations as they are fixed.

The easiest transition rule (CSAF 2.0=> CSAF 2.1) would be:
- Downgrade to `csaf_base` or
- add a version range below the fixed product (assuming that those are affected - less preferred as we are adding information that is not there and might be wrong).

mprpic:
I am not a big fan of this being enforced. It's easy to perhaps add a catch-all that anything _before_ the fixed version is affected, but then you might also be implicitly including versions that might not be vulnerable because they were released before the vulnerability was introduced into the product. Figuring out the exact ranges of affected versions costs a lot of time in analysis. In a perfect world, we'd have this data already but that's just not the case, especially for older CVEs. It would be a considerable amount of effort to figure out the exact affected versions for CVEs from 2010 and earlier, and I'm not sure it's worth it at this point. Publishing the information on at least fixed versions is still valuable and I would strongly dislike to see us having to remove that information or hobble it somehow (by downgrading the files to a lesser state) just because of this requirement.

tschmidtb51:
@mprpic Thank you for your input. I see this question between the poles of keeping old data valid and encourage to produce better data for tomorrow.

Here is my suggestion:

----

## Profile X: Security Advisory

This profile SHOULD be used to provide information which is related to vulnerabilities, affected products and corresponding remediations.

A CSAF document SHALL fulfill the following requirements to satisfy the profile "Security Advisory":

* The following elements MUST exist and be valid:
  * all elements required by the profile "CSAF Base".
  * `/product_tree` which lists all products referenced later on in the CSAF document regardless of their state.
  * `/vulnerabilities` which lists all vulnerabilities.
  * `/vulnerabilities[]/notes`
    > Provides details about the vulnerability.
  * `/vulnerabilities[]/product_status`
    > Lists each product's status in regard to the vulnerability.
  * `/vulnerabilities[]/product_status/affected`
    > Lists affected products in regard to the vulnerability.
[...]

PR #901, opened 2025-03-14
Profile: Withdrawn

tschmidtb51:
- addresses parts of https://github.com/oasis-tcs/csaf/issues/771
- add profile "Withdrawn"
- add conversion rules
- adapt test 6.1.26
- adapt test 6.1.27.3
- add test 6.1.27.14 for document notes
- add test 6.1.27.15 for product_tree
- add test 6.1.27.16 for revision history
- add test 6.1.27.17 for reasoning for withdrawal
- add test 6.2.37.2 for language specific translation
- add invalid examples
- add valid examples
- add language specific translations of terms used in the standard
- add conformance target CSAF withdrawer

- addresses parts of https://github.com/oasis-tcs/csaf/issues/672
- add missing failing example to prose


csaf_2.1/language_specific_translation/translations.json
sthagen, line 1: Let us think about the place (why not inside test)?

tschmidtb51: We want to use that file/structure also for the defined `title`/`category` combination, e.g. in https://github.com/oasis-tcs/csaf/blob/master/csaf_2.1/prose/edit/src/schema-elements-02-props-02-document.md#document-property---notes or https://github.com/oasis-tcs/csaf/blob/master/csaf_2.1/prose/edit/src/schema-elements-02-props-04-vulnerabilities.md#vulnerabilities-property---notes

csaf_2.1/language_specific_translation/translations.json
```diff
@@ -0,0 +1,12 @@
...
+  "$schema": "https://raw.githubusercontent.com/oasis-tcs/csaf/master/csaf_2.1/test/language_specific_translation/translations_json_schema.json",
+  "translation_version": "2.1",
+  "translations": [
+    {
+      "lang": "de",
+      "term": {
+        "reasoning_for_withdrawal": "Begründung für die Zurückziehung"
+      }
+    }
+  ]
```
tschmidtb51, line 11: We can also do:
```
  "translation": 
    {
      "de": {
        "reasoning_for_withdrawal": "Begründung für die Zurückziehung"
      }
    }
```

However, I guess we need either specify all accepted languages (adapt the schema each time - which makes it harder for all that use the JSON as they might need to update their program if we add keys...) or use `patternProperties`. The latter is definitely less work on the schema, but not sure were that leaves us with the implementations :thinking:

tschmidtb51:
@sthagen - sorry, forgot to mark it as ready for review.

csaf_2.1/language_specific_translation/translations.json
sthagen: I still seem to not understand: We try to detect combinations of (title, category), right? But, we do so as a test. Should these artifacts supporting the detection not be in the test tree? Sorry, if I do not see the forest for the trees.

tschmidtb51: > I still seem to not understand: We try to detect combinations of (title, category), right?

Yes.

> But, we do so as a test.

In this specific case: yes. But we want to provide the official/recognized translations for all users of the standard - independent of whether they are checked by a test (as done here) or not (e.g. the document/vulnerability notes). As the latter are independent of the tests, I thought it would surprise users to look for the official/recognized translations in the test folder.

> Should these artifacts supporting the detection not be in the test tree?

In this specific case, we could move it to the test tree. But that would hide the other translations that do not have a test (and probably will never have one). Given the principle of least surprise, I would suggest that this should live alongside prose and schema.

tschmidtb51: We could also put it in the test tree from now and move it with the next PR up to the intended level. However, that would require implementers to touch the newly implemented test again (instead of just updating the file).

csaf_2.1/language_specific_translation/translations.json
sthagen: Could we not use a pattern in the JSON schema for these keys?

tschmidtb51: > Could we not use a pattern in the JSON schema for these keys?

Sorry - just to get this right: Do you want to use the `patternProperties` for the JSON keys?

sthagen: Yes, that was my thinking.

tschmidtb51: Done in [bcc250b](https://github.com/oasis-tcs/csaf/pull/901/commits/bcc250b4194d8ebee1fceeb411e1d94f7c5fe74c)

sthagen (approved):
LGETM - let us resolve the conflicts, accept, merge it, and move on.

tschmidtb51:
> LGETM - let us resolve the conflicts, accept, merge it, and move on.

Will do.

Issue #653, opened 2023-09-25
Mandate `Access-Control-Allow-Origin: *` for files available in public to allow browser based clients


bernhardreiter:
If a server is serving CSAF 2.0 related files to the public it MUST set the HTTP-Header `Access-Control-Allow-Origin: *` to allow web browser based clients.

This header setting allows "cross origin resource-sharing" (CORS), so that an application within a browser from a third party domain can access the CSAF information (like security.txt, provider-medata.json, ROLIE files and csaf documents itself).

Without this header for example a single page application cannot access the contents of the files of a CSAF Provider. See the following code example, which you can run in a web browser with javascript enabled (save it as `testcors.html` and open it with the browser).

```html
<!DOCTYPE html>
<html lang="en">
<body>
  <h1>JSON Data Fetched From URL</h1>
  <div id="data"></div>

  <script type="text/javascript">
    // Replace this URL with the actual URL of the JSON file you want to fetch
    var url = 'https://wid.cert-bund.de/.well-known/csaf/provider-metadata.json';
    //var url = 'https://jsonplaceholder.typicode.com/todos/1';

    async function getData() {
        try {
            const response = await fetch(url);
            // variant showing that we indeed get no response with `no-cors`
            //const response = await fetch(url, { mode: 'no-cors' });
            const data = await response.json();

            let ul = document.createElement("ul");
            for (let key in data) {
                let li = document.createElement('li');
                li.textContent = `${key}: ${data[key]}`;
                ul.appendChild(li);
            }
            document.getElementById('data').appendChild(ul);
        } catch (error) {
            console.log(`Error fetching data: ${error}`);
        }
    }

    getData();
  </script>
</body>
</html>
```

You will not see any contents. If you comment out the WID url and use the one from a testing json file, it can be seen that the contents is displayed. A look into the development tools and the network requests shows that a respective header is given.

References:
 * https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin
 * https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS

sthagen:
During the 2023-09-27 meeting of the TC the members considered the inclusion of the proposed addition to the CSAF v2.1 standard.
> If a server is **serving CSAF 2.0 related files to the public** it MUST set the HTTP-Header `Access-Control-Allow-Origin: *` to allow web browser based clients. 

[...]

PR #903, opened 2025-03-14
Add `Access-Control-Allow-Origin` requirement

tschmidtb51:
- addresses parts of https://github.com/oasis-tcs/csaf/issues/653
- add requirement 25 as SHOULD
- include it for CSAF provider
- include it for CSAF aggregators
- add security considerations

### DIFF
--- a/csaf_2.1/prose/edit/src/distributing.md
+++ b/csaf_2.1/prose/edit/src/distributing.md
@@ -597,6 +597,17 @@ The temporary blocking SHOULD be removed as soon as possible, at latest two week
 
 > Also confer to the TC's guidance on content delivery networks and caching.
 
+### Requirement 25: Access-Control-Allow-Origin
+
+For any CSAF documents and related metadata, the web server SHOULD set the HTTP header `Access-Control-Allow-Origin: *`.
+
+> The HTTP header enables users to access the CSAF data with web browser based clients.
+
+The HTTP header MAY be altered to allow just specified domains.
+In such case, the response SHOULD also include the `Vary` header.
+
+> Such restriction may allow the allow-listed domains to send credentials.
+
 ## Roles
 
 This subsection groups the requirements from the previous subsection into named sets which target the roles with the same name.
@@ -634,7 +645,7 @@ A CSAF publisher satisfies the "CSAF provider" role if the party fulfills the fo
 Firstly, the party:
 
 * satisfies the "CSAF publisher" role profile.
-* additionally satisfies the requirements 5 to 7 and 24 in section [sec](#requirements).
+* additionally satisfies the requirements 5 to 7, 24 and 25 in section [sec](#requirements).
 
 Secondly, the party:
 
@@ -658,7 +669,7 @@ A CSAF provider satisfies the "CSAF trusted provider" role if the party:
 
 A distributing party satisfies the "CSAF lister" role if the party:
 
-* satisfies the requirements 6, 21, 22 and 24 in section [sec](#requirements).
+* satisfies the requirements 6, 21, 22, 24 and 25 in section [sec](#requirements).
 * uses the value `lister` for `/aggregator/category`.
 * does not list any mirror pointing to a domain under its own control.
 
@@ -669,7 +680,7 @@ A distributing party satisfies the "CSAF lister" role if the party:
 
 A distributing party satisfies the "CSAF aggregator" role if the party:
 
-* satisfies the requirements 1 to 6 and 21 to 24 in section [sec](#requirements).
+* satisfies the requirements 1 to 6 and 21 to 25 in section [sec](#requirements).
 * uses the value `aggregator` for `/aggregator/category`.
 * lists a mirror for at least two disjoint issuing parties pointing to a domain under its own control.
 * links the public part of the OpenPGP key used to sign CSAF documents for each mirrored issuing party in
--- a/csaf_2.1/prose/edit/src/distributing.md
+++ b/csaf_2.1/prose/edit/src/distributing.md
@@ -604,7 +604,7 @@ For any CSAF documents and related metadata, the web server SHOULD set the HTTP 
 > The HTTP header enables users to access the CSAF data with web browser based clients.
 
 The value of the HTTP header MAY be altered to allow just specified domains.
-In such case, the response SHOULD follow the recommendation of [cite](#FETCH) including but not limited to about the `Vary` header.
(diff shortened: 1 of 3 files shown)


tschmidtb51:
@bernhardreiter Please review

csaf_2.1/prose/edit/src/distributing.md
```diff
@@ -597,6 +597,17 @@ The temporary blocking SHOULD be removed as soon as possible, at latest two week
...
 > Also confer to the TC's guidance on content delivery networks and caching.
 
+### Requirement 25: Access-Control-Allow-Origin
+
+For any CSAF documents and related metadata, the web server SHOULD set the HTTP header `Access-Control-Allow-Origin: *`.
+
+> The HTTP header enables users to access the CSAF data with web browser based clients.
+
+The HTTP header MAY be altered to allow just specified domains.
+In such case, the response SHOULD also include the `Vary` header.
```
bernhardreiter, line 607: The recommendation for the `Vary` header here is more complex. 
https://fetch.spec.whatwg.org/#cors-protocol-and-http-caches explains that setting `Vary: Origin` is only good, if there are two or more domains. 

So we could either:
 1. refer to that recommendation, like ``In such case follow the response SHOULD follow the recommendations of [FETCH 6] about setting the `Vary` header for caching.``

BTW: I wonder if caching really is relevant here as protected feeds will have to send credentials via TLS anyway. But in theory caching could happen before the TLS termination on the servers side.

2. Be more specific: ``In case of two or more domain, the response SHOULD also include a `Vary: Origin` header.``

csaf_2.1/prose/edit/src/safety-security-and-data-protection.md
```diff
@@ -31,4 +31,16 @@ Moreover, it SHALL be treated as unsafe (user) input.
   > Additional, supporting mitigation measures like retrieving only CSAF documents from trusted sources and check their integrity and
   > signature before parsing the document SHOULD be in place to reduce the risk further.
 
+The distribution requirements of CSAF data allow to specify domains as the value of the HTTP header `Access-Control-Allow-Origin`.
+While a wildcard (`*`) as header value usually prevents implementing browsers from sending credentials during the CORS request,
+the restriction to specified domains often enables sending credentials.
+As such configuration is more complex, there is a higher risk of misconfiguration.
```
bernhardreiter, line 37: Better: `Allowing several specific domains means using dynamics on the server, which can widen the attack surface slightly by more active code and configuration.`

csaf_2.1/prose/edit/src/safety-security-and-data-protection.md
```diff
@@ -31,4 +31,16 @@ Moreover, it SHALL be treated as unsafe (user) input.
...
+Furthermore, this might leak information about internal structures or tools used.
```
bernhardreiter, line 38: Better: `Specifying a fixed single domain will let others know which domain can send javascript code that is allowed to send credentials.`

csaf_2.1/prose/edit/src/distributing.md
tschmidtb51: Addressed with [fa234ce](https://github.com/oasis-tcs/csaf/pull/903/commits/fa234cebca8b99a0d6a22916937895774c48a44f)

csaf_2.1/prose/edit/src/safety-security-and-data-protection.md
tschmidtb51: Addressed with [62144c5](https://github.com/oasis-tcs/csaf/pull/903/commits/62144c5c728a2f1be4dc6d108cecfe22640f7e8f)

csaf_2.1/prose/edit/src/safety-security-and-data-protection.md
tschmidtb51: It is not only about "who can sen credentials. Those domains (e.g. webview.example.com, secvisogram.example.com) might also leak the tools used, e.g. based on the domain name.

addressed in [a78c83c](https://github.com/oasis-tcs/csaf/pull/903/commits/a78c83c2c936e6719f925a2f84fd2eabf545cffa)

csaf_2.1/prose/edit/src/distributing.md
bernhardreiter: LGTM

csaf_2.1/prose/edit/src/safety-security-and-data-protection.md
bernhardreiter: Good enough. :)

(That a server trusts Javascript code coming from a specific domain to send credential can be a hint that this tools used by the server admnistrators as well and if an internal domain is allowed then it also reveals a bit of internal structure.)

bernhardreiter (approved):
LGTM

csaf_2.1/prose/edit/src/safety-security-and-data-protection.md
```diff
@@ -31,4 +31,16 @@ Moreover, it SHALL be treated as unsafe (user) input.
   > Additional, supporting mitigation measures like retrieving only CSAF documents from trusted sources and check their integrity and
   > signature before parsing the document SHOULD be in place to reduce the risk further.
 
+The distribution requirements of CSAF data allow to specify domains as the value of the HTTP header `Access-Control-Allow-Origin`.
+While a wildcard (`*`) as header value usually prevents implementing browsers from sending credentials during the CORS request,
+the restriction to specified domains often enables sending credentials.
+Allowing several specified domains results in using dynamics on the server, which can widen the attack surface slightly by more code and configuration.
```
sthagen, line 37: Maybe the "slightly" quantifier could go, or even better we could use another sentence to weight the consequences in preparation of the following "therefore ..." conclusion?

tschmidtb51: Removed the "slightly" in [8c119fc](https://github.com/oasis-tcs/csaf/pull/903/commits/8c119fc6b975a99952a5cc4763aa9c492412e961).

tschmidtb51: Do you have a suggestion for the sentences?

sthagen: I would now just keep it as it is - finishing mode ...

sthagen (approved):
LGTM &rightarrow; let's ship it!

PR #923, opened 2025-03-28
Model and Serial Number

tschmidtb51:
- addresses parts of oasis-tcs/csaf#384
- remove phrase `partial or full` from model and serial numbers


sthagen (approved):
LGTM.

Issue #864, opened 2025-02-11
Editorial


tschmidtb51:
- [x] add `s` to path in https://github.com/oasis-tcs/csaf/blob/master/csaf_2.1/prose/edit/src/tests-03-informative.md#missing-cwe ( `/vulnerabilities[]/cwe` => `  /vulnerabilities[]/cwes`
- [x] check conformance clause 19 to see whether it needs to be extended to mention section 2 and 5
- [x] make `version` code syntax in 6.1.11 
- [x] fix format in Enum section of guidance on size

tschmidtb51:
- [ ] Check whether the term abbreviated should be removed from the initial field description of serial and model numbers

tschmidtb51:
[ ] Check whether we should clarify that UUIDs are in general unique (as suggested in https://github.com/oasis-tcs/csaf/issues/705#issuecomment-2710338914)

tschmidtb51:
- [ ] check SSVC v1.x.y => correct?

tschmidtb51:
- [ ] check path in tests; `flags` missing

tschmidtb51:
- [ ] 6.3 Informative Test**s**

tschmidtb51:
- [ ] correct `metrics` in conformance

Issue #863, opened 2025-02-10
Add product_ids and group_ids to note objects under vulnerability


chripluEri:
Support product specific notes in vulnerabilities by adding product_ids and group_ids to note objects under vulnerability.

**Problem Statement**
Currently, it is not possible to add a notes with additional details to a vulnerability where the note only refers to one of several products.

In a SW product with two container images (i1 and i2) and where both images contain a vulnerable FOSS component (f1) and where the vulnerability (v1) in f1 has a different impact on the product depending whether it is analyzed in image i1 or in image i2. CSAF support relationships and `product_ids` in the `product_status` object so I can use this to provide one status` known_affected` for image i1 and another status `known_not_affected` for image i2. I can use flags with `product_ids` to provide a justification why v1 in i2 does not affect the product. But how can I add additional information like a written note to explain why the same vulnerability does not affect the product in the context of i2?

The object `threat `has `product_ids/group_ids` to specify to which products it refers to. But all three child-object (`exploit_status, impact, target_set`) only refer to `known_affected` products, and not to `known_not_affected` products. For `known_not_affected` cases, I cannot see a suitable place to provide additional explanation beyond a justification flag.

**Proposal**
I suggest to add optional fields `product_ids` and `group_ids` under note type in vulnerability (not needed on the document level).

Like this:

```
"vulnerabilities": [
    {
      "notes": [
        {
          "category": "details",
          "text": "context specific to product",
          "title": "further details and explanation context specific to product"
          "product_ids": [
            "CSAFPID-0001"
          ],
          "group_ids": [
            ""
          ]
        }
      ]
    }
  ]
```
Christoph Plutte, TC member

tschmidtb51:
In the TC meeting on 2025-02-26, the TC agreed to extend the suggestion to notes in general.

rjb4standards:
Question regarding product ID's:

Do these product ID's refer to the id's in the CSAF product tree or a PURL product ID?

cplutte:
> Question regarding product ID's:
> 
> Do these product ID's refer to the id's in the CSAF product tree or a PURL product ID?

they refer to the product_ids/group_ids as defined in the product_tree similarly to the example given in JSON and similar to the ids in score-, threat-, flag- and product_status-elements.

tschmidtb51:
[...]

PR #925, opened 2025-03-31
Translation - Product description

tschmidtb51:
- addresses parts of oasis-tcs/csaf#863, oasis-tcs/csaf#864
- add translation for product description


tschmidtb51:
Note: CPE test fails due to unavailability of NIST NVD servers

Issue #905, opened 2025-03-14
Split optional tests


tschmidtb51:
The CSAF 2.1 source file that holds the optional tests is quite long. Time to split it.

PR #927, opened 2025-04-03
Editorial - Split optional tests

tschmidtb51:
- addresses parts of oasis-tcs/csaf#905
- split optional tests into single files


sthagen (approved):
LGTM.

Thanks for investing the effort in splitting, it will make our future content work on the spec in that area a lot easier.

Issue #652, opened 2023-08-30
Support CVSS 4.0 in CSAF 2.x


fjscao:
CVSS 4.0 official publication is scheduled for Q4, 2023. CSAF 2.x schema needs to add the new 
                "cvss_v4": {
                      "$ref": "https://www.first.org/cvss/cvss-v4.0.json"
                  },

santosomar:
I recently received an identical question from a member of our community, highlighting the importance of this issue. From a technical standpoint, it's entirely feasible to integrate the CVSS v4.0 standard into our current system. You can easily add the property CVSS v4 (cvss_v4) with a corresponding value that adheres to the CVSS v4.0 schema, as outlined at [FIRST's CVSS v4.0 JSON schema](https://www.first.org/cvss/cvss-v4.0.json). Here's how you can reference it in your JSON file:

```json
"cvss_v4": {
  "$ref": "https://www.first.org/cvss/cvss-v4.0.json"
},
```

However, there's a caveat. Implementing this would lead to a failure in the [mandatory tests as described in CSAF documentation](https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#618-invalid-cvss). This raises a critical point: The landscape of cybersecurity standards like CVSS is continuously evolving. Rigidity in CSAF's requirements could make it less adaptive and potentially obsolete as these standards undergo updates or revisions.

To future-proof CSAF and encourage broader community adoption, it may be prudent to adopt a more flexible approach that can easily accommodate changes in CVSS and other related or supported standards. This way, we can ensure that CSAF remains up-to-date in the rapidly evolving field of cybersecurity.

tschmidtb51:
A future-proof standard is one that is actively maintained - and we are doing this with CSAF.

The most important point to me is, that people can develop their tools to a fixed version of the standard and know what they can expect. Custom extensions and adding special fields makes it harder to validate files (you can't tell anymore whether it is a misspelling or a custom extension) and to write tools that process files as you have to implement all different extensions.

Therefore, I'm much more in favor of writing a new version of the standard to avoid the complications...

fjscao:
First, CVSSv3.0 was released in June 2015, CVSSv3.1 in June 2019.  The frequency for CVSS new versions isn't a concern, assuming CSAF also has new versons.

[...]

PR #929, opened 2025-04-05
Editor revision for TC meeting 2025-03-26

### LINKED ISSUE #653: Mandate `Access-Control-Allow-Origin: *` for files available in public to allow browser based clients

### LINKED ISSUE #672: Enforce use of affected in `csaf_security_advisory`

### LINKED ISSUE #771: Add new profile: "Withdrawn"

### LINKED ISSUE #780: Support for Multiple Notes, Products, and IDs
Note: Background information in #773 

As discussed in our TC meeting on 2024-08-28, there were two issues presented in #773 . This issue is to address the challenge that there's currently no mechanism to bind a specific `note` instance to an individual `product_id` or a set of `product_ids`.

### LINKED ISSUE #803: Include support for SSVC
We should include support for SSVC, as discussed in #462 and during July TC meeting.

### LINKED ISSUE #863: Add product_ids and group_ids to note objects under vulnerability

### LINKED ISSUE #864: Editorial

### LINKED ISSUE #876: Requirement 14: Prohibit JavaScript
In [Requirement 14](https://github.com/oasis-tcs/csaf/blob/master/csaf_2.1/prose/share/csaf-v2.1-draft.md#requirement-14-directory-listings), the directory listings are described to support manual navigation. We should add a sentence to prohibit the use of client-side scripts to generate the listing as the data is static and this is a problem for text-based browsers (and such entities that filter JavaScript at their firewall). Moreover, it can also not correctly be detected by a CSAF checker.

Suggested text to add:
> Client side scripts SHALL NOT be used to create the directory listing.
>> As the content of the directory listing is more or less static, there is little to no benefit in using of client side scripts. Moreover, client side scripts, like JavaScript, are usually not evaluated in text-based browsers and are also hard to check programmatically.

### LINKED ISSUE #897: Fix a few document/lang mentions

### LINKED ISSUE #905: Split optional tests

### LINKED ISSUE #909: `flags` not mentioned in tests
Tests 6.1.1 and 6.1.4. are missing the `flags`-path.

### LINKED ISSUE #920: Put new Co-Chairs on the standard
We recently elected co-chairs. They need to be written on the standard.

tschmidtb51:
- fixes #876 
- fixes #672 
- fixes #897 
- fixes #771 
- addresses parts of https://github.com/oasis-tcs/csaf/issues/652
- fixes #653 
- fixes #863, fixes #780
- fixes #803 
- fixes #864 
- fixes #909 
- fixes #905 
- fixes #920


tschmidtb51:
@sthagen Please generate the full versions under share and feel free to set the motion.

sthagen:
We need to modify some changes made, before we can generate the delivery items, approve, and merge the change set. E.g. many changes to sections have not been propagated to the reference data mappings.

sthagen (approved):
LGTM.

tschmidtb51:
https://groups.oasis-open.org/discussion/motion-for-929

sthagen:
Still going over uite some section title changes, sorry.

sthagen (approved):
LGTM &rightarrow; Ship it!

csaf_2.1/prose/edit/src/tests-03-informative.md
```diff
@@ -508,117 +508,4 @@ The relevant path for this test is:
...
-          }
-        ]
-      }
-    ]
-  }
-```
-
-> The namespace contains the extension `additional-technical-impacts`.
-> Its decision point definitions might therefore not be known to the reader of the document.
-
```
tschmidtb51: @sthagen: Please restore these - it is not a copy-paste mistake: They look like the optional tests but  in fact they are different. The optional tests cover the `TLP:CLEAR` case. These ones here cover the non `TLP:CLEAR` cases.

Reasoning: In a closed user group, the usage of private extensions and private decision points have a valid use case.

tschmidtb51 (commented):
@sthagen Please see my comment

csaf_2.1/prose/edit/src/tests-03-informative.md
sthagen: Ah, sorry. Looked very similar (I did on spot a difference) and then the convenient theory sneaked in of a half-completed move. Problem is, that the headings cannot be left as is (the link part). They need disambiguation from the others. Otherwise, the rendering / display device does not know where to jump to.

sthagen: Done in [b3e4578](https://github.com/oasis-tcs/csaf/pull/929/commits/b3e45783f0a1f67575f1ffb05850ae97a1f87b1c):

```
commit b3e45783f0a1f67575f1ffb05850ae97a1f87b1c (HEAD -> editor-revision-2025-03-26, origin/editor-revision-2025-03-26)
Author: Stefan Hagen <stefan@hagen.link>
Date:   Fri May 2 01:14:02 2025 +0200

    Reverted the removal of two informative test sections

    - fixed an oversight where long section titles with an added term non
      were taken to be duplicates
    - thanks to Thomas Schmidt for noticing.
    - added missing section and example entries to mappings

    Signed-off-by: Stefan Hagen <stefan@hagen.link>
```

tschmidtb51:
@sthagen Thank you so much!

Issue #935, opened 2025-04-11
Consistency: mention `flags` in Product Status


tschmidtb51:
Currently, we don't mention the `flags` in section 3.2.3.9 Product Status for `known_not_affected`. We should do that for consistency.

tschmidtb51:
This was derived from the question by @bs37208 in #932

Issue #926, opened 2025-04-02
Fix example in relationships


tschmidtb51:
The example with Cisco Any Connect mentions the version  `2.3.185` and `4.9.04053`. We should simplify it and align it with the text.

PR #939, opened 2025-04-14
Known Issues for CSAF 2.0

tschmidtb51:
- addresses parts of oasis-tcs/csaf#909
- add missing `flags` path
- correct spelling mistake
- addresses parts of https://github.com/oasis-tcs/csaf/issues/926
- mention potential confusion with example 42
- addresses parts of https://github.com/oasis-tcs/csaf/issues/935
- document `flags` as target for impact statements


sthagen (approved):
LGTM.
Not important: I think a title should be in Title Case and we might drop the "an " in front of the plural for errors (Errata).

PR #940, opened 2025-04-14
Fix SSVC ID for test 6.1.47-15

milux:
Fixed the `id` field in the mandatory positive test 15 JSON to ensure consistency with the document id.


tschmidtb51:
Thank you for your comment. Due to IPR rules, we can't accept code from non-TC members. I'll close that PR and open a new one.

milux:
No problem at all! Just get it fixed, I won't claim any rights for this typo fix. :laughing:

Issue #945, opened 2025-04-15
Questions (& bugs) regarding test 6.1.48 (CSAF 2.1)


milux:
Whilst implementing test 6.1.48 (https://github.com/oasis-tcs/csaf/blob/master/csaf_2.1/prose/share/csaf-v2.1-draft.md#6148-ssvc-decision-points-) the following question(s) came to my mind:

- How are SSVC decision points uniquely identified? Is the tuple `<name, version>` sufficient, or does it have to be `<namespace, name, version>`? The latter could make sense if one expects decision points with identical names in different namespaces.
- If it is meant to be `<namespace, name, version>`, then why do we need to check if `namespace` is one of the "registered namespaces" `{cvss, nciss, ssvc}`? The lookup is bound to fail anyway if the namespace is incorrect, isn't it?
- If it is `<name, version>`, then wouldn't it make more sense to generally check the namespace of the selection against the one given in the referenced decision point?

Finally, found a bug in two of the positive test files for 6.1.48, namely:
```
csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-48-12.json
csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-48-13.json
```
The values (`Negligible`, `Marginal`, ...) belong to `Safety Impact` decision point `2.0.0`, but version `1.0.0` is referenced.
Either the values or the version of the decision point have to be fixed, if this was not meant as a negative test.

milux:
@tschmidtb51 No PR this time. 😉

milux:
> Please also announce your ticket on the [csaf-comment mailing list](https://groups.oasis-open.org/communities/community-home?communitykey=3720cda2-3056-4183-a759-018f5aa7b5f2).

I tried writing to `OASIS-csaf-comment@ConnectedCommunity.org`, but no reaction so far... Anything special I have to know about posting to this mailing list?

tschmidtb51:
> How are SSVC decision points uniquely identified? Is the tuple `<name, version>` sufficient, or does it have to be `<namespace, name, version>`? The latter could make sense if one expects decision points with identical names in different namespaces.

[...]

Issue #944, opened 2025-04-15
Error in test file


domachine:
Hello everyone,

I think there is a typo in the `disclosure_date` string of the test file oasis_csaf_tc-csaf_2_1-2024-6-1-37-13.json.

https://github.com/oasis-tcs/csaf/blob/500de13452015136756e64ae03f84bdc9fd6405b/csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-37-13.json#L34

The date should be `2016-12-31T00:00:60-23:59` (mind the dash instead of the plus)

@tschmidtb51 

Announcement: https://groups.oasis-open.org/discussion/error-in-test-file

PR #946, opened 2025-04-16
Tests - 6.1.37 and 6.1.48

tschmidtb51:
- resolves oasis-tcs/csaf#944
- correct sign mistake in timezone in valid example of 6.1.37
- add additional invalid example for 6.1.37
- resolves https://github.com/oasis-tcs/csaf/issues/945
- correct version of SSVC decision point


sthagen (approved):
LGTM.

PR #947, opened 2025-04-16
Test 6.1.48

tschmidtb51:
- addresses parts of oasis-tcs/csaf945
- clarify text regarding non-registered namespaces
- add valid example


sthagen (approved):
LGTM.

Issue #910, opened 2025-03-24
oasis_csaf_tc-csaf_2_1-2024-6-2-06-02.json in marked as failed but it does not fail


rainer-exxcellent:
The current_release_date is newer, as all dates in the revision history

current_release_date : "2024-01-21T11:00:00.000-01:00"  (2024-01-21T12:00:00.000Z)

revision_history date: "2024-01-21T11:00:00.000+10:00" (2024-01-21T01:00:00.000Z)

rainer-exxcellent:
This has been announced on the CSAF TC comment mailing list: https://groups.oasis-open.org/discussion/add-issue-about-test-oasis-csaf-tc-csaf-2-1-2024-6-2-06-02json

PR #950, opened 2025-04-17
Tests - 6.2.6

tschmidtb51:
- resolves oasis-tcs/csaf#910
- correct mistake in `current_release_time`
- add invalid example
- add valid example


sthagen (approved):
LGTM.

Issue #919, opened 2025-03-27
Clarify empty lines in index.txt


tschmidtb51:
Currently, there is some confusion whether an empty line (e.g. at the end) in an `index.txt` is allowed.

mprpic:
I don't really have a preference, but if you're allowing trailing new lines, then why not just generalize it to "ignore any new lines"?

sthagen:
I am in favor of Martin's proposal to enforce "ignoring any new lines" albeit stated as "ignoring empty lines". In case something blocks us from being more tolerant, I suggest we write:

> An index.txt file SHOULD end with a trailing empty line.

That would help detect truncations and match the often found editing tool behavior of ensuring an EOL sequence at the end of a file upon save.

tschmidtb51:
@bernhardreiter, @oxisto Please provide your implementers view on that.

oxisto:
I would be in favor of a clear stance on this. Having a "SHOULD" in this leave once again room for interpretation, so I would rather be specific that empty lines are not allowed - except as the last line, because as @tschmidtb51 said correctly, often editors add an empty line at the end. Enforcing an empty line on the other hand is probably also not really feasible as most (if not all) known providers seem to NOT end with an empty line. It seems the "offending" provider also fixed its index.txt in the mean time.

bernhardreiter:
Note that POSIX  (IEEE Std 1003.1-2024  [1]) defines a line as:

> A sequence of zero or more non- <newline> characters plus a terminating <newline> character.

So a newline character at the end of the last line does **not** mean a new empty line in that defintion. For implementing it is good if a consistent definition of a line is adopted.

> Some editors always append an empty line at the end.

While the Wikipedia page mentions this behaviour, it does not give examples.  Is this a confusion with editors adhering to standard termination of lines like in POSIX?

Here is an comment that explains why having all lines ending consistently makes the unix tools consistent:
https://www.reddit.com/r/linuxquestions/comments/nworc0/why_do_linux_text_editors_add_a_new_line_to_the/

Assuming that we are really talking about an empty line, which would be represented like
```
2018/example_company_-_2018-yh2312.json\r\n
\r\n
```
or 
```
2018/example_company_-_2018-yh2312.json\n
\n
```

I think it is not necessary to allow.

 Rationale: A detection of unwanted truncation by the empty line is not necessary for the TCP based transfer that is used by HTTP.  It also would have the drawback to requires an empty line in all cases which would complicate implementations.

[...]

PR #957, opened 2025-04-25
Clarify content of index.txt

tschmidtb51:
- resolves oasis-tcs/csaf#919
- clarify that entries are newline sequence terminated
- add hint about multiple index.txt


sthagen (approved):
LGTM.

Issue #955, opened 2025-04-25
Replace undefined "line break"


tschmidtb51:
We use in some places `line break` which is not defined. However, we have a definition of `newline sequence`. We might want to replace that 😎

Issue #14, opened 2019-07-15
CVRF 1.2 - Document Tracking – Revision History – Revision – Date - documented incorrectly


ericjohnson-tibco:
Section 4.5.4.1.2 says "The element cvrf:Date MUST appear exactly once in cvrf:Revision and SHOULD record the date the revision was made and MUST be valid representative of the Date and Time model documented in section 2.2.1 Date and Time. "

However the XMLSchema indicates minOccurs = 0!

Proposal: update specification text to read:
The optional element cvrf:Date can appear exactly once in cvrf:Revision and SHOULD record the date the revision was made and MUST be valid representative of the Date and Time model documented in section 2.2.1 Date and Time.

Alternative proposal - update the schema to remove minOccurs (effectively making it a "1").

sdrees:
Do we always have a date with an updated revision? If so, I would go with the update of the schema, else with the "weasel words" 😄 (commenting from the sidelines ...)

PR #958, opened 2025-04-25
Line Breaks

tschmidtb51:
- resolves oasis-tcs/csaf#955
- replace "line breaks" with "newline sequence"
- fix format of normative statement

### DIFF
--- a/csaf_2.1/prose/edit/src/conformance.md
+++ b/csaf_2.1/prose/edit/src/conformance.md
@@ -154,7 +154,7 @@ Secondly, the program fulfills the following for all items of:
   Sets the value as given in the configuration of the program or the corresponding argument the program was invoked with.
   If values from both sources are present, the program SHOULD prefer the latter one.
   The program SHALL NOT use hard-coded values.
-* `/document/tracking/id`: If the element `cvrf:ID` contains any line breaks or leading or trailing white space,
+* `/document/tracking/id`: If the element `cvrf:ID` contains any newline sequence or leading or trailing white space,
   the CVRF CSAF converter removes those characters.
   In addition, the converter outputs a warning that the ID was changed.
 * `/product_tree/relationships[]`: If more than one `prod:FullProductName` instance is given,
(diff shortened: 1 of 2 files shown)


csaf_2.1/prose/edit/src/schema-elements-02-props-02-document.md
```diff
@@ -627,8 +627,7 @@ Unique identifier for the document (`id`) of value type `string` with 1 or more
 ```
 
 Unique identifier for the document holds the Identifier.
-
-> It SHALL NOT start or end with a white space and SHALL NOT contain a line break.
+It SHALL NOT start or end with a white space and SHALL NOT contain a newline sequence.
```
sthagen, line 630: I would prefer space. The days of print are over and I assume we also do not want horizontal tabs and friends to be there either.

csaf_2.1/prose/edit/src/conformance.md
```diff
@@ -154,7 +154,7 @@ Secondly, the program fulfills the following for all items of:
   Sets the value as given in the configuration of the program or the corresponding argument the program was invoked with.
   If values from both sources are present, the program SHOULD prefer the latter one.
   The program SHALL NOT use hard-coded values.
-* `/document/tracking/id`: If the element `cvrf:ID` contains any line breaks or leading or trailing white space,
+* `/document/tracking/id`: If the element `cvrf:ID` contains any newline sequence or leading or trailing white space,
```
sthagen, line 157: I am not so sure anymore, if our made up trerm "newline sequence" is as helpful as we think. We want unicode (utf-8) and thus the [Unicode Standard Annex \#14 - UNICODE LINE BREAKING ALGORITHM](https://www.unicode.org/reports/tr14/tr14-32.html) might help find us a better term ... or not 🙈.

csaf_2.1/prose/edit/src/schema-elements-02-props-02-document.md
tschmidtb51: I was using the term "white space" as this a fixed term in [EMCA 262](https://262.ecma-international.org/15.0/index.html#sec-white-space) which is the standard for [JSON patterns](https://json-schema.org/understanding-json-schema/reference/regular_expressions).
Should we add that to the terminology? (If so, how?)

PR #970, opened 2025-05-10
Editor revision for TC meeting 2025-04-30

### LINKED ISSUE #622: feature request: add EPSS to vulnerability scores
We would like the optional inclusion of the EPSS score of a vulnerability in the `scores` section within the `vulnerability properties` of a CSAF document.

### LINKED ISSUE #636: Enforce `product_version_range` rules
Currently, we have [rules](https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#31232-branches-type---name-under-product-version-range) how the value of `name` should look like when the `category` is `product_version_range`. But as they are not yet enforced, all kind of values have been put into the field in practice...

Therefore, we need to enforce the rules.

### LINKED ISSUE #638: Consider adding an "unknown" state to CSAF and eventually VEX
There is an expectation when someone selects "under_investigation" that someone is actively investigating the issue (including the vendor). That is not necessarily true. In some cases, vendors will not investigate a vulnerability (e.g., when the product is end-of-life [EoL] or end-of-support [EoS].  Since "under_investigation" is the default value, perhaps we can add a new option called "unknown" or document that "under_investigation" also could mean unknown or that the software provider / vendor will not investigate.

### LINKED ISSUE #722: suggestion: add a type to a link
In the advisory under (document->references) and vulnerability under (vulnerabilities-> references) it is possible to add an external link but it is not possible to specify what the link is about. We think that being able to specify either the type of the link or maybe even create separate categories like references, exploits, news, sources would allow us to make it clear what the link is about instead of just a link to an external source. Because that is not possible now we are using the (summary) to specify what it is including the name, for example “Source – nvd” or “Exploit – nvd” etc..

### LINKED ISSUE #772: Add new profile: Superseded
Looking at #771, it might be helpful to add a profile to inform recipients that a specific document was superseded by a different one.

## Profile S: Superseded

This profile MUST be used for any CSAF document that is superseded.

A CSAF document SHALL fulfill the following requirements to satisfy the profile "Superseded":

* The following elements MUST exist and be valid:
  * all elements required by the profile "CSAF Base".
  * `/document[]/notes` with exactly one item using the `category` `summary` and the `title` `Reasoning for Supersession` describing the original content and the reasons for the supersession
  * `/document/tracking/revision_history` with at least 2 entries. Any previous items MUST NOT be removed.
  * `/document/references` containing at least one item with `category` `external` and `summary` starting with `Superseding document`.
* The value of `/document/category` SHALL be `csaf_superseded`.

### LINKED ISSUE #919: Clarify empty lines in index.txt

### LINKED ISSUE #922: Add some complex version ranges for vls
@thomas-proell pointed out that the examples in vls create the impression that you can't do complex ranges. We should address that.

### LINKED ISSUE #926: Fix example in relationships

### LINKED ISSUE #933: Extended product definition
We should extend the product definition to include services. Then, it would read:

The definition would then read:

**product**: is any deliverable (e.g. software, hardware, specification, or service) which can be referred to with a name. This applies regardless of the origin, the license model, or the mode of distribution of the deliverable.

### LINKED ISSUE #935: Consistency: mention `flags` in Product Status

### LINKED ISSUE #955: Replace undefined "line break"

tschmidtb51:
- fixes #638 
- fixes #922 
- fixes #926
- fixes #935 
- fixes #772 
- fixes #919 
- fixes #955 
- fixes #933 
- fixes #636 
- addresses parts of https://github.com/oasis-tcs/csaf/issues/803
- addresses parts of https://github.com/oasis-tcs/csaf/issues/782
- fixes #622


tschmidtb51:
Motion set at: https://groups.oasis-open.org/discussion/motion-for-970

tschmidtb51:
@sthagen Thank you for fixing the mappings and producing the user facing delivery items

tschmidtb51:
@sthagen I might be wrong - but it looks like I'm missing a review from you on this one.

tschmidtb51:
> Motion set at: https://groups.oasis-open.org/discussion/motion-for-970

And passed with no objections, discussions or concerns.

sthagen (approved):
LGTM.

Issue #592, opened 2022-11-17
aggregator.json


tschmidtb51:
We should consider to make the `role` mandatory in the `metadata_t` of the aggregator.json.

PR #972, opened 2025-05-10
Aggregator

tschmidtb51:
- resolves oasis-tcs/csaf#592
- make `role` mandatory for each entry in the aggregator.json
- update examples


sthagen (approved):
LGTM.

Issue #980, opened 2025-05-15
oasis_csaf_tc-csaf_2_1-2024-6-2-33-13.json is marked as valid but it fails


christopher-exx:
The disclosure_date "2024-02-24T14:00:00.000+07:00" is after the date of the newest item of the revision_history "2024-01-24T10:00:00.000Z", which fails the respective test (optional test 6.2.33).

tschmidtb51:
Thank you for spotting this

PR #982, opened 2025-05-15
Test 6.2.33

tschmidtb51:
- resolves oasis-tcs/csaf#980
- correct valid example for test 6.2.33
- add invalid example for test 6.2.33


sthagen (approved):
LGTM.

Issue #999, opened 2025-05-23
Test 6.2.37: wrong example


tschmidtb51:
The test 6.2.37 has the wrong example and explanation. We need to fix that.

PR #1000, opened 2025-05-23
Test 6.2.37

tschmidtb51:
- resolves oasis-tcs/csaf#999
- correct copy&paste mistake
- update test description from example
- use a better wording


sthagen (approved):
LGTM. Thank you for providing.

Issue #899, opened 2025-03-12
Add grammar check


bernhardreiter:
In addition to spell check a grammar check can be useful (informative).

sthagen:
Maybe. Not so sure, given the often flaky "blue wiggly lines" in some funny apps as compared to the "red wiggly lines" ...

tschmidtb51:
As it is an informative test, I think it helps to specify it. Then, people can rely on a standard test instead of trying to implement something themselves (which would result in multiple incompatible implementations).

tschmidtb51:
@bernhardreiter Please announce and link the csaf-comment mailing list.

bernhardreiter:
Raised on csaf-comment today: https://groups.oasis-open.org/discussion/suggestion-grammer-check

oxisto:
reminder for @oxisto myself: look for suitable open-source tools as possible references.

oxisto:
Next to a lot of "free" online services, the most prominent tool seems to be https://github.com/languagetool-org/languagetool, (which also offers a free online service), but can be installed locally.

Other approaches that I have found include custom made scripts that use "open-source" LLMs such as Llama locally to then check the grammar in a ChatGPT-style.

justmurphy:
Motion to accept #899 as a new informative test into CSAF v2.1 passed 4/30/25 during TC meeting.

PR #1002, opened 2025-05-26
Grammar Check

tschmidtb51:
- resolves oasis-tcs/csaf#899
- add test 6.3.16 as grammar check
- add invalid example
- add valid example
- adapt test schema and json


sthagen (approved):
LGTM.

Issue #597, opened 2022-12-03
Add party email or  `product_id` to the items of `involvements`


tschmidtb51:
The involvements section https://docs.oasis-open.org/csaf/csaf/v2.0/csaf-v2.0.html#3237-vulnerabilities-property---involvements seems unclear to me about how you would identify the name of the vendor or coordination or other `party` that was involved as a stakeholder in the vulnerability disclosure. Can you clarify? May be I am missing something clear in the document. 

For example: Say if I want to mention vendor named "Example" has a `status` of `contact_attempted` on the date 2022-11-16T17:00:00.000Z - will look like below. There is no place to indicate the vendor's name (Example) or official PSIRT email (psirt@example.com) or any unique/readable field to identify the vendor to whom the contact was attempted. If what I see as missing is correct, I recommend adding fields to this involvement section that either tie it to a CSAF PID vendor or a URI field  mailto email address field or a web endpoint where vulnerabilities can be submitted.

```
      "involvements": [
        {
          "party": "vendor",
          "status": "contact_attempted",
          "date": "2022-11-16T17:00:00.000Z",
          "summary": "An email was sent to vendor@example.com and an automated response was received. There was no NDR to indicate any kind of email delivery failure."
        }
```

The example shown in the validator testing at  https://github.com/oasis-tcs/csaf/blob/ba8639f0602ac2cb76b7b8435790615a43987ba4/csaf_2.0/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_0-2021-6-1-24-11.json also does not indicate who actually is the "vendor" that was contacted.

_Originally posted by @sei-vsarvepalli in https://github.com/oasis-tcs/csaf/issues/586#issuecomment-1334458428_

tschmidtb51:
I guess we should add optional fields `contact` as `contact information of the party that was used` and `product_ids`/ `group_ids` to list the products the communication was about.

justmurphy:
Motion to accept change as suggested in #597 passed on 4/30/25 TC meeting.

PR #1007, opened 2025-05-27
Involvements: add options `contact` and `product_ids`/`group_ids`

tschmidtb51:
- addresses parts of https://github.com/oasis-tcs/csaf/issues/597
- add new optional field `contact`, `product_ids` and `group_ids` to schema
- adapt prose
- add missing fields and explanation
- add new fields to guidance on size
- update part of tests (6.1.1 and 6.1.4 by integration)

### DIFF
--- a/csaf_2.1/prose/edit/src/schema-elements-02-props-04-vulnerabilities.md
+++ b/csaf_2.1/prose/edit/src/schema-elements-02-props-04-vulnerabilities.md
@@ -382,18 +382,27 @@ List of involvements (`involvements`) of value type `array` with 1 or more uniqu
 ```
 
 Every Involvement item of value type `object` with the 2 mandatory properties Party (`party`), Status (`status`) and
-the 2 optional properties Date of involvement (`date`) and Summary (`summary`) is a container that allows the document producers to
-comment on the level of involvement (or engagement) of themselves (or third parties) in the vulnerability identification, scoping,
-and remediation process.
+the 5 optional properties Party contact information (`contact`), Date of involvement (`date`), Group IDs (`group_ids`),
+Product IDs (`product_ids`), and Summary (`summary`) is a container that allows the document producers to comment on the level of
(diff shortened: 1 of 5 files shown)


csaf_2.1/prose/edit/src/schema-elements-02-props-04-vulnerabilities.md
```diff
@@ -382,20 +382,29 @@ List of involvements (`involvements`) of value type `array` with 1 or more uniqu
 ```
 
 Every Involvement item of value type `object` with the 2 mandatory properties Party (`party`), Status (`status`) and
-the 2 optional properties Date of involvement (`date`) and Summary (`summary`) is a container that allows the document producers to
-comment on the level of involvement (or engagement) of themselves (or third parties) in the vulnerability identification, scoping,
-and remediation process.
+the 5 optional properties Party contact information (`contact`), Date of involvement (`date`), Group IDs (`group_ids`),
```
sthagen, line 385: I do not remember how we ended up using this extensive Upper Case Usage when followed already by a specific and controlled `term`, but it might enlarge the attention span of the readers when we use normal english case and expand ArtWords like IDs to identifiers if and only if being followed by a parenthiszed controlled term.

Finally-feeling-like-marketing-department-changing-wording-in-the-middle-of-trhe release-processly-yours,
Stefan.

sthagen (approved):
LGTM.

csaf_2.1/prose/edit/src/schema-elements-02-props-04-vulnerabilities.md
tschmidtb51: Those are the exact values of the `title` field from the JSON. If we want to change that, we need to be consistent and do it throughout the standard. Given your release plan and the list of tickets that should still go in there, I'm reluctant to tackle the problem now.
We could have that as separate issue - I guess it would only be a non-material change...

Issue #976, opened 2025-05-13
clarify comparison of `product_identification_helper` objects for test 6.2.32


domachine:
Hello,

for me it's not 100% clear how to compare `product_identification_helper` objects. These objects consist of string arrays and object arrays. I can imagine that the string arrays are seen as "sets"s and therefore for each item the existence is checked independent of the position. But how are the object arrays compared? This is not clear for me from the information provided in section [6.2.32 Use of same Product Identification Helper for different Products](https://github.com/oasis-tcs/csaf/blob/master/csaf_2.1/prose/share/csaf-v2.1-draft.md#6232-use-of-same-product-identification-helper-for-different-products-). Maybe I'm missing something but it might also be an opportunity to improve the spec.

Regards,
Dominik

santosomar:
For the `string-array` properties (`serial_numbers`, `purl`, `model_numbers`, etc.), the intent is indeed to treat them as unordered sets:

`"serial_numbers": [ "A", "B", "C" ]`

is equivalent to

`"serial_numbers": [ "C", "A", "B" ]`

— you test set-equality.

⸻

Object arrays (`hashes`, `file_hashes`, `sbom-urls`)

The spec doesn’t (yet) spell out how to compare arrays of objects (e.g. the hashes array of `{ algorithm, value }` objects).  A sensible interpretation—and the one I recommend we add to the spec—is:
	1.	Treat the array itself as a set of objects.
	2.	Define object-equality as deep, order-insensitive key/value equality.
- Keys within each object can be in any order.
- Any array‐valued fields inside those objects (if they existed) would themselves be set-compared recursively.

So, for example, these two helpers would be considered identical:

```
// Product A helper
"product_identification_helper": {
  "hashes": [
    { "algorithm": "sha256", "value": "aaa" },
    { "algorithm": "md5",    "value": "bbb" }
  ]
}

// Product B helper
"product_identification_helper": {
  "hashes": [
    { "value": "bbb",    "algorithm": "md5"    },
    { "value": "aaa",    "algorithm": "sha256" }
  ]
}

```

Because they carry the same two hash-objects, just in different order.

⸻

I’d suggest adding a short note in section 6.2.32 (or in the general “how to compare arrays” guidance) along the lines of:

Comparison rule for object arrays:

[...]

PR #1014, opened 2025-05-28
Test 6.2.32

tschmidtb51:
- resolves oasis-tcs/csaf#976
- clarify comparison rules

Kudos to @santosomar for the text suggestion


sthagen (approved):
LGTM.

PR #1016, opened 2025-05-28
Editor revision for TC meeting 2025-05-28

### LINKED ISSUE #586: Embargo and Date Public fields for a vulnerability that is private and not yet public
Hello CSAF team,

I have tried to look at the past discussions to find ay information on where embargo dates for a vulnerability disclosure may fit in. It looks like the TLP status is possible with CSAF 2.0, but what about embargo date?  Is this worth considering for interaction between a Vendor, Security Researcher (or other stakeholders) of a vulnerability in private settings. 

Vijay

### LINKED ISSUE #592: aggregator.json

### LINKED ISSUE #597: Add party email or  `product_id` to the items of `involvements`

### LINKED ISSUE #620: Explicit reference `csaf` as ROLIE information type
> The CSAF TC recently requested IANA registration of "csaf" as a ROLIE [...] information type [...]. 
>
> The IANA registry with the new "csaf" element is now available at https://www.iana.org/assignments/rolie/rolie.xhtml.
>
> The IANA reviewers also provided the following comment:
>- "It would be ideal if the definition for the 'csaf' information type was explicit, although the implication here is that the information referenced is a CSAF document. Can this be clarified in the next revision?"
>
> I believe this refers to the text around CSAF v2.0, Section 7.1.15 [2].
>
> [2] CSAF v2.0 reference: https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#7115-requirement-15-rolie-feed

_Originally posted by Paul Knight on the TC mailing list_

### LINKED ISSUE #660: Add version to CWE
CWE list changes over time (see https://cwe.mitre.org/data/archive.html). As we always check against the latest version this might result in invalid documents over time, e.g., if old CWEs get deprecated. To avoid these issues and aid in validation, we should add a version to CWE and declare a minimum acceptable version.

Here is an example, how this would look:

```
      "cwe": {
        "id": "CWE-20",
        "name": "Improper Input Validation",
        "version": "4.11"
      },
```

The JSON schema definition would unfold as:
```
          "cwe": {
            "title": "CWE",
            "description": "Holds the MITRE standard Common Weakness Enumeration (CWE) for the weakness associated.",
            "type": "object",
            "required": [
              "id",
              "name",
              "version"
            ],
            "properties": {
              "id": {
                "title": "Weakness ID",
                "description": "Holds the ID for the weakness associated.",
                "type": "string",
                "pattern": "^CWE-[1-9]\\d{0,5}$",
                "examples": [
                  "CWE-22",
                  "CWE-352",
                  "CWE-79"
                ]
              },
              "name": {
                "title": "Weakness name",
                "description": "Holds the full name of the weakness as given in the CWE specification.",
                "type": "string",
                "minLength": 1,
                "examples": [
                  "Cross-Site Request Forgery (CSRF)",
                  "Improper Limitation of a Pathname to a Restricted Directory ('Path Traversal')",
                  "Improper Neutralization of Input During Web Page Generation ('Cross-site Scripting')"
                ]
              },
              "version": {
                "title": "CWE version",
                "description": "Holds the version string of the CWE specification this weakness was extracted from.",
                "type": "string",
                "minLength": 1,
                "pattern": "\d\.\d{1,3}"
                "examples": [
                  "4.0",
                  "4.11",
                  "4.12"
                ]
              }
            }
          },

### LINKED ISSUE #812: Add test: Consistent PIH
We have seen CSAF documents that have `product_identification_helper` that do not match the product described in `branches`, e.g. the version number is missing in a `CPE` or `purl`. We should add a mandatory (?) test to check for the low-hanging fruits.

### LINKED ISSUE #813: Enforce 'format` validation
The format validation is not strictly required in Draft 2020-12, even though many test tools check it. As we expect the `format` restrictions to be applied, we should enforce the check.

[This website](https://www.learnjsonschema.com/2020-12/format-assertion/format/) describes how that can be achieved.

### LINKED ISSUE #843: Empower providers to provide advisories in different CSAF versions in parallel
During the CSAF Community Days 2024, 12.-13. December 2024 in Munich, it became clear, that providers often map directories directly served per HTTPS to SCAF format versions.

> Examples: `foo/v2/bar` or `baz/2.0/quux`.

>Other providers do not indicate any CSAF format version in the paths or the provider.json
resource.

All these use cases are plausible and fit the provider and consumer needs it seems,
but when CSAF v2.1 will be available there is the reasonable assumption, that for some
time out of control of the TC there will be many mixed format situations:

- Only CSAF v2.0
- CSAF v2.0 for older advisories and CSAF v2.1 for "fresher" advisories
- CSAF v2.0 for all advisories and CSAF v2.1 for "fresher" advisories
- Only CAF v2.1

Questions to seed the discussion:

- Do we need to adjust the provider json schema of CSAF v2.1?
- Do we need to amend also the CSAF v2.0 provider json schema?
- Do we lack informative text in our CSAF v2.1 prose to guide and support the communities?
- Would the CSAF eco system profit from FAQ entries providing solutions for these scenarios?

### LINKED ISSUE #851: Add a first known exploitation date
This is split out from #782:

It is suggested to add an optional field that provides the opportunity to convey the information, when a vulnerability was first exploited in a specific product. A JSON structure could look like this:

```
          "known_exploitations": {
            "title": "List of known exploitations",
            "description": "Contains a list of known exploitations",
            "type": "array",
            "minItems": 1,
            "uniqueItems": true,
            "items": {
              "title": "Flag",
              "description": "Contains information on when this vulnerability was first known to be exploited in the product.",
              "type": "object",
              "required": [
                "date",
                "exploitation_date"
              ],
              "properties": {
                "date": {
                  "title": "Date of the information",
                  "description": "Contains the date when the information was last updated.",
                  "type": "string",
                  "format": "date-time"
                },
                "exploitation_date": {
                  "title": "Date of the exploitation",
                  "description": "Contains the date when the exploitation happened",
                  "type": "string",
                  "format": "date-time"
                },
                "group_ids": {
                  "$ref": "#/$defs/product_groups_t"
                },
                "product_ids": {
                  "$ref": "#/$defs/products_t"
                }
              }
            }
          },
```

### LINKED ISSUE #877: Set explicit `"additionalProperties": "false"`
In CSAF 2.0, we already disallow "additionalProperties". However, that is currently done via a test. We should integrate that into the schema to reduce confusion and surprises when running the tests. This will also help implementers to generate their data classes directly from our schema..

### LINKED ISSUE #882: Explicitly name test presets
We define certain test presets through our conformance targets (9.1.14-9.1.16):
- schema
- basic
- extended
- full
Some are derived from the tests itself (6.1-6.3):
- mandatory
- optional
- informative

Some other might be useful:
- external-request-free (any test that does not do a request into the internet)
- consistent-revision-history (basic+6.2.4-6.2.6)

We should clarify the presets to help implementers produce consistent tools across the ecosystem.

### LINKED ISSUE #884: Proposition to add a license key in the provider-metadata.json file
Hello,

I think that it could be interesting to give the possibility to specify the license of a feed in the provider-metada.json file. What do you think ?

In Vulnerability-Lookup we are currently checking the licenses used by our various sources:
https://vulnerability.circl.lu/about#sources

And sometimes it is challenging to find a license somewhere.
We are building datasets for NLP processing and technically we should check the compatibility of the licenses involved in each dataset. For example if we want to create an "enriched" dataset with data from various sources.

As mentioned [here](https://github.com/oasis-open/csaf-documentation/issues/85#issuecomment-2692427482) by @tschmidtb51 there is as well the possibility to have that in each individual CSAF document.

Thank you.

### LINKED ISSUE #899: Add grammar check

### LINKED ISSUE #916: Find better name for optional tests
We are currently using the name ["optional" tests](https://github.com/oasis-tcs/csaf/blob/master/csaf_2.1/prose/share/csaf-v2.1-draft.md#optional-tests) but in fact they are "SHOULD" tests. We should find a clearer wording.

Carved out from our discussion in today's TC meeting on #882.

### LINKED ISSUE #963: How to deal with leap seconds in CSAF?
While writing tests for 6.1.36 (see #960) we encountered a problem with leap seconds. Test case 06 was introduced for 6.1.36 which includes a CSAF document with a RFC 3339 date with an invalid leap second, i.e., a date string with "60" seconds at a non-leap second timestamp.

Unfortunately, the state of "correct" implementation of RFC 3339 is inconsistent and not 100 % complete in the programming ecosystems. If one would correctly implement RFC 3339, a "60" seconds date string would only be allowed at the official list of leap seconds defined by the International Earth Rotation and Reference Systems Service and published by NIST [1].

The following is a short list of how popular libraries support or not support leap seconds:
- Rust `time` and `chromo`: All timestamps that contain a "60"th second are parsed and 60 is replaced by 59 / 59.999.999
- Python `date time`: All timestamps with "60" seconds are regarded as invalid (even valid ones)
- Go: TBD - @bernhardreiter ?

Basically, the wider (open-source) community decided that the extra effort of implementing RFC 3339 is not worth the benefit and choose to "ignore" this aspect of the specification by either not allowing leap seconds at all or parsed them slightly incorrect.

Currently, we specify in CSAF that
- A date must be RFC 3339 compliant (therefore leap seconds should be supported in theory), but
- Leap seconds SHOULD be avoided if possible
- We also include a test that MUST fail if an invalid leap second is used

This currently leads to the following problems:
- Implementations in python can most likely not follow our specification because they would throw an error for a VALID leap seconds (without custom date parsing)
- Implementations in rust sort of work but are currently NOT throwing an error for an INVALID leap second, making it currently impossible to write a compliant validator (without custom date parsing)

This begs the question how we should proceed with the support of leap seconds? My 2 cents would be that it is probably easier to either forbid leap seconds or not to include an explicit test for invalid ones. Two things should be considered:
- Actually the possibility that a CSAF document is generated at a VALID leap second is really not that high
- We actually do not need accurate time-stamps in a per-second resolution. Even standards like JWT, where there is explicit demand for time-based evaluation of tokens (expiry, issued-at), the standard allows a lee-way of several seconds to account for clock skew between clients and servers in a distributed system.

[1]: https://www.nist.gov/pml/time-and-frequency-division/time-realization/leap-seconds

Announced in https://groups.oasis-open.org/discussion/dealing-with-leap-seconds-in-csaf

### LINKED ISSUE #966: Clarify CWE `name` should be trimmed
Currently, it is not clear that a CWE `name` MUST not start nor end with white space. We need to clarify that.

### LINKED ISSUE #976: clarify comparison of `product_identification_helper` objects for test 6.2.32

### LINKED ISSUE #978: Unify phrasing for strings with `format`
I realized that we use multiple different phrases around `value type` and `format`:
- `of value type [...] with format`
- `with value type [...] of format`

@justmurphy suggested to use `of value type [...] with format`

### LINKED ISSUE #980: oasis_csaf_tc-csaf_2_1-2024-6-2-33-13.json is marked as valid but it fails

### LINKED ISSUE #987: Clarify handling multiple entries in security.txt
I propose to clarify the handling of multiple entries in the security.txt (Requirement 8 and possibly in 7.3.1). This is something that we usually try to avoid but will probably come up once providers transition from 2.0 to 2.1 (see also #843). I would like to clarify / change:
- "If one of the URLs fulfills requirement 9, this MUST be used as the first CSAF entry in the security.txt." to "If one of the URLs fulfills requirement 9, this MUST be the first CSAF entry in the security.txt.". This clearly states that the well known must be the first entry in the file - previously it was ambiguous whether it is the first entry or whether it will be handled ("used") as the first entry
- We should explicitly state that the entries are sees a a list and that the list should be handled in order as they are specified. If one entry is not readable / parseable (e.g., because it is a future version), the next one should be used
- Lastly (but this should be done in #843). We should explicitly mention the use-case of transitioning from 2.0 to 2.1 as one where multiple entries in the security.txt are accepted, for example "However, this SHOULD NOT be done and removed as soon as possible. A valid use-case includes the transitioning from CSAF 2.0 to CSAF 2.1"

### LINKED ISSUE #988: Split distribution source files
Given that https://github.com/oasis-tcs/csaf/issues/843 wants to add a section in the distribution chapter, we should split those files to make them easier in handling.

### LINKED ISSUE #992: How to handle transition from CSAF 2.0 to CSAF 2.1 for `aggregator.json`
This issue is split from #843 because of the added complexity. Similar to handling the transition in the PMD, we also need to handle it for aggregators.

### LINKED ISSUE #995: Define white space
We are using the term "white space" as this a fixed term in [EMCA 262](https://262.ecma-international.org/15.0/index.html#sec-white-space) which is the standard for [JSON patterns](https://json-schema.org/understanding-json-schema/reference/regular_expressions).

We need to clarify that.

### LINKED ISSUE #999: Test 6.2.37: wrong example

### LINKED ISSUE #1003: Editorial 2025-05-28
A place to track various editorial things that came up:

- [x] use title case in 6.3.8
- [x] make custom dictionary mandatory for CSAF full validators

tschmidtb51:
- fixes #813 
- addresses parts of #616
- fixes #592
- fixes #916 
- fixes #978
- fixes #963 
- fixes #980 
- fixes #877 
- fixes #988
- fixes #987 
- fixes #843 
- fixes #966 
- fixes #660
- addresses parts #652
- fixes #995 
- fixes #851 
- fixes #999 
- addresses parts of https://github.com/oasis-tcs/csaf/issues/803
- fixes #992 
- fixes #899 
- fixes #1003 
- fixes #597 
- fixes #586 
- fixes #812 
- fixes #884 
- fixes #620 
- fixes #976 
- fixes #882

### DIFF
--- a/csaf_2.1/prose/share/csaf-v2.1-draft.html
+++ b/csaf_2.1/prose/share/csaf-v2.1-draft.html
@@ -821,6 +821,16 @@
                 </li>
               </ul>
             </li>
+            <li>6.4 <a href="#presets">Presets</a>
+              <ul>
+                <li>6.4.1 <a href="#presets-defined-through-test-subsections">Presets defined through Test Subsections</a>
+                </li>
+                <li>6.4.2 <a href="#presets-defined-through-conformance-targets">Presets defined through Conformance Targets</a>
+                </li>
+                <li>6.4.3 <a href="#additional-presets">Additional Presets</a>
+                </li>
+              </ul>
+            </li>
           </ul>
         </li>
         <li>7. <a href="#distributing-csaf-documents">Distributing CSAF documents</a>
@@ -6474,7 +6484,7 @@
       6. Tests <a id='tests'></a>
     </h1>
     <p>
-      The following three subsections list a number of tests which all will have a short description and an excerpt of an example which fails the test.
+      The first three subsections list a number of tests which all will have a short description and an excerpt of an example which fails the test. The forth subsection groups tests into preset.
     </p>
     <h2 id="61-mandatory-tests-">
       6.1 Mandatory Tests <a id='mandatory-tests'></a>
@@ -9711,7 +9721,7 @@
     </h3>
     <blockquote>
       <p>
-        The TLP label is now required by the schema. Therefore, the recommended test is obsolete. This section is kept to document that change and keep the numbering of the remaining sections stable.
+        The TLP label is now required by the schema. Therefore, the recommended test is obsolete. This section is kept to document that change and keep the numbering of the remaining sections stable. The test is excluded from any preset and requirement to be executed.
       </p>
     </blockquote>
     <h3 id="6211-missing-canonical-url-">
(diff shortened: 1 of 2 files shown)


sthagen (approved):

tschmidtb51:
During the TC meeting on 2025-05-28, a motion was set by Thomas Schmidt ans seconded to accept this Editor Revision into our `master` branch.

Issue #341, opened 2021-08-23
Add valid examples


tschmidtb51:
We should add also valid examples in https://github.com/oasis-tcs/csaf/tree/master/csaf_2.0/test/validator/data to aid in implementing tools.

PR #1055, opened 2025-06-16
Add test data for `flags` in 6.1.1 and 6.1.4

tschmidtb51:
- addresses parts of https://github.com/oasis-tcs/csaf/issues/909, https://github.com/oasis-tcs/csaf/issues/341
- add valid examples for 6.1.1/6.1.4
- add invalid examples for 6.1.1/6.1.4 to detect result in flags
- add valid example for 6.1.1/6.1.4 to detect result in flags

### DIFF
--- a/csaf_2.0/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_0-2021-6-1-04-02.json
+++ b/csaf_2.0/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_0-2021-6-1-04-02.json
@@ -1,0 +1,64 @@
+{
+  "document": {
+    "category": "csaf_base",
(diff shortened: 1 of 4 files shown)


tschmidtb51:
@sthagen This is in the "known issues" list of CSAF 2.0. Nevertheless, I think that the test files should be added as this helps to point out what was intended.

sthagen (approved):
LGTM. We could technically target the upcoming editorial branch first, but as the test data is outside of the work product, it may be best for implementers to already have that data available in master.

tschmidtb51:
Agreed - this is outside of the CSAF 2.1 work product and should be handled as any other CSAF 2.0 test addition.

Issue #1085, opened 2025-07-02
Inconsistencies in Section 6.1.53 – Example and unclear status dependency


bendo-eXX:
In section 6.1.53 "_Inconsistent Exploitation Date_" of the CSAF 2.1 specification, I noticed two potential issues:

**1. The example is correct but described as failing**
    The provided example claims that the validation fails, although the condition `exploitation_date `<= `date` is actually met.

**2. Unclear role of the status field**
    The `status `(e.g., "`final`") is mentioned only in the description of the failure, but not in the rule itself. It is unclear whether the document status influences the validation and should be explicitly addressed.

bendo-eXX:
This has been announced on the CSAF TC comment mailing list: https://groups.oasis-open.org/discussion/inconsistencies-in-section-6153-example-and-unclear-status-dependency-1085

PR #1096, opened 2025-07-11
Test 6.1.53

tschmidtb51:
- resolves oasis-tcs/csaf#1085
- remove status from explanation (as it was a copy paste leftover)
- update example from test data (as it was a copy paste mistake as well)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -18,7 +18,7 @@ indent_style = space
 indent_size = 2
 
 # Makefiles require tabs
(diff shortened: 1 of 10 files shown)


tschmidtb51:
Motion set at: https://groups.oasis-open.org/discussion/csdpr01-feedback-resolution-1096

justmurphy (approved):
lgtm. Motion passed.

Issue #1052, opened 2025-06-13
6.3.13: remove inconsistent limitation


tschmidtb51:
In 6.3.13, we limit ourselves to `ssvc`for checking where as we should cover any registered namespace (or at least any register one that has decision points defined).

PR #1105, opened 2025-07-18
Limitation in Test 6.3.13

tschmidtb51:
- resolves oasis-tcs/csaf#1052
- remove limitation to `ssvc` namespace
- rephrase to clarify that registered namespaces are in scope


tschmidtb51:
Motion set at: https://groups.oasis-open.org/discussion/csdpr01-feedback-resolution-1105

Issue #1053, opened 2025-06-13
6.3.12: Clarify relationship CVSS vs. SSVC


tschmidtb51:
Currently, we specify that each metric item should have a CVSS v4.0. However, that it incorrect - especially if it comes to SSVC. We need to rephrase that.

tschmidtb51:
Resolved through #1108

PR #1154, opened 2025-08-23
Comment Resolution CSD01PR: #1053

tschmidtb51:
- resolves oasis-tcs/csaf#1053
- add restriction to exclude content objects without CVSS
- add requirement that at least one CVSS needs to be there per affected product
- add additional examples

### DIFF
--- a/csaf_2.1/prose/edit/src/tests-03-informative.md
+++ b/csaf_2.1/prose/edit/src/tests-03-informative.md
@@ -421,7 +421,7 @@ The relevant paths for this test are:
 ### Missing CVSS v4.0
 
 For each item in the list of metrics that contains any CVSS object it MUST be tested that a `cvss_v4` object is present.
-The test MUST fail, if any Product ID (type `/$defs/product_id_t`) in the product status group Affected is not covered by
+The test MUST fail, if any Product ID (type `/$defs/product_id_t`) in the product status group Affected (see section [sec](#vulnerabilities-property-product-status)) is not covered by
 any CVSS object.
 
 The relevant path for this test is:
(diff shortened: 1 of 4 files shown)


csaf_2.1/prose/edit/src/tests-03-informative.md
```diff
@@ -420,7 +420,9 @@ The relevant paths for this test are:
 
 ### Missing CVSS v4.0
 
-For each item in the list of metrics it MUST be tested that a `cvss_v4` object is present.
+For each item in the list of metrics that contains any CVSS object it MUST be tested that a `cvss_v4` object is present.
+The test MUST fail, if any Product ID (type `/$defs/product_id_t`) in the product status group Affected is not covered by
```
sthagen: Not sure, if double quotes around the group name or injecting 
>with the name
before Affected would be smoother.

tschmidtb51: We use the same/similar wording regarding the product status groups in multiple places. Therefore, I think this should be sufficient. However, we could add (as a service for the reader a direct link to the product status groups definition:

```suggestion
The test MUST fail, if any Product ID (type `/$defs/product_id_t`) in the product status group Affected (see section [sec](#vulnerabilities-property-product-status)) is not covered by
```

Thoughts?

justmurphy (approved):

justmurphy:
Motion set by Thomas Schmidt during Monthly CSAF TC Meeting on 2025-08-27.
Christoph Plutte seconded. No objections, no discussion. Motion passed.

Issue #1077, opened 2025-07-01
Add `extensions`


tschmidtb51:
To be faster with adopting new standards (e.g. CVSS 4.1 etc.), we should add an extension option. Even though I dislike a situation where a tool can't be sure whether is is able to process a CSAF document, I see the need for this as the ecosystem is constantly evolving.

To avoid a situation, where publishing parties just invent new attributes all the time - as this would counter the intend of having one standardized format - safeguards have to be implemented:

1. Extensions are versioned - each has a publicly available schema that can be use to validate the extension.
2. There are 2 kind of extensions: Official ones from the TC (e.g. marked by a property name beginning with `csaf_`) and experimental ones (e.g. marked by a property name beginning with `x_<orgname>`).
3. If the use of a extension would result in any failing mandatory (or recommended) test, it MUST be marked as critical. Any tool, which does not understand this extension, MUST fail when processing the CSAF document (unless forced by the user to do otherwise).
4. A tool MAY ignore safely any information in an uncritical extension it does not understand. If the author values the information given in the extension as critical for the understanding of the CSAF document, it MUST be marked as critical.
5. The TC maintains a list of official TC extensions.
6. The TC maintains a list of deprecated extensions. Such extensions can still be used but support for them is removed in the near future.
7. The TC maintains a list of deny-listed extensions. Such extensions MUST NOT be used.
8. The TC is free in its choice to put any extensions (also experimental ones) on any of those list. (e.g. if there is an official property for the "information" previously conveyed by the extension, then the official property MUST be used. Otherwise, it makes interoperability hard.)

As a first suggestion, extensions should be possible at the following paths:
```
/
/document/
/$defs/full_product_name_t/
/vulnerabilities[]/
```

bernhardreiter:
Is the extension mechanism a _standardization light_? It seems a bit like it. Instead wouldn't it be easier to to faster minor releases of the standard. In case of CVSS 4.1  it seems fast enough, because it will come with some lead-time.

However to evaluate if an extension mechanism is really needed and worth the effort, I'd personally would first look at more real life use cases and discuss those. Only CVSS was mentioned.

tschmidtb51:
@dennypage I chose that wording deliberately. While I see your concern that we don't want to track each and every extension that has been introduced by a random party, we might want to deprecate some experimental extensions - especially when their equivalent becomes part of the standard ;-)

tschmidtb51:
[...]

PR #1165, opened 2025-08-27
Extensions

tschmidtb51:
*just a quick draft for initial review*
*see todo list at https://github.com/oasis-tcs/csaf/issues/1077#issuecomment-3227434576*


sthagen (approved):

sthagen (approved):

csaf_2.1/json_schema/extension-content.json
sthagen, line 1: We could name it simply `csaf_2.1/json_schema/extension.json` and in case we later create an extra schema for secondary concerns like extensions-available name such a schema accordingly using a two-term form.

csaf_2.1/json_schema/csaf.json
```diff
@@ -335,6 +345,11 @@
             }
           },
           "additionalProperties": false
+        },
+        "x_extensions": {
```
sthagen, line 349: Why do we prefix the hierarchy of four slot levels for extensions with an x? This feels awkward. I thought extensions go into the array of the key extensions. So, it would be logically:

```yaml
extensions:
- x-foo
- x-bar
- x-baz
- x-quux
```

Did I misunderstand something? Thanks.

csaf_2.1/json_schema/extension-content.json
tschmidtb51: Hmm, I have the extension-schema.json schema on my list to be implemented...

The content schema checks that the JSON instance has the correct content where as the schema schema checks the schema of an extension... I know - this is confusing, I'll have to think more about that.

csaf_2.1/json_schema/csaf.json
tschmidtb51: `x_extensions` is an array that holds a list of extensions. However, to show that this is something extra and to make it the last element in our lexicographical sorting, I prefixed it with `x_`. This way, it looks like:

```
$schema
document
product_tree
vulnerabilities
x_extensions
```
instead of
```
$schema
document
extensions
product_tree
vulnerabilities
```
The latter feels like extensions would be an integral part of the spec and could (and should) be used in every document which is not the intention.

csaf_2.1/prose/edit/src/design-considerations-04-extensions.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+The value `critical` indicates, that the content provided through this extension is crucial to understand of the CSAF document this extension is included in.
+CSAF consumer MUST warn if they process a CSAF Document including such extension instance and do not have the extension in question already implement.
+
+The value `high_value` indicates, that the content provided through this extension is highly relevant and significantly aids in understanding the
+overall content of the CSAF document.
+
+The value `informational` indicates, that the content provided through this extension just provides additional information.
+CSAF consumer SHOULD warn if they process a CSAF Document including such sn extension instance and do not have the extension in
```
tschmidtb51:
```suggestion
CSAF consumer SHOULD warn if they process a CSAF Document including such an extension instance and do not have the extension in
```

csaf_2.1/json_schema/csaf.json
sthagen: Independent of kebab-case, we could use a dash then to show that and have the sorting: `x-extensions` (If I were younger, maybe `xtensions` might appeal to me 🙈).

But, I think extensions as slot should be embraced, and the array length itself should be minimized by users. I am still in favor of extensions to not micro-manage the perception of members.

sthagen (approved):

sthagen (approved):

sthagen (approved):
x-yes

sthagen (approved):

csaf_2.1/prose/edit/src/design-considerations-01-construction-principles.md
```diff
@@ -68,6 +68,11 @@ The JSON schemas defined in this standard do not allow the use of additional pro
 > The standardized fields allow for scalability across different issuing parties and dramatically reduce the human effort and
 > need for dedicated parsers as well as other tools on the side of the consuming parties.
 
+Sole exception are the dedicated extension points (`x_extensions`) which allow the usage of extension defined outside this specification.
+To still ensure interoperability and guide tools how to deal with this data a CSAF Extension Content Schema
+and CSAF Extension Metaschema has been defined and must be adhered to (cf. section [sec](#extensions)).
```
sthagen: Uppercase `MUST`? If not, we should reword for disambiguation from requirements.

tschmidtb51:
@sthagen I tried to minimize merge conflicts. Maybe it makes sense to provide the rendered versions from a separate branch? :thinking:

sthagen (approved):

sthagen:
I will merge this change-set and look forward to the proposal for the life cycle management of such extensions by the TC. I understand that this proposal will include guardrails to constrain extensions to core security needs in contrast to unrelated foreign payloads.

Issue #1179, opened 2025-10-01
Clarify obligation to insert correct revision history


tschmidtb51:
Some CSAF multipliers do not modify the revision history while republishing. This creates significant issues for incremental downloaders as they miss those files. Incremental downloaders use the timestamps given in the `current_release_date` (as they are stated in the ROLIE feed and changes.csv) and compare them with their last run. As new CSAF documents must have a later `current_release_date` than their last run (otherwise they would have caught them in the last run), they stop when they reach the time of their last run (potentially minus a grace period).

If a CSAF multiplier inserts documents way before the timestamp of their last run (and therefore changes the ROLIE feed/changes.csv with new content afterwards), incremental downloader are missing these files.

PR #1197, opened 2025-10-28
Clarify Revision History

tschmidtb51:
- resolves oasis-tcs/csaf#1179
- clarify the need for a new revision history for modified CSAFs

### DIFF
--- a/csaf_2.1/prose/edit/src/schema-elements-02-props-02-document.md
+++ b/csaf_2.1/prose/edit/src/schema-elements-02-props-02-document.md
@@ -779,8 +779,9 @@ Build metadata SHOULD NOT be included in the `number` of any revision item.
 Any issuing party using a CSAF document as basis and modifying it MUST create a new revision history tracking the modified document.
 This applies especially to CSAF multiplier.
 
-> The new revision history ensures that the CSAF document can be found and downloaded by CSAF downloaders that retrieve just those
-> CSAF documents that have a newer `current_release_date` than the timestamp of their last visit at this source (incremental download).
+> The new revision history (and consequently the corresponding ensures `current_release_date`) that the CSAF document can be found
+> and downloaded by CSAF downloaders that retrieve just those CSAF documents that have a newer `current_release_date` than the timestamp
+> of their last visit at this source (incremental download).
 
 ##### Document Property - Tracking - Status
 


csaf_2.1/prose/edit/src/schema-elements-02-props-02-document.md
```diff
@@ -776,6 +776,12 @@ Versions of the document which are pre-release SHALL NOT have its own revision i
 All changes MUST be tracked in the item for the next release version.
 Build metadata SHOULD NOT be included in the `number` of any revision item.
 
+Any issuing party using a CSAF document as basis and modifying it MUST create a new revision history tracking the modified document.
+This applies especially for CSAF multiplier.
```
sthagen: I think that in this case the preposition is `to` and not `for`.

csaf_2.1/prose/edit/src/schema-elements-02-props-02-document.md
```diff
@@ -776,6 +776,12 @@ Versions of the document which are pre-release SHALL NOT have its own revision i
 All changes MUST be tracked in the item for the next release version.
 Build metadata SHOULD NOT be included in the `number` of any revision item.
 
+Any issuing party using a CSAF document as basis and modifying it MUST create a new revision history tracking the modified document.
```
sthagen, line 779: Not sure, but I would have thought that modifying a CSAF document would lead to a new entry appended to the incoming revision history and not a removal of the latter and replacement with a completely new history, no?

csaf_2.1/prose/edit/src/schema-elements-02-props-02-document.md
```diff
@@ -776,6 +776,12 @@ Versions of the document which are pre-release SHALL NOT have its own revision i
...
+This applies especially for CSAF multiplier.
+
+> The new revision history ensures that the CSAF document can be found and downloaded by CSAF downloaders that retrieve just those
```
sthagen: I think also in this case, I would expect that new entries to a revision history and thus an added later `current_release_date` would be enough to trigger matching in what's new searches and thus lead to download as updated. 
So in short: Can we instead write:

> The new revision history entry and the corresponding `current_release_date` ensure that ...

csaf_2.1/prose/edit/src/schema-elements-02-props-02-document.md
tschmidtb51: No. The revision history MUST reflect the factual revision history of the document. So, in 99% it is a new one.

csaf_2.1/prose/edit/src/schema-elements-02-props-02-document.md
tschmidtb51: No. As the revision history reflects the revision of this document such thing would create confusion. Let's assume, that we have a publisher X that creates a document ABC at `2025-10-14T10:00:00Z` (=> `current_release_date` and `initial_release_date`). A multiplier Y picks the file up a few days later and decides to republish it (as DEF). The multiplier was last collected by (incremental) downloader Z (customer) at `2025-10-20T10:00:00Z`.

Let's take a look at the following scenarios:

### Scenario 1: Revision History kept
Y keeps the revision history as is. Z can't find DEF as it is inserted into the distribution metadata before `2025-10-20T10:00:00Z` as DEF's `current_release_date` is `2025-10-14T10:00:00Z`.

### Scenario 2: Revision history with new entries
Y keeps the revision history as is (from ABC) and adds a new entry for its release. (That entry just exists in DEF, not in ABC.) Consequently, Z can find DEF. However, there might be a compliance issue as the revision history lists a different version for the "initial release to the intended target group." I. E. a company could be punished for not taking action at a certain point in time (on DEF) but actually be innocent as DEF was published later - just the revision history items that were carried over from ABC make it seem like it was available earlier.
Also, this created problems regarding the diff and as soon as upstream X creates a new version of ABC: This revision entry might contain the number that is already assigned in the revision history of DEF. After a while if becomes unclear to consumers which revision history items belong to DEF itself and which are inherited.

### Scenario 3: New revision history
Y starts with a fresh and empty revision history for DEF. Z can automatically find and retrieve the DEF. The `current_release_date` and `initial_release_date` as well as the revision history reflect the history of DEF.

sthagen: I think secnario 1 is clearly out. Scenarios 2 and 3 seem to both have their trade offs. No. 2 is similar to git repositories with divergent histories, no? No. 3 is like plagiarism, it can lead to "old wine in new bottles" dispatch problems, when a multiplier goes crazy (bug) and derives thousands of advisories from a single incoming document upon each receipt.

I get, that scenario 2 is kind of intertwined (as is the reality of reusing documents with a history of their own).

My spontaneous reaction was really only triggered by ideas of ownership, continuation of history (traceability) and the assumption, that the publication by the multiplier would be under a new name (which may be a silly assumption).

In summary, please feel free to merge as is, I only thought it might be a left out word, while it seems I misunderstood the principle of operation in real-world scenarios.

sthagen (approved):

csaf_2.1/prose/edit/src/schema-elements-02-props-02-document.md
tschmidtb51: > No. 2 is similar to git repositories with divergent histories, no?

Kinda, but here we create collisions that need to be fixed for every release.

> No. 3 is like plagiarism, 

I wouldn't call it "plagiarism" as modifying a CSAF document is intended - e.g. to add your own knowledge.

> it can lead to "old wine in new bottles" dispatch problems, when a multiplier goes crazy (bug) and derives thousands of advisories from a single incoming document upon each receipt.

Yes. That can happen. Maybe, we should introduce a conformance target for such a tool that helps the multiplier to keep track...

> In summary, please feel free to merge as is, I only thought it might be a left out word, while it seems I misunderstood the principle of operation in real-world scenarios.

Your feedback is appreciated as always - such misunderstandings help to uncover parts that are not well-specified.
I intended to bring that to the TC tomorrow.

Issue #1201, opened 2025-10-29
Clarify requirement 18 as case-sensitive


tschmidtb51:
We should clarify requirement 18 whether we expect the hash to be case sensitive.

tschmidtb51:
I tend to "just lower case" right now.
@sthagen opinions?

PR #1227, opened 2025-11-24
Requirement 18: case-sensitive

tschmidtb51:
- resolves oasis-tcs/csaf#1201
- clarify that the hash is expected to be lower case


tschmidtb51:
@sthagen / @justmurphy Do you consider that to be an "editorial"?

Issue #1208, opened 2025-11-07
Refactor external link to SSVC repository into informative reference


sthagen:
We should make this externaql link into a reference entry in informative references.

Here: https://github.com/oasis-tcs/csaf/blob/master/csaf_2.1/prose/share/csaf-v2.1-draft.md#6313-usage-of-non-latest-ssvc-decision-point-version-

First informative comment

6.3.13 Usage of Non-Latest SSVC Decision Point Version

For each SSVC decision point given under selections with a registered namespace, it MUST be tested the latest decision point version available at the time of the timestamp was used. The test SHALL fail if a later version was used. Namespaces reserved for special purpose MUST be treated as per their definition.

A list of all valid decision points of registered namespaces including their values is available at the [SSVC repository](https://github.com/CERTCC/SSVC/tree/main/data/json/decision_points).

tschmidtb51:
It also appears in 6.1.48.

tschmidtb51:
@sthagen I unassigned you and assigned myself as I was already working in that area to clean up SSVC.

Issue #1182, opened 2025-10-21
List of registered SSVC namespaces in mandatory test 6.1.48 differs from list in SSVC documentation


christopher-exx:
According to [mandatory test 6.1.48](https://github.com/oasis-tcs/csaf/blob/master/csaf_2.1/prose/edit/src/tests-01-mndtr-48-ssvc-decision-points.md), there are three registered SSVC namespaces. In contrast, the [SSVC documentation](https://certcc.github.io/SSVC/reference/code/namespaces/#registered-namespace) lists much more registered namesspaces.

Issue #1175, opened 2025-09-15
Link to Decision_Point_Value_Selection-2-0-0.schema.json has changed


rainer-exxcellent:
The link to Decision_Point_Value_Selection-2-0-0.schema.json referenced at 

https://github.com/oasis-tcs/csaf/blob/316c1007055a7f0a8defc2eec9750bed36ee73f6/csaf_2.1/prose/edit/src/schema-elements-02-props-04-vulnerabilities.md?plain=1#L642

has changed.

See discussion in: https://github.com/CERTCC/SSVC/discussions/947

sthagen:
So, the correct link is 

https://raw.githubusercontent.com/CERTCC/SSVC/refs/heads/main/data/schema/v2/SelectionList_2_0_0.schema.json

for some time ...

tschmidtb51:
Todo:

- [] update metrics section
- [] update 2.1
- [] make sure, we test against a fresh copy

tschmidtb51:
This is just deal with in #1228.
It is related to #1152.

Issue #1152, opened 2025-08-23
Switch to SSVC v2


tschmidtb51:
Currently, we are on SSVC v1 but the most current one is SSVC v2 - we need to make the switch right now to solve some issues.

PR #1228, opened 2025-11-24
SSVC v2 - clean up

tschmidtb51:
- addresses parts of oasis-tcs/csaf#1152, oasis-tcs/csaf#1175 
  - update to latest schema
  - update to new `$id`
  - update tests
  - adapt schema
  - adapt prose
  - adapt examples to be valid decision points against the schema again
  - replace `description` by `definition` as per SSVC keyword change
- fixes https://github.com/oasis-tcs/csaf/issues/1182
  - replace mentioned namespaces with link to documentation
  - add link as informative reference
  - fix format issue
- resolves https://github.com/oasis-tcs/csaf/issues/1208
  - add deep link to SSVC repo with decision points as informative reference
  - use new link in 6.1.48 and 6.3.13

### DIFF
--- a/csaf_2.1/prose/edit/src/introduction-04-informative-references.md
+++ b/csaf_2.1/prose/edit/src/introduction-04-informative-references.md
@@ -117,6 +117,9 @@ SSVC
 SSVC-RNS
 :    _Namespaces - SSVC: Stakeholder-Specific Vulnerability Categorization_, CERT/CC, <https://certcc.github.io/SSVC/reference/code/namespaces/#registered-namespace>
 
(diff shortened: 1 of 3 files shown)


sthagen:
It seems two tests fail in the SSVC arena, which is good, because we changed the world without changing tests, no?

tschmidtb51:
You are right. A cheers for test automation ;-)
I was lazy and did not check all tests before dinner - but I suspected that this could happen. Thank you for already analyzing and pointing out the change.

sthagen (approved):

Issue #1224, opened 2025-11-19
Editorial


tschmidtb51:
- [ ] Check order in 9.1
- [ ] Check removal of "Informative Comment" in 9.1

tschmidtb51:
- [x] add comment about vers and normalization / canonical form

PR #1234, opened 2025-11-26
Editorial

tschmidtb51:
- addresses parts of oasis-tcs/csaf#1224
- add comment about interpretation with normalization and canonical form

### DIFF
--- a/csaf_2.1/prose/edit/src/schema-elements-01-defs-02-branches.md
+++ b/csaf_2.1/prose/edit/src/schema-elements-01-defs-02-branches.md
@@ -156,7 +156,7 @@ The value of MUST obey to exactly one of the following options:
 
     vers MUST be used in its canonical form. To convey the term "all versions", the special string `vers:all/*` MUST be used.
 
-   > According to the interpretation used here, the canonical form includes that the vers is normalized.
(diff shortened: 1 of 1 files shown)


csaf_2.1/prose/edit/src/schema-elements-01-defs-02-branches.md
```diff
@@ -156,6 +156,8 @@ The value of MUST obey to exactly one of the following options:
 
     vers MUST be used in its canonical form. To convey the term "all versions", the special string `vers:all/*` MUST be used.
 
+   > According to the interpretation used here, the canonical form includes that the vers is normalized.
```
sthagen: We can always add informational content. What do we want to add to the MUST level requirement two lines above? Is it that "oh, by the way, canonical form requires normalization? Then we maybe best replace "includes" with "requires"

tschmidtb51: Should be solved with [9c6ccd9](https://github.com/oasis-tcs/csaf/pull/1234/commits/9c6ccd904bd832c7f369b7b60430ac0ba15d90d0)

Issue #1216, opened 2025-11-15
Licenses:  point out commercial tools


tschmidtb51:
We added the `license_expression` in CSAF 2.1. The feedback at the CSAF Community Days 2025 pointed out that we should also add a hint that restrictive licenses prevent users to easily use tools that are hosted in the cloud as they can not pre-import CSAF documents but users need to activate the sources themselves.

PR #1240, opened 2025-11-26
Licenses

tschmidtb51:
- resolves oasis-tcs/csaf#1216
- add comment about cloud tool

### DIFF
--- a/csaf_2.1/prose/edit/src/schema-elements-02-props-02-document.md
+++ b/csaf_2.1/prose/edit/src/schema-elements-02-props-02-document.md
@@ -369,7 +369,7 @@ through the ecosystem.
 
 > Restrictive licenses can prevent users to easily utilize tools that are hosted in the cloud as the tool providers cannot pre-import
 > CSAF document but users need at least to activate the sources themselves.
-> This applies for both, commercial and non-commercial, services and tools.
+> This is relevant for services and tools, regardless of whether they are commercial or not.
 
 #### Document Property - Notes{#document-property-notes}
 


sthagen (approved):

csaf_2.1/prose/edit/src/schema-elements-02-props-02-document.md
```diff
@@ -364,6 +364,13 @@ In addition, the following rules apply:
   MIT OR any-OSI
 ```
 
+When choosing a license expression, CSAF issuing parties SHALL consider to use a permissive license to enable the flow of information
+through the ecosystem.
+
+> Restrictive licenses can prevent users to easily utilize tools that are hosted in the cloud as the tool providers cannot pre-import
+> CSAF document but users need at least to activate the sources themselves.
+> This applies for both, commercial and non-commercial, services and tools.
```
sthagen: I think it is "applies to" or we could write "relevant for".
The both followed by two disjoint pairs being four "things" makes it a bit hard to grok. Maybe replace the second comma between the "pairs" with "as well as"?

tschmidtb51: Here is the sense what I intended to write:

> This is relevant for services and tools, regardless of whether they are commercial or not.

tschmidtb51: Should be fixed with: [1b60267](https://github.com/oasis-tcs/csaf/pull/1240/commits/1b60267c692597f2e8970c69c76fdd2801a0e28d)

sthagen (approved):

PR #1244, opened 2025-11-27
Update definition of 'checking' in retrieving rules

### LINKED ISSUE #1243: 7.3.1: Clarification on "requirement" when retrieving provider metadata
In CSAF 2.1 draft 7.3.1 where we describe the discovery of a `provider-metadata.json`, the sentence about "checking" needs clarification that it means the respective requirement of that check (namely requirement 8, 9 and 10).

oxisto:
* fixes #1243 by clarifying the definition of 'checking' in the context of retrieving provider metadata.

### DIFF
--- a/csaf_2.1/prose/edit/src/distributing-03-retrieving-rules.md
+++ b/csaf_2.1/prose/edit/src/distributing-03-retrieving-rules.md
@@ -23,7 +23,7 @@ Second, select one or more `provider-metadata.json` to use from the list of vali
 If the retrieving party is only able to process one `provider-metadata.json`, the first one in the list SHOULD be chosen.
 
 > The term "checking" used in the listing above SHOULD be understood as follows:
-> Try to access the resource and test whether the response provides an expected result as defined in the respective requirement (of that step) in section [sec](#requirements).
+> Try to access the resource and test whether the response provides an expected result as defined in the respective requirement of that step as defined in section [sec](#requirements).
 > If that is the case, the response is added to the list of candidates - otherwise not.
 > If the resource yields more than one response, the responses are added to the list in the order they are returned from the resource.
 


tschmidtb51 (approved):
LGTM

csaf_2.1/prose/edit/src/distributing-03-retrieving-rules.md
```diff
@@ -23,7 +23,7 @@ Second, select one or more `provider-metadata.json` to use from the list of vali
 If the retrieving party is only able to process one `provider-metadata.json`, the first one in the list SHOULD be chosen.
 
 > The term "checking" used in the listing above SHOULD be understood as follows:
-> Try to access the resource and test whether the response provides an expected result as defined in the requirement in section [sec](#requirements).
+> Try to access the resource and test whether the response provides an expected result as defined in the respective requirement (of that step) in section [sec](#requirements).
```
sthagen: Can we remove the parentheses? It is just an informal comment, but still the CCN could be 1 instead of 2, no?

tschmidtb51: > Can we remove the parentheses?

Not as easy as 

> the respective requirement of that step in section [sec](#requirements).

mean something different than

> the respective requirement in section [sec](#requirements).

The first one seems to link to the definition of steps - which would be the wrong link as that points to the requirements overall. The second one tells where to find the requirements.

Here is one alternative suggestion:
```suggestion
> Try to access the resource and test whether the response provides an expected result of that step as defined in the respective requirement in section [sec](#requirements).
```

tschmidtb51: Here is another one:

```suggestion
> Try to access the resource and test whether the response provides an expected result as defined in the respective requirement of that step as defined in section [sec](#requirements).
```

tschmidtb51: @oxisto, @sthagen  Thoughts? Preferences?

sthagen: I like this one.

oxisto: The second one is slightly better I think

tschmidtb51:
I'm not fully convinced that the addition does not add confusion (as the "step" in not defined in the linked section) but I'm going to merge it for now and hope that we improve the wording when fixing #1245.

PR #1247, opened 2025-11-30
Fixed undefined label hardware-and-software-within-the-product-tree-eg-1

sthagen:
Editorial Changes:

- [x] Fixed undefined label hardware-and-software-within-the-product-tree-eg-1
  - the example is directly following the section anchor, so we can easily refer to teh example 1 by teleporting the reader to the hosting section's title
  - this fixes the undefined label warning/error in PDF generation
  - we currently do have labels we can jump to only for selected structural elements like sections and tables

### DIFF
--- a/csaf_2.1/prose/edit/src/tests-02-rcmmndd-31-hardware-and-software.md
+++ b/csaf_2.1/prose/edit/src/tests-02-rcmmndd-31-hardware-and-software.md
@@ -51,5 +51,4 @@ The relevant paths for this test are:
 
 > The `product_tree` mentions the hardware product Example Company Controller A and combines it with the Firmware version 4.1.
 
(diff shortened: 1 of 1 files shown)


tschmidtb51:
As we have a couple of those links where we want to jump to examples: Can we implement that jump targets via labels? We are doing some stuff there in LaTeX for the frontpage, correct?

sthagen:
We may add link targets, but write for human readers. Most people seem to not know how to get back in their rendering applications to a recent reading position 🙈 and most examples in a section are alone there as example 1. So, we wish to offer that, but maybe when we got version 2.1 out.  Seriously, examples albeit separated formally from their embedding context often bring the most to readers when the readers approach them per the embedding section from above to understand the context of the example.

Issue #1226, opened 2025-11-24
`lowercase` vs `lower case`


tschmidtb51:
We write "lower case" but I have a hunch that "lowercase" might be correct. Native speaker, please...

justmurphy:
As a native English speaker, let stick to "lowercase" one word as the more widely accepted convention.

sthagen:
Should we use `lower-case` when using it as a compound adjective? Somewhere on the Internets I read that in this case the hyphenated form is the most accepted across the UK/US spelling "schism".

PR #1253, opened 2025-12-12
Lowercase

tschmidtb51:
- resolves oasis-tcs/csaf#1226
- unify spelling of "lowercase"


sthagen:
In case someone likes to revert my reviewer feedback change, please maintain the lowercase form of the compound adjective when it starts a list item (like for the other terms starting the subsequent items of that list).

justmurphy (approved):

Issue #1232, opened 2025-11-26
6.1.7: Clarify errors on unknown version


tschmidtb51:
We need to clarify whether an unknown (at least to CSAF) version should also result in an error. (Current reading is "yes").

PR #1268, opened 2026-01-23
Clarify errors on unknown versions (6.1.7)

tschmidtb51:
- resolves oasis-tcs/csaf#1232
- clarify that 6.1.7 also applies for unknown versions

### DIFF
--- a/csaf_2.1/prose/edit/src/tests-01-mndtr-07-multiple-scores-with-same-version-per-product.md
+++ b/csaf_2.1/prose/edit/src/tests-01-mndtr-07-multiple-scores-with-same-version-per-product.md
@@ -4,7 +4,7 @@ For each item in `/vulnerabilities` it MUST be tested that the same Product ID i
 
 > Different source might assign different scores for the same product.
 
-> The test also applies for versions not known by CSAF.
+> The test also applies for versions not included in CSAF.
 
 The relevant path for this test is:
 


csaf_2.1/prose/edit/src/tests-01-mndtr-07-multiple-scores-with-same-version-per-product.md
```diff
@@ -4,6 +4,8 @@ For each item in `/vulnerabilities` it MUST be tested that the same Product ID i
 
 > Different source might assign different scores for the same product.
 
+> The test also applies for versions not known by CSAF.
```
sthagen: I would write "known to" but then CSAF is not even a thing in this sentence, right? Maybe we can describe in a comment what the situation is, and then the best English wording should surface by itself.

tschmidtb51: Theoretically, a CVSS 3.2 could come along. Someone could put that in CSAF. Test 6.1.7 should alert on that instance even if the implementation does not known how to compute CVSS scores in CVSS v3.2.

sthagen: Thanks for presenting a use case. We can then maybe write instead:

> \[...] versions not included in CSAF.

Because, CSAF cannot act, it is only a format or specification thereof.
What do yout think?

tschmidtb51:
```suggestion
> The test also applies for versions not included in CSAF.
```

Issue #1211, opened 2025-11-07
Discourage open ended (upper) product version ranges


tschmidtb51:
We should add a recommended test to check whether open ended product version ranges are used. They predict the future and are rarely correct but mostly misused.

tschmidtb51:
Discussion 2025-12-17:
- (upper) open ended version ranges are a problem
- (lower) open ended version ranges are less a problem

PR #1275, opened 2026-01-24
Upper Open Ended Product Version Range

tschmidtb51:
- resolves oasis-tcs/csaf#1211
- add recommended test 6.2.49 to detect upper open ended product version ranges
- add invalid examples
- add valid examples
- add conversion rule
- update metadata


csaf_2.1/prose/edit/src/conformance.md
```diff
@@ -706,6 +706,32 @@ Secondly, the program fulfills the following for all items of:
 
     > A tool MAY provide a non-default option to output the invalid document.
 
+  * If the value of `name` of an item categorized as `product_version_range` and contains an upper open ended product version range,
```
sthagen, line 709: I tried rewriting esp. terms like 

> upper open ended product version range

with terms like 

> no upper limit

but fell short over the current newly suggested text.

Maybe an example in a comment would help to find an optimal term readers understand?

csaf_2.1/prose/edit/src/tests-02-rcmmndd-49-upper-open-ended-product-version-range.md
```diff
@@ -0,0 +1,30 @@
+### Upper Open Ended Product Version Range
+
+For each element of type `/$defs/branches_t` with `category` of `product_version_range`, it MUST be tested that the value of `name` is not an
+upper open ended product version range.
```
sthagen, line 4: Same here: I feel there must be some simpler wording not using six words: three for the openness and "direction" and three for the "measure" (the product version range). I found alternatives like half open interval, German terms like "nach oben offen" did not convince me either. Maybe, schrinking the "measure" to a "point" by using an acronym (PVR) might help focus on the "value situation" we are writing about?

csaf_2.1/prose/edit/src/conformance.md
tschmidtb51: I was tempted to use a mathematical term like

> right-open interval

but that wouldn't be correct at all.

An easy way out could be to define in the glossary 

> upper open ended product version range: a product version range that has no upper limit and therefore includes any future versions.

What do you think?

csaf_2.1/prose/edit/src/tests-02-rcmmndd-49-upper-open-ended-product-version-range.md
tschmidtb51: I used the right-open interval also somewhere in the text, but I realized this morning that this isn't correct at all as the half-open mean `[a,b)` (right-open) or `(a,b]` (left-open). I like the idea of using an acronym as it forces the reader to look how we defined it (and not just assume things), maybe in conjunction with my suggestion above?

Issue #1293, opened 2026-02-17
Editorial


tschmidtb51:
- [ ] Fix RSICV-M link (missing `id-`)

tschmidtb51:
- [ ] fix reference definition in 6.3.19
- [ ] point out, where abbreviations are defined

tschmidtb51:
- [ ] fix sorting in glossary
- [ ] fix sorting in examples

tschmidtb51:
- [ ] fix skus
  - [ ] add examples
  - [ ] adapt text from serial numbers
  - [ ] adapt conversion rule from serial numbers
  - [ ] check on tests

Issue #1059, opened 2025-06-16
Inconsistencies with use of numerals vs words


justmurphy:
There are multiple instances throughout the documentation where inconsistencies can be found with use of numerals vs words for the phrases such as: "at least one", "at least 1", "at least two", "at least 2", "at least 3", "with one", "with 1", "of one", "of 1", "exactly one", "exactly 3", etc.

![Image](https://github.com/user-attachments/assets/cdf54f4f-1f4c-432b-96f2-9a4effa676f2)

Convention is to use the word form for numbers zero through nine, and to write all other numbers as numerals (i.e. .

Applying the convention, "a minimum of 1" should be "a minimum of one", while "a minimum of 15" would be an appropriate use of the convention.

Recommend changes:

There are 62 instances where "with 1" is used. There are 10 instances where "with one" is used. It is recommended that all be changed to "with one".

Current state of incorrect usage:

> "with 1"

Recommended change:

> "with one"

There are multiple instances where "at least 1", "at least 2", or "at least 3" is used. It is recommended that all be changed to "at least one", "at least two", or "at least three".

Current state of incorrect usages:

> "at least 1"

> "at least 2"

> "at least 3"

Recommended changes:

> "at least one"

> "at least two"

> "at least three"

Happy to provide a checklist or a PR to help find identify places where the inconsistencies take place.

tschmidtb51:
Decision from the editors meeting today:

- any number in a `description` from the schema: use words
- any number derived from `minProperties`/`minItems`/`minLength` or `maxProperties`/`maxItems`/`maxLength`: use number in backticks
- any number providing a constraint for a tool (e.g. in 9.1.X): use number in backticks
- count of properties: use words

PR #1298, opened 2026-02-19
Editorial

tschmidtb51:
- addresses parts of https://github.com/oasis-tcs/csaf/issues/1293
- correct RVISC-M link and changing it to raw.githubusercontent.com domain
- write zero instead of 0 (to conform typographical convention set in https://github.com/oasis-tcs/csaf/issues/1059)

### DIFF
--- a/registry/id/schema/registry.schema.json
+++ b/registry/id/schema/registry.schema.json
@@ -92,7 +92,7 @@
       "title": "JSON schema",
       "description": "Contains the URL of the JSON schema for test data which the document promises to be valid for.",
       "type": "string",
-      "const": "https://raw.githubusercontent.com/oasis-tcs/csaf/master/registry/id/schema/id-registry.schema.json"
+      "const": "https://raw.githubusercontent.com/oasis-tcs/csaf/master/registry/id/schema/registry.schema.json"
     },
     "entries": {
       "title": "List of registry entries",


csaf_2.1/prose/edit/src/introduction-04-informative-references.md
```diff
@@ -106,7 +106,7 @@ RVISC
 :    _Registry for Vulnerability ID Systems for CSAF (RVISC)_, Part of the OASIS CSAF TC GitHub Repository, https://github.com/oasis-tcs/csaf/tree/master/registry/id.
 
 RVISC-M
-:    _Mapping for RVISC_, Part of the OASIS CSAF TC GitHub Repository, https://github.com/oasis-tcs/csaf/tree/master/registry/id/mapping.json.
+:    _Mapping for RVISC_, Part of the OASIS CSAF TC GitHub Repository, https://raw.githubusercontent.com/oasis-tcs/csaf/refs/heads/master/registry/id/id-mapping.json.
```
sthagen: Is there slomething else we plan to map (other than the id)? If not, why repeat the identifier.
Could we even use /registry/id/extensions.json? 

It is unfortunate, that we cannot offer a raw payload URL like the wrapped one, which allows to use blob/HEAD/ as branch selector for the default branch, instead of tree/master/ (for the wrapped) or refs/heads/master/ (for the raw).

If we could offer access to the registry through csaf.io or csaf.tech we could for example have:

- https://csaf.tech/registry/id/extensions/

Compared to:

- https://raw.githubusercontent.com/oasis-tcs/csaf/refs/heads/master/registry/id/extensions.json

the former is easier to communicate and to me looks more like it is taken "serious" (as a place to visit).

sthagen (approved):
Please see my comment on the URL / name of the registry.

csaf_2.1/prose/edit/src/introduction-04-informative-references.md
tschmidtb51: > Is there slomething else we plan to map (other than the id)?

Yes - the plan is not fully solid yet, but I think we want to provide something for the `x_generic_uris` (e.g. `/registry/x_generic_uri/namespace-registry.json` and `/registry/x_generic_uri/namespace-mapping.json`).
I guess I used `id-mapping.json` instead of just `mapping.json` to avoid confusion in filenames if we have multiple mappings (that just differentiate in path). However, if you think, if would be better to drop the `id-`, I'm open to discuss that. Should that then also apply for the `id-registry.json`?

>  Could we even use /registry/id/extensions.json?

It would more likely be `/registry/extension/registry.json`, `/registry/extension/deprecated.json`, `/registry/extension/deny.json`.

> If we could offer access to the registry through csaf.io or csaf.tech we could for example have:

That is definitely easy to do on the csaf.io domain (but the owner did not want to codify the domain in the standard the last time I asked and nobody knows what happens to the `.io` TLD as it is currently a country one and [the British Indian Ocean Territory does not longer exist](https://en.wikipedia.org/wiki/.io#Future)).
So, we could use csaf.tech or csaf.dev... I guess both would be possible. The question is: Do we want to have a redirect or a copy there... Or do we want to have that on our website?

sthagen: We can always use the csaf.tech domain. I know the owner and they operate rapid.rocks sionce years ... (for the OData TC).

On the paths: I beleive paths should be irreducible and plausible. Ideally, for simple result-sets hierarchically ordered. I was irritated by the length and the anti-climax reading that like a REST or folder path: The mapping to me is more the result type and not the function. Like foo/id/extension to me is close to a function like foo_id(extension) -> MappingType: in Python (kind of). Let us find a good way during the Editor meeting today. Thanks for providing details and outlook on use cases.

csaf_2.1/prose/edit/src/introduction-04-informative-references.md
```diff
@@ -106,7 +106,7 @@ RVISC
 :    _Registry for Vulnerability ID Systems for CSAF (RVISC)_, Part of the OASIS CSAF TC GitHub Repository, https://github.com/oasis-tcs/csaf/tree/master/registry/id.
```
sthagen:
```
:    _Registry for Vulnerability ID Systems for CSAF (RVISC)_, Part of the OASIS CSAF TC Registry, <https://registry.csaf.dev/id/>.
```

csaf_2.1/prose/edit/src/introduction-04-informative-references.md
sthagen:
```
:    _Mapping for RVISC_, Mapping part of the OASIS CSAF TC Registry, <https://registry.csaf.dev/id/mapping/>.
```

csaf_2.1/prose/edit/src/introduction-04-informative-references.md
tschmidtb51: Done in [0a550da](https://github.com/oasis-tcs/csaf/pull/1298/commits/0a550da3ace41c14ab400584826cb4fe5e8ba97a)

csaf_2.1/prose/edit/src/introduction-04-informative-references.md
tschmidtb51: Done in [0a550da](https://github.com/oasis-tcs/csaf/pull/1298/commits/0a550da3ace41c14ab400584826cb4fe5e8ba97a)

PR #1450, opened 2026-05-14
Makefile

### LINKED ISSUE #1449: makefile: set SHELL
The `SHELL` variable defaults to `/bin/sh` if not set in the makefile. Under Linux, it is not taken from the environment, so we should set it manually (cf. https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html)
The target `render-pdf` uses `pushd` which is not available in `/bin/sh`.

tschmidtb51:
- resolves oasis-tcs/csaf#1449
- add SHELL variable

### DIFF
--- a/csaf_2.1/prose/edit/makefile
+++ b/csaf_2.1/prose/edit/makefile
@@ -1,5 +1,3 @@
-SHELL=/bin/bash
-
 .POSIX:
 .SILENT:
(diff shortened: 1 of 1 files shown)


sthagen (approved):

tschmidtb51:
@sthagen Thank you!

PR #1452, opened 2026-05-16
Added distclean and amended clone-info/name targets

sthagen:
Maintenance only changes.

The `clone-info` target (with alias `name`) is now more informative for editors:

```bash
❯ make clone-info
Clone-info:
- name.is(): 'csaf'
- branch.is(): 'makefile-enhancements'
- revision.is(): sha1:092a4dfe461208be9ec2a11cc4b714cd8961ea92
- location.is(): (orga/repo='oasis-tcs/csaf', rel-path='csaf_2.1/prose/edit')
- node.id(): 'c79891e5-aabf-3a83-95b9-588edcd8327f'
```

Note: The usefulness of the location info relies in part on the placement on the local disk.
If it had been at `/some/funny/path` in above example, the output for `orga/repo` would have been `funny/path`.

### DIFF
--- a/csaf_2.1/prose/edit/makefile
+++ b/csaf_2.1/prose/edit/makefile
@@ -134,8 +134,18 @@ clean:
 	rm -fr $(build_root)
 	mkdir -p $(build_root)
 
+.PHONY: distclean
+distclean: clean
+	rm -fr $(build_root)
+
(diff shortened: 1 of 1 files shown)


tschmidtb51 (approved):
LGTM

Issue #820, opened 2024-10-31
Provide an expected failure "code" or "message" in testcases.json


oxisto:
This is a follow-up to the discussion in https://github.com/csaf-sbom/kotlin-csaf/issues/80#issuecomment-2448675774. We make use of the TC test cases JSONs in testing the validation tests in kotlin-csaf. While exploring options to keep tests in sync with the TC, we came up with the idea whether it is possible to include an expected failure "message" in the testcases.json. This way we could align the error messages in our library with the official ones and we could query the testcases.json to automatically assert good and bad test cases even with the appropriate error.

Further discussion points
- Does it make sense to introduce some kind of official error codes for the validation?
- Should it be mandatory for CSAF libraries to use these error codes or messages

This would make developing compliant CSAF libraries much more easier, sort of like a [Technology Compatibility Kit](https://en.wikipedia.org/wiki/Technology_Compatibility_Kit).

tschmidtb51:
We need to take into consideration, what is already existing, e.g. the codes used by the [csaf-validator-lib](https://github.com/secvisogram/csaf-validator-lib?tab=readme-ov-file#interfaces).

The `instancePath` is crucial as it provides the opportunity for viewers and editors to highlight it message in the right place.

tschmidtb51:
@oxisto It looks like the issue was submitted before you became a TC member. Therefore, please announce it on the TC comment list to fulfill the formal process requirements.

PR #1475, opened 2026-06-15
Comment Resolution CSD02PR: #1363

### LINKED ISSUE #1363: Clarification for missing item in revision history 6.1.21
What is the expectation for test 6.1.21 (missing items in revision history) when the input is not sorted by date? 
Let's assume a revision history containing (ordered by date) 2, 1 and 3. 
Of course, there will be errors from test 6.1.14, but what should test 6.1.21 do?
- Be skipped, because the revision history is not sorted by date?
- Report an error that version 1 is missing because the first revision is 2, and 
  report an error, that revision 2 is missing between 1 and 3?
- Report success, because there will be no missing items once the date has been corrected?
 
If the third behaviour is desired, the standard should be adapted so that it no longer says
> when the items are sorted ascending by `date`,

but instead 
> when the items are sorted ascending by `number`

---

Next case, assume a revision history (order by date) with 4, 5, 6. I expect two different errors here:
- First item is not 0 or 1
- Missing items 0 (or 1) to 3

Is that assumption correct?

Announced: https://groups.oasis-open.org/discussion/new-issue-clarification-for-missing-item-in-revision-history-6121

tschmidtb51:
- resolves #1363
- addresses parts of #820
- clarify sorting by date
- clarify that missing (in general) and missing in the sorted list must be different error messages
- clarify second level sorting criteria `number`
- add results
- add invalid examples
- add valid examples

### DIFF
--- a/csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-01.result.json
+++ b/csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-01.result.json
@@ -5,7 +5,7 @@
     "errors": [
       {
         "instance_path": "/document/tracking/revision_history",
-        "message": "missing revision history item with number `2`"
+        "message": "missing revision history item with number `2` at all"
       }
     ],
     "id": "6.1.21",
--- a/csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-01.result.json
+++ b/csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-01.result.json
@@ -1,0 +1,15 @@
+{
+  "$schema": "https://raw.githubusercontent.com/oasis-tcs/csaf/master/csaf_2.1/test/validator/testresult_json_schema.json",
+  "overall_valid": false,
+  "primary_result": {
+    "errors": [
+      {
+        "instance_path": "/document/tracking/revision_history",
+        "message": "missing revision history item with number `2`"
+      }
+    ],
+    "id": "6.1.21",
+    "passed": false
+  },
+  "resultschema_version": "2.1"
+}
--- a/csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-02.result.json
+++ b/csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-02.result.json
@@ -9,7 +9,7 @@
       },
       {
         "instance_path": "/document/tracking/revision_history",
-        "message": "missing revision history item with number `1`"
+        "message": "missing revision history item with number `1` at all"
       }
     ],
     "id": "6.1.21",
--- a/csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-02.result.json
+++ b/csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-02.result.json
@@ -6,6 +6,10 @@
       {
         "instance_path": "/document/tracking/revision_history",
         "message": "revision history does not start with a version of `0` or `1` when sorted by date (was `2`)"
+      },
+      {
+        "instance_path": "/document/tracking/revision_history",
+        "message": "missing revision history item with number `1`"
       }
     ],
     "id": "6.1.21",
--- a/csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-02.result.json
+++ b/csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-02.result.json
@@ -1,0 +1,15 @@
+{
+  "$schema": "https://raw.githubusercontent.com/oasis-tcs/csaf/master/csaf_2.1/test/validator/testresult_json_schema.json",
+  "overall_valid": false,
+  "primary_result": {
+    "errors": [
+      {
+        "instance_path": "/document/tracking/revision_history",
+        "message": "revision history does not start with a version of `0` or `1` when sorted by date (was `2`)"
+      }
+    ],
+    "id": "6.1.21",
+    "passed": false
+  },
+  "resultschema_version": "2.1"
+}
--- a/csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-03.result.json
+++ b/csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-03.result.json
@@ -5,7 +5,7 @@
     "errors": [
       {
         "instance_path": "/document/tracking/revision_history",
-        "message": "missing revision history item with number `2`"
+        "message": "missing revision history item with number `2` at all"
       }
     ],
     "id": "6.1.21",
--- a/csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-03.result.json
+++ b/csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-03.result.json
@@ -1,0 +1,15 @@
+{
+  "$schema": "https://raw.githubusercontent.com/oasis-tcs/csaf/master/csaf_2.1/test/validator/testresult_json_schema.json",
+  "overall_valid": false,
+  "primary_result": {
+    "errors": [
+      {
+        "instance_path": "/document/tracking/revision_history",
+        "message": "missing revision history item with number `2`"
+      }
+    ],
+    "id": "6.1.21",
+    "passed": false
+  },
+  "resultschema_version": "2.1"
+}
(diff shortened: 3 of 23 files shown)


csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-05.result.json
tziemek, line 1: The 05 test file would also lead to the following errors from 6.1.14:
- Misplaced number 2
- Misplaced number 1
- Misplaced number 4

which is then really confusing. Are 1 and 4 only misplaced or are they missing? 
Or do we only report "the items must be in ascending order according to the date" from 6.1.14?

sthagen (approved):

csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-05.result.json
tschmidtb51: >  Or do we only report "the items must be in ascending order according to the date" from 6.1.14?

Nearly: "items must be in ascending order when sorted by `date` and `number`"

> Are 1 and 4 only misplaced or are they missing?

We don't know for sure. Interpretation is that the `date` is correct (therefore sorted by date first) - `number` might be incorrect (as counting is hard). So it is most likely that the version entries are in fact missing in the `revision_history`. But for every example one can come up with there is at least one example where it was the other way around. Therefore, we can't automatically fix the issue.

tziemek: So the expected results would be
- "items must be in ascending order when sorted by date and number" from 6.1.14 (we don't report what numbers or dates are not according to the spec, but only the fact that it is not correct; this way we can also stop at first misplaced number)
- "missing number 2, 1, 4" (as separate messages) from 6.1.21 (albeit they might exist somewhere; which could be tested rather easily)

Is this correct?

tschmidtb51: IMHO clarified in [d5d2ce2](https://github.com/oasis-tcs/csaf/pull/1475/commits/d5d2ce286a4cb3752b59dec985113bc23525adde).
@tziemek Please double check.

*Note: I chose to not use the 'between'-phrasing for any error message occurring before the first item when sorted by `date` and `number` as the lower bound would be unclear. One could argue that this should be the `initial_release_date` but I'm not sure whether I would rely on that if I have a broken `revision_history`.*

tziemek: Yes, this is much clearer now, thank you.

csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-04.result.json
```diff
@@ -0,0 +1,15 @@
+{
+  "$schema": "https://raw.githubusercontent.com/oasis-tcs/csaf/master/csaf_2.1/test/validator/testresult_json_schema.json",
+  "overall_valid": false,
+  "primary_result": {
+    "errors": [
+      {
+        "instance_path": "/document/tracking/revision_history",
+        "message": "missing revision history item with number `2.0.0`"
```
tziemek:
```suggestion
        "message": "missing revision history item with number `2.0.0` between `2023-08-22T10:00:00.000Z` and `2024-01-21T10:00:00.000Z`"
```

csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-03.result.json
```diff
@@ -0,0 +1,15 @@
+{
+  "$schema": "https://raw.githubusercontent.com/oasis-tcs/csaf/master/csaf_2.1/test/validator/testresult_json_schema.json",
+  "overall_valid": false,
+  "primary_result": {
+    "errors": [
+      {
+        "instance_path": "/document/tracking/revision_history",
+        "message": "missing revision history item with number `2`"
```
tziemek:
```suggestion
        "message": "missing revision history item with number `2` between `2023-08-22T10:00:00.000Z` and `2024-01-21T11:00:00.000+03:00`"
```

csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-02.result.json
```diff
@@ -0,0 +1,19 @@
...
+  "overall_valid": false,
+  "primary_result": {
+    "errors": [
+      {
+        "instance_path": "/document/tracking/revision_history",
+        "message": "revision history does not start with a version of `0` or `1` when sorted by date (was `2`)"
+      },
+      {
+        "instance_path": "/document/tracking/revision_history",
+        "message": "missing revision history item with number `1`"
```
tziemek:
```suggestion
        "message": "missing revision history item with number `1` before `2023-08-22T10:00:00.000Z`"
```

csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-01.result.json
```diff
@@ -0,0 +1,15 @@
+{
+  "$schema": "https://raw.githubusercontent.com/oasis-tcs/csaf/master/csaf_2.1/test/validator/testresult_json_schema.json",
+  "overall_valid": false,
+  "primary_result": {
+    "errors": [
+      {
+        "instance_path": "/document/tracking/revision_history",
+        "message": "missing revision history item with number `2`"
```
tziemek:
```suggestion
        "message": "missing revision history item with number `2` between `2023-08-22T10:00:00.000Z` and `2024-01-21T10:00:00.000Z`"
```

tziemek (commented):
I made suggestions for the other files.

csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-04.result.json
tschmidtb51: We need to differentiate the error messages between an item missing in the list at all (quick fix: insert a stub with that number in the right place) and an item where the version exists in the list but is missing when sorted by `date` and `number` (quick fix: insert a stub and deal with the other item appropriately by assigning a new number to it; this will have a bigger impact on the revision history as multiple changes might be needed). I updated the error message as discussed over the phone.

tschmidtb51: We need to differentiate the error messages between an item missing in the list at all (quick fix: insert a stub with that number in the right place) and an item where the version exists in the list but is missing when sorted by `date` and `number` (quick fix: insert a stub and deal with the other item appropriately by assigning a new number to it; this will have a bigger impact on the revision history as multiple changes might be needed). I updated the error message as discussed over the phone.

csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-03.result.json
tschmidtb51: We need to differentiate the error messages between an item missing in the list at all (quick fix: insert a stub with that number in the right place) and an item where the version exists in the list but is missing when sorted by `date` and `number` (quick fix: insert a stub and deal with the other item appropriately by assigning a new number to it; this will have a bigger impact on the revision history as multiple changes might be needed). I updated the error message as discussed over the phone.

csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-01.result.json
tschmidtb51: We need to differentiate the error messages between an item missing in the list at all (quick fix: insert a stub with that number in the right place) and an item where the version exists in the list but is missing when sorted by `date` and `number` (quick fix: insert a stub and deal with the other item appropriately by assigning a new number to it; this will have a bigger impact on the revision history as multiple changes might be needed). I updated the error message as discussed over the phone.

csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-21-02.result.json
tschmidtb51: We need to differentiate the error messages between an item missing in the list at all (quick fix: insert a stub with that number in the right place) and an item where the version exists in the list but is missing when sorted by `date` and `number` (quick fix: insert a stub and deal with the other item appropriately by assigning a new number to it; this will have a bigger impact on the revision history as multiple changes might be needed). I updated the error message as discussed over the phone.

sthagen (approved):

justmurphy (approved):

ndemianc (approved):

Issue #1446, opened 2026-05-13
Question: unique / non-unique `product_identification_helpers`


peinjoh:
The product identification helpers arrays `purls`, `model_numbers`, `serial_numbers` have `unique: true` set.

What is the reason why these 3 properties are required to have unique items, but others, like `skus`, `sbom_urls`, `x_generic_uris`, are allowed to contain non-unique items?

What information is conveyed / how is product identification being aided by having the same sku string / url / uri present multiple times?

This was announced here: https://groups.oasis-open.org/discussion/new-issue-question-unique-non-unique-product-identification-helpers

tschmidtb51:
> What is the reason why these 3 properties are required to have unique items, but others, like `skus`, `sbom_urls`, `x_generic_uris`, are allowed to contain non-unique items?
>
> What information is conveyed / how is product identification being aided by having the same sku string / url / uri present multiple times?

None. IMHO that is an oversight that should be corrected.

PR #1483, opened 2026-06-26
Unique PIH

tschmidtb51:
- resolves oasis-tcs/csaf#1446
- make items in `hashes`, `file_hashes`, `sbom_urls`, `skus` and `x_generic_uris` unique
- adapt prose
- add conversion rule
- correct JSON path mistake

### DIFF
--- a/csaf_2.1/json_schema/csaf.json
+++ b/csaf_2.1/json_schema/csaf.json
@@ -179,6 +179,7 @@
               "description": "Contains a list of cryptographic hashes usable to identify files.",
               "type": "array",
               "minItems": 1,
+              "uniqueItems": true,
               "items": {
                 "title": "Cryptographic hashes",
                 "description": "Contains all information to identify a file based on its cryptographic hash values.",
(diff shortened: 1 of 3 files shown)


sthagen (approved):
Maybe the user can profit from us reviewing the existing prose (around these changes (the use of the terms "non-default option" and "excepted").
 
1. "non-default option" to do quux - I think we mean that there is an optional behavior to do quux, where being optional means that the user has to explicitly enable that.
2. "excepted" - I think "exempted" or "excluded" might be better separated from "accepted" - or is this a typo and we really want to write "accepted"?

tschmidtb51:
> Maybe the user can profit from us reviewing the existing prose (around these changes (the use of the terms "non-default option" and "excepted").
> 
>     1. "non-default option" to do quux - I think we mean that there is an optional behavior to do quux, where being optional means that the user has to explicitly enable that.
> 
>     2. "excepted" - I think "exempted" or "excluded" might be better separated from "accepted" - or is this a typo and we really want to write "accepted"?

tracked as #1485

tschmidtb51:
Motion done during TC meeting 2026-06-24 in #1446

Issue #1491, opened 2026-07-22
Message in oasis_csaf_tc-csaf_2_1-2024-6-1-14-01.result.json differs from oasis_csaf_tc-csaf_2_1-2024-6-1-14-02.result.json


rainer-exxcellent:
The message in :

https://github.com/oasis-tcs/csaf/blob/1a28d6b2d29265f4a034e38422204b2c042d9ea8/csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-14-01.result.json#L8

differs from:

https://github.com/oasis-tcs/csaf/blob/1a28d6b2d29265f4a034e38422204b2c042d9ea8/csaf_2.1/test/validator/data/mandatory/oasis_csaf_tc-csaf_2_1-2024-6-1-14-02.result.json#L8

IMHO we should not use backticks ` for quotation. Lexically correct is the single quote.
In addition, the backtick has special functions in many languages

tschmidtb51:
@rainer-exxcellent Thank you for spotting the discrepancy. Regarding backticks: My plan was to write the message directly as markdown - which has some nice advantages for future plans with theses messages... So, I'm not yet convinced by your argument...
@sthagen Thoughts?

rainer-exxcellent:
Announced in: https://groups.oasis-open.org/discussion/new-issue-1491-message-in-oasis-csaf-tc-csaf-2-1-2024-6-1-14-01resultjson-differs-from-oasis-csaf-tc-csaf-2-1-2024-6-1-14-02resultjson

PR #1493, opened 2026-07-22
Test results: 6.1.14

tschmidtb51:
- resolves oasis-tcs/csaf#1491
- fix inconsistent message


tziemek:
@tschmidtb51 Could you also correct invalid case 2 as the csaf file is the same as case 1. I think case 2 should be using semver versioning.

Sorry, I missed the changed current version in these files. All good :)

sthagen (approved):

PR #1507, opened 2026-07-28
Fixed wording on profile 4.1 clarification

### LINKED ISSUE #1454: Base profile
We need to double check that we have all mandatory fields included in the profile section.

oxisto:
Changed the wording on profile 4.1 (csaf_base) category clarification

Fixes #1454


tschmidtb51:
@justmurphy This is mostly word-smiting - so I would like to see a review from you as native speaker.

justmurphy (approved):

PR #1508, opened 2026-07-28
Clarify requirements on CSAF content management systems

### LINKED ISSUE #1456: CSAF content management system
Check list of template-based / prefilled items.

oxisto:
This aligns the requirements of the CSAF content management system, mainly with regards to required fields. Basically we add an additional validation check before the document reaches an `interim` or `final` state.

Fixes #1456

### DIFF
--- a/csaf_2.1/prose/edit/src/conformance-01-cls-06-content-management-system.md
+++ b/csaf_2.1/prose/edit/src/conformance-01-cls-06-content-management-system.md
@@ -68,6 +68,7 @@ A CSAF Content Management System satisfies the "CSAF Content Management System" 
 
   - `$['$schema']` with the value prescribed by the schema
   - `$.document.csaf_version` with the value prescribed by the schema
+  - `$.document.distribution.tlp.label` with a default of `TLP:CLEAR`
   - `$.document.lang`
   - `$.document.notes`
     - `legal_disclaimer` (Terms of use from the configuration)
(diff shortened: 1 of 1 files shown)


csaf_2.1/prose/edit/src/conformance-01-cls-06-content-management-system.md
```diff
@@ -68,6 +68,7 @@ A CSAF Content Management System satisfies the "CSAF Content Management System"
 
   - `$['$schema']` with the value prescribed by the schema
   - `$.document.csaf_version` with the value prescribed by the schema
+  - `$.document.distribution.tlp.label` with a default of `TLP:CLEAR`
```
tschmidtb51, line 71:
```suggestion
  - `$.document.distribution.tlp.label` (based on the templates from configuration; default: "TLP:CLEAR")
```

csaf_2.1/prose/edit/src/conformance-01-cls-06-content-management-system.md
tschmidtb51, line 94: Why are we not requiring the TLP here?

tschmidtb51 (changes requested):
We should also add `license_expression` (based on the templates from configuration) here to add a push/incentive for this important information.

Please also sort the items lexicographically (not your fault, but as we touch it, we should improve it).

csaf_2.1/prose/edit/src/conformance-01-cls-06-content-management-system.md
oxisto: the whole section is a bit confusing. what is the intent of listing TLP here? if its already set, I don't really need to update it? Or should it update the document if the value is not set (but its not worded that way)

tschmidtb51: The second requires the system to support certain steps a user has to do anyway when creating a new version (so hopefully changing the license does not belong to this :wink:)

As a user of the system, I might want to share drafts default as "TLP:AMBER+Strict" (quite usual during coordination) and just change to TLP:CLEAR when going into a released version.
And I realize that this use case might make it necessary to add another requirement...

sthagen (approved):

PR #1523, opened 2026-08-14
Fixed inconsistend date in citation example and adapted bin/bump.py

sthagen:
- [x] Fixed inconsistent date in citation example
- [x] Adapted bump tool to the brave new world


tschmidtb51 (approved):
LGTM

PR #1526, opened 2026-08-14
Bumped checkout, setup-python, and upload-artifact actions in render runner declaration

sthagen:
The title says it all.


tschmidtb51 (approved):
LGTM

Issue #1511, opened 2026-07-29
Restructure `contact_details`


tschmidtb51:
We should restructure `contact_details` to be able to provide a dedicated field for an email address as well as the corresponding public openpgp key.
Structure will be:

```yaml
<csaf-instance>:
  document:
    publisher:
      # ...
      contact:
         details: String                               # catch all
         email: String.Email                       # just the primary email contact
         public_openpgp_url: String.URL  # the URL to the public openpgp of the email above
      # ...
```

tschmidtb51:
2026-07-29 TC meeting: Thomas moved to accept the change into CSAF 2.1, Sergii seconded, *Justin clarified that the issue will be created afterwards and motion will be documented after creation*, no objection, no discussion, unambiguous consent. Motion carried

tschmidtb51:
Todo:

- [x] adapt schema
- [x] adapt prose
- [x] tests?
  - [x] 6.3.6
  - [x] `public_openpgp_key_url` should provide directly key
  - [x] `public_openpgp_key_url` should contain matching email in user ID
- [x] guidance on size
- [x] conversion rules
- [x] base profile
- [x] conformance targets
  - [ ] ~6 CSAF Content Management System~ *no change needed - already covering "`$.document.publisher` and children"*
- [x] examples

peinjoh:
Should the public_openpgp_key_url be part of the Non-Self Referencing URLs tested by 6.3.6?

adulau:
Just a small question, why not having all the CNA/GNA fields as optional for a publisher?

~~~json
  {
    "partner": "Computer Incident Response Center Luxembourg (CIRCL)",
    "scope": "Vulnerability assignment related to vulnerability coordination in Luxembourg, which may include reports on software, products, hardware, or services submitted to CIRCL in its role as the national authority for vulnerability coordination in Luxembourg and within the CSIRTs network.",
    "program_role": "CNA",
    "organization_type": "CERT",
    "country": "Luxembourg",
    "metadata": {
      "cna_id": "CNA-2026-0027",
      "gna_id": "1",
      "short_name": "CIRCL",
      "contact": [
        {
          "email": [
            {
              "label": "Email",
              "emailAddr": "info@circl.lu"
            }
          ],
          "contact": [],
          "form": []
        }
      ],
      "disclosure_policy": [
        {
          "label": "Policy",
          "language": "",
          "url": "https://www.circl.lu/pub/coordinated-vulnerability-disclosure/"
        }
[...]

PR #1548, opened 2026-08-21
Contact Details

tschmidtb51:
- addresses parts of https://github.com/oasis-tcs/csaf/issues/1511
- rework `contact_details` to become object with individual fields
- add `public_openpgp_key_url`
- adapt schema description
- add new outline


sthagen (approved):

sei-vsarvepalli:
@tschmidtb51 
May be this was discussed but I missed it, 

- Will this Contact have a Contact_ID similarly to product_id - allowing for it to be cross-referenced throughout the document ?
- How will this Contact be associated say with sharing_group ? 
- Can this Contact be also cross-referenced in the document for the pending `involvements` so throughout the document any Contact is distinctly recognizable?
- Can this Contact be also cross-reference in document.distribution.sharing_groups

csaf_2.1/test/validator/data/informative/oasis_csaf_tc-csaf_2_1-2024-6-3-06-02.json
```diff
@@ -3,7 +3,7 @@
   "document": {
     "acknowledgments": [
       {
-        "summary": "A URL that does resolve with HTTP status code 400.",
+        "summary": "A URL that does resolve with HTTP status code >=400.",
```
sthagen, line 6: Let us not use operators here. "400 or higher"? Also, are these client and server errors or only client errors?

tschmidtb51: This example returns a 404 (client error). But the test in general should also cover server errors and informational responses. However, they are somehow hard to test. We could use something like https://httpbingo.org, https://httpbin.org/ or https://http.dev/tools but as far as I see they are all owned and operated privately (so nothing by W3C or IANA or IETF that would be safe to use). Are you aware of any official service that we could use?
We could provide something on our own (e.g. https://status.http.test.csaf.dev) but the question is then, who operates that (as it would need to be in operation as long as CSAF lives). How did you solve that is MQTT or other TCs?

tschmidtb51:
> @tschmidtb51 May be this was discussed but I missed it,
> 
>     * Will this Contact have a Contact_ID similarly to product_id - allowing for it to be cross-referenced throughout the document ?

No. This is just a contact point for the issuing party. Currently, I would say we should not couple these two.

>     * How will this Contact be associated say with sharing_group ?

Not. A sharing group is a distribution identifier. There are no plans currently to assign a publisher to a sharing group. The linking of values to users happens outside of a CSAF Document.

>     * Can this Contact be also cross-referenced in the document for the pending `involvements` so throughout the document any Contact is distinctly recognizable?

No. This contact is intended to be limited to serve as contact information for the issuing party.

>     * Can this Contact be also cross-reference in document.distribution.sharing_groups

Not inside a CSAF document - and there is no plan to do so. The sharing groups are handled outside of the CSAF document itself.

sei-vsarvepalli:
Okay understood the Contact Details is here is meant really for "Issuing Party Contact" and it is NOT all contacts of stakeholders involved in the vulnerability advisory.

csaf_2.1/test/validator/data/informative/oasis_csaf_tc-csaf_2_1-2024-6-3-06-02.json
sthagen: AMQP, DSS, MQTT, MQTT-SN, ODATA, and SARIF all have strong support from existing implementations and their origanizations - not so much from OASIS Open when it comes to providing supportive infrastructure for interoperability or "dynamic" responses.

On the text here. Given that we also need to consider server side errors I suggest to follow others like (https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status) and list the full ranges: 
> "summary": "A URL that does resolve with an HTTP status code in the ranges of 400-499 or 500-599",

tschmidtb51: I guess we should discuss the wording details in a meeting - it feels like we are talking past each other - but maybe I'm just tired.. :face_with_peeking_eye:

sthagen (approved):

Issue #1555, opened 2026-08-26
MUST vs SHALL


tschmidtb51:
As we target ISO, we should consider to use MUST and SHALL in the way [ISO wants us to use them](https://www.iso.org/sites/directives/current/part2/index.xhtml):

- SHALL: is the preferred form to define a requirement (see [ISO/IEC Directives, Part 2 — Principles and rules for the structure and drafting of ISO and IEC documents: section 7.2](https://www.iso.org/sites/directives/current/part2/index.xhtml#_idTextAnchor082)
- MUST: is used to state external constraints (see [ISO/IEC Directives, Part 2 — Principles and rules for the structure and drafting of ISO and IEC documents: section 7.6](https://www.iso.org/sites/directives/current/part2/index.xhtml#_idTextAnchor095)

*Note: We SHALL ensue that the use is consistent with BCP 14 (which is the binding keyword definition for us as defined in [section 2](https://github.com/oasis-tcs/csaf/blob/master/csaf_2.1/prose/share/csaf-v2.1-draft.md#terminology))*

PR #1560, opened 2026-08-26
Editorial: MUST vs SHALL

tschmidtb51:
- addresses parts of oasis-tcs/csaf#1555
- change MUST to SHALL where applicable
- correct normative language in informative sections


tschmidtb51:
@sthagen: I saw some instance where we use normative (capitalized) keywords in informative text. I guess that is okay for MAY, but we should talk about the other instance in our meeting to decide how to deal with them.

sthagen (approved):

Issue #1554, opened 2026-08-25
Adapt redirect reference to Fetch standard section (minor)


bernhardreiter:
https://github.com/oasis-tcs/csaf/blob/27af37f5c9a107abb166e855fa6505fc4ffaade8/csaf_2.1/prose/edit/src/distributing-01-requirements.md?plain=1#L49

should be changed to something like
```md
This aligns with the "HTTP-redirect fetch" section of the [cite](#FETCH) specification (accessed 2026-08-25).
```

as currently, it is section 4.5 and not 4.4.

As https://fetch.spec.whatwg.org/#http-redirect-fetch  is a "living" standard, it can change any day. Giving the date either here or in the reference section will identify which revision of the standard was meant precisely.

When added to the standard in August 2025, https://github.com/oasis-tcs/csaf/commit/d28c4a2caf650fca8add5f20aad2066106536fa1,
it was section 4.4. See https://web.archive.org/web/20250823170448/fetch.spec.whatwg.org/

bernhardreiter:
Mentioned with https://groups.oasis-open.org/discussion/21-distributing-01-requirementsmd-reference-fetch-better

tschmidtb51:
Closed via #1524 (through integration of #1562)

PR #1562, opened 2026-08-27
FETCH

tschmidtb51:
- addresses parts of oasis-tcs/csaf#1554
- remove section number from FETCH as it is a living standard


tschmidtb51:
@sthagen The issue originally suggested to include an accessed date (as it is a living standard the document seems to change at some point so that we can't predict stable references). What do you think?

sthagen (approved):

sthagen:
@tschmidtb51 
> @sthagen The issue originally suggested to include an accessed date (as it is a living standard the document seems to change at some point so that we can't predict stable references). What do you think?

Fully in favor. Hyperlinks (even more so with fragments) are overrated. Meaning with stability always wins in my book.

sthagen (approved):

Issue #1519, opened 2026-08-14
Declare usage of AI in CSAF advisories in notes


cplutte:
I propose to add a `note.category` `ai_usage` to facilitate the declaration of the usage of AI in the vulnerability management process. This note category can be used during vulnerability reporting as well as in publication of CSAF advisories to declare if and how AI tooling was used for example during the discovery, triaging or report generation process.

## Motivation
With the increasing adoption of AI tooling, it is more and more common to see suspicion that certain artefacts were created by AI without declaration. To address this, we suggest to introduce a way to explicitly declare the usage of AI. The `note `element is the simplest way to provide such an option, details can be provided in the note text and the `note `can be used on the `document `as well as on the `vulnerability`-level.

## Definition: 
`ai_usage` — Provides information on whether and how AI systems were used in discovering, analyzing, remediating, or documenting the vulnerability, including the human oversight applied.

## Guidance for producers

An `ai_usage` note SHOULD state, for each activity where AI was involved:

1. **What the AI did** — discovery, validation, severity assessment, product matching, remediation development, authoring, document generation, translation, or enrichment.
2. **How far it went** — whether it suggested, drafted, or produced the result as used.
3. **What review was applied** — in particular whether a human confirmed the technical claims independently (for example by reproducing the issue or checking versions against source) or only read the output and found it plausible. This is the distinction recipients care about most.
4. **Whose work it was**, where the note describes another party's declaration rather than the publisher's own practice.

Naming the model or tool is OPTIONAL. Where a producer maintains a published AI usage policy, the note SHOULD reference it.

## Example:
```
 {
    "document": {
      "category": "csaf_vex",
      "notes": [
        {
          "category": "ai_usage",
          "title": "AI usage",
          "text": "An automated code-review agent (internal tooling, build 2026.7) flagged the unchecked length field during a scheduled
  scan. The finding was reproduced manually against release 4.2.1 by the researcher before submission; no AI system was involved in
  validation. The description and impact text in this report were drafted by a general-purpose LLM and edited by the researcher."
        }
      ]
    }
  }
```

 ```
{
    "category": "ai_usage",
    "title": "AI usage",
    "text": "No AI or machine learning system contributed to the discovery, analysis, remediation, or authoring of this advisory."
  }
```

### Further questions

**Naming.** Is `ai_usage` the right value name? Alternatives: `ai_disclosure` (emphasises the act of declaring), `automation` (covers non-AI tooling such as scanners, but loses the specific concern).

[...]

PR #1568, opened 2026-08-31
AI Usage

tschmidtb51:
- addresses parts of oasis-tcs/csaf#1519
- add special category/title combination for AI usage

### DIFF
--- a/csaf_2.1/prose/edit/src/schema-elements-02-props-02-document.md
+++ b/csaf_2.1/prose/edit/src/schema-elements-02-props-02-document.md
@@ -361,12 +361,13 @@ The following combinations of `category` and `title` have a special meaning and 
 
 \columns=15%,35%,
 
-| category           | title                            | content of text                                                                                                                                                                               |
-|:-------------------|:---------------------------------|:----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| description        | Product Description              | Contains a description of a product given in the `product_tree` in regard to field of application and core functionality. This SHOULD be bound to the corresponding product or product group. |
-| general            | General Security Recommendations | Contains general advise and security recommendations that are related, generic and might be independently applicable of the content of the CSAF document.                                     |
-| legal\_disclaimer  | License                          | Contains the only license text of the document license.                                                                                                                                       |
-| summary            | Summary                          | Contains a short summary of the content of the advisory.                                                                                                                                      |
+| category           | title                            | content of text                                                                                                                                                                                    |
+|:-------------------|:---------------------------------|:---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| description        | Product Description              | Contains a description of a product given in the `product_tree` in regard to field of application and core functionality. This SHOULD be bound to the corresponding product or product group.      |
+| general            | AI Usage                         | Contains information on whether and how AI systems were used in discovering, analyzing, remediating, or documenting the content of the whole CSAF document, including the human oversight applied. |
+| general            | General Security Recommendations | Contains general advise and security recommendations that are related, generic and might be independently applicable of the content of the CSAF document.                                          |
+| legal\_disclaimer  | License                          | Contains the only license text of the document license.                                                                                                                                            |
+| summary            | Summary                          | Contains a short summary of the content of the advisory.                                                                                                                                           |
 
 Table: Requirements for combinations of `category` and `title` that have a special meaning.{#tab:special-combinations-of-category-and-title}
 
--- a/csaf_2.1/prose/edit/src/schema-elements-02-props-04-vulnerabilities.md
+++ b/csaf_2.1/prose/edit/src/schema-elements-02-props-04-vulnerabilities.md
@@ -711,11 +711,12 @@ The following combinations of `category` and `title` have a special meaning and 
 
 \columns=12%,24%,
 
-| category    |  title                | content of text                                                                                                                                      |
-|:------------|:----------------------|:-----------------------------------------------------------------------------------------------------------------------------------------------------|
-| description | CVE Description       | Contains the official and unchanged CVE description for this specific vulnerability.                                                                 |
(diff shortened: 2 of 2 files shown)


tschmidtb51:
@cplutte Please review.

tschmidtb51:
@sthagen I added the new category-title-combination in the document and vulnerability part separately to make it more accessible. Alternative way would be to add it in the defs-notes part. However, that would
1. create a new table with just one entry (as we don't have a table there yet and no other combinations have been defined),
2. IMHO make it less accessible (as being in a separate place),
3. does not longer stress the different scopes.

Should we keep it this way or extract it into the type and link the new table from the document and vulnerability part.
Thoughts?

cplutte:
looks good to me

peinjoh:
Not part of the changeset, so fixing this someplace else might be a good idea, but I believe there are two typos in that section:
`Contains general advise` -> `Contains general advice` (verb vs. noun)
`might be independently applicable of the content` -> I think this should either be `independently applicable to the content` or `applicable independently of the content`, depending on what is meant here.

csaf_2.1/prose/edit/src/schema-elements-02-props-04-vulnerabilities.md
```diff
@@ -711,11 +711,12 @@ The following combinations of `category` and `title` have a special meaning and
...
-| summary     | Vulnerability Summary | Contains a summary of the vulnerability which is not the official CVE description.                                                                   |
+| category    |  title                | content of text                                                                                                                                                                           |
+|:------------|:----------------------|:------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| description | CVE Description       | Contains the official and unchanged CVE description for this specific vulnerability.                                                                                                      |
+| description | Preconditions         | Contains a description of the preconditions that have to be fulfilled to be able to exploit this specific vulnerability, e.g. user account or physical access.                            |
```
sthagen, line 717: I wonder what `this specific vulnerability` adds to `the vulnerability` in the detached table display? The table makes it hard to refer to any row rule. Why did we not use sections and within definition lists? Something like:

```
### 42.3.7 Constraints on Text for Combinations of Categories and Titles

#### 42.3.7.1 Special Meaning of Titles for the Description Category

For the description category the combination of the following title values impose the following expectations on the content of the text:

CVE Description
:    Contains the official and unchanged CVE description for this specific vulnerability. 

Preconditions
:    Contains a description of the preconditions that have to be fulfilled
     to be able to exploit the vulnerability, e.g. user account or physical access.

#### 42.3.7.2 Special Meaning of Titles for the General Category

For the general category the combination of the following title value impose the following expectation on the content of the text:

AI Usage
:    Contains information on whether and how AI systems were used in 
     discovering, analyzing, remediating, or documenting this specific 
     vulnerability, including the human oversight applied.

#### 42.3.7.3 Special Meaning of Titles for the Summary Category

For the summary category the combination of the following title value impose the following expectation on the content of the text:

Vulnerability Summary
:    Contains a summary of this specific vulnerability which is not
     the official CVE description.
```

csaf_2.1/prose/edit/src/schema-elements-02-props-02-document.md
```diff
@@ -361,12 +361,13 @@ The following combinations of `category` and `title` have a special meaning and
...
+| general            | AI Usage                         | Contains information on whether and how AI systems were used in discovering, analyzing, remediating, or documenting the content of the whole CSAF document, including the human oversight applied. |
+| general            | General Security Recommendations | Contains general advise and security recommendations that are related, generic and might be independently applicable of the content of the CSAF document.                                          |
+| legal\_disclaimer  | License                          | Contains the only license text of the document license.                                                                                                                                            |
+| summary            | Summary                          | Contains a short summary of the content of the advisory.                                                                                                                                           |
```
sthagen, line 370: SHOULD in the third column cell of a table is not my favorite nad the rows are hard to reference.  We could transform these triplets into subsections per category as sketched in https://github.com/oasis-tcs/csaf/pull/1568/changes#r3906132486 and provide definition lists where the special title value is the term and the description/definition part explains the expectation.

sthagen (approved):
I can live with the tables, but would prefer sections for discoverability and referenceability.

csaf_2.1/prose/edit/src/schema-elements-02-props-04-vulnerabilities.md
tschmidtb51: > I wonder what `this specific vulnerability` adds to `the vulnerability` in the detached table display?

Just trying to be consistent in the wording. Alternative wording would be `the current vulnerability`.

> The table makes it hard to refer to any row rule.

Sorry, I'm not sure whether I understand what the change should be. Should we add numbers to the table? :thinking: 

>  Why did we not use sections and within definition lists? 

There is no hard reason but IMHO it is easier to refer to a single table from the converter conformance targets than having to refer to multiple sections (were I'm most likely going to forget to add it in all places ones another category/title pair is added). Personally, I think the table might make it easier to digest but I haven't seen the other form as rendered result, so I can't really speak to that.

tschmidtb51:
> I can live with the tables, but would prefer sections for discoverability and referenceability.

I'm a bit confused right now: How do we replace the reference to the table (for example, in the conformance targets) if the values are then listed in multiple sections? Do we have to list them all individually? Wouldn't that be prone to errors?
We can either discuss that during the editor meeting or if you want, feel free to provide an alternative PR.

csaf_2.1/prose/edit/src/schema-elements-02-props-04-vulnerabilities.md
sthagen: I updated the sketch ^^^ with a wrapping section for single reference. I would assume we are inside one vulnerability item of the vulnerabilities, which makes the vulnerability an exact reference.

tschmidtb51: As this is unrelated to this PR - I would like to split that out and deal with that in a separate editorial issue => see #1575

PR #1573, opened 2026-09-01
11 years time warp

sthagen:
Feel free to retarget to master branch - whatever is more convenient for you, @tschmidtb51 

# Description

The jsonpath-rw package is from April 2015 an installs in the bin folder of the python environment a script called jsonpath.py.

This unfortunately shadows the import of jsonpath from the python-jsonpath module.

Symptom: make render fails.

Workaround: remove the bin/jsonpath.py file from that active python environment after install of jsonpath-rw.
With pyenv one can find that path by executing pyenv which jsonpath.py

Solution (IMO): Just use python-jsonpath which is already installed by nide.

This minimally changed version of the "strictifier" script works with python-jsonpath ...

### DIFF
--- a/.github/workflows/csaf_2.0_main.yml
+++ b/.github/workflows/csaf_2.0_main.yml
@@ -21,11 +21,10 @@ jobs:
         sudo apt-get update -q && sudo apt-get install -y --no-install-recommends -qq \
           python3 \
           python3-simplejson \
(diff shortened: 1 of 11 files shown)


tschmidtb51 (approved):
LGTM - Thank you so much!

Issue #1482, opened 2026-06-25
Adoption of CSAF for vulnerability reports submission pre-embargo using TLP markings and phased release to public CSAF report.


sei-vsarvepalli:
## Background

This is a result of few discussion in FIRSTCON26 with partners including Ericson PSIRT Christoph Plutte - Our discussion surrounded around handling of volumetric reporting from AI discovered vulnerabilities including the Frontier Model generated reports should be standardized to provide good reliable workflow for coordinators and for vendors. Perhaps of interest to @tschmidtb51 as well.

Currently CERT/CC VINCE has been moving to an API plugin to do so, but this seems bigger than us as the number of vulnerability reports continues to grow. 

## Summary

I would like to propose work in this repository to support broader adoption of **CSAF as the standard interface for vulnerability report submission**, especially during the **pre-embargo** phase using **TLP markings**.

Today, vulnerability reports may be authored by security researchers or generated/assisted by frontier AI models. If the eventual public advisory is expected to be CSAF, it would be valuable for initial private submissions to also be represented in CSAF-compatible structures from the beginning.

## Motivation

Using CSAF earlier in the disclosure lifecycle can improve consistency, reduce translation errors, and accelerate coordination among reporters, vendors, and coordinators.

A structured CSAF submission format for pre-embargo reports can help:
- Set clear expectations for required and optional fields.
- Improve machine-readability and automation readiness.
- Make handoff from initial report to coordinated disclosure and eventual publication more efficient.
- Support outreach to impacted parties using consistent and actionable data.

## Proposal

Develop guidance and examples for using CSAF in pre-embargo vulnerability submission workflows, including:

1. **Examples**
   - Minimal and complete sample pre-embargo CSAF documents.
   - Examples for reports from:
     - Individual researchers
     - Coordinators/CERTs
     - AI-assisted or AI-generated draft reports (with human validation context)

2. **Profile / extension guidance (if needed)**
   - Identify whether current CSAF fields are sufficient for pre-embargo submission use cases.
   - If gaps exist, propose best-practice patterns and/or extension mechanisms.
   - Clarify how to represent uncertainty, confidence, and evolving findings before public disclosure.

3. **TLP usage guidance**
   - Clear recommendations for expressing and handling TLP markings in pre-embargo CSAF content.
   - Mapping of TLP intent to practical sharing constraints in submission and coordination workflows.

4. **Interoperability and process expectations**
[...]

PR #1582, opened 2026-09-07
Vulnerability Report Profile

tschmidtb51:
- addresses parts of oasis-tcs/csaf#1482
- remove uncheckble parts


tschmidtb51:
@sei-vsarvepalli I suggest to remove the parts as stated in this PR. Reasons - tests do not seem to be applicable broadly. Nevertheless, we should discuss that.

sthagen (approved):

csaf_2.1/prose/edit/src/profiles.md
sei-vsarvepalli, line 1: Working in involvements in a distinct PR I presume. so this is okay as-is

sei-vsarvepalli (approved):
All looking right - as we move more of the involvements work into its own setup.

Issue #1549, opened 2026-08-22
Clarify: Extension detection in 6.2.36 and 6.3.15


peinjoh:
The prose of 6.2.26 (and 6.3.15, but for Non-TLP:CLEAR) state:

> For each SSVC decision point given under selections, it MUST be tested that the namespace does not use an extension if the document is labeled TLP:CLEAR. Namespaces reserved for special purpose MUST be treated as per their definition.

During the implementation as part of csaf-rs, multiple questions came to mind that I would like to clarify:

1) Does this test require the namespace to be valid in the first place? Syntactical validity is to a big part ensured via the schema by a regex, but even if within the schema/regex, the namespace can still, for example, use the `invalid` / `x_invalid` base namespace or use a invalid BCP-47 language tag (like `ssvc/EZ`). Should a test finding still be reported in that case? 
This assumes strict parsing. We are currently discussing non-strict parsing, where more way-to-be-invalid would be possible. Or would matching the schema/regex be a prequisite for this test to run there?

2) If that is the case, i.e. if namespace validity is not required for this test to run, is the namespace containing the the extension delimiter `/` enough to identify it to contain extensions?

3) More of a nit: Does the default first extension, exp. `ssvc/`, which implies `ssvc/en-US`, and to my understanding is equivalent to `ssvc`, match this test description? The informal comment describes the reason for this test as extensions potentially changing the decision point definitions, and the changed definition being potentially unknown to the readers. I would argue this would not be the case here though, as `ssvc/` represents the decision points in their original/default language.

Also: I believe Question/Clarification 1 also has implications for the other "namespaces meets criteria" tests.

It also touches 6.1.46, which checks that "SSVC object is valid according to the referenced schema". Could the invalid BCP-47 tags mentioned above also be reported on there?

Announced here: https://groups.oasis-open.org/discussion/new-issue-clarify-extension-detection-in-6226-and-6315

tschmidtb51:
1. Yes. If we do not have a valid namespace, there is not much we can do.
2. Should be.
3. Not sure whether `ssvc/` should be allowed at all... Need to double check.

tschmidtb51:
Todo

- [x] clarify that 6.1.46 includes checking for languages
- [x] specify implementers note "looking for /" is sufficient
  - [x] 6.2.36
  - [x] 6.3.15
- [x] clarify special case `ssvc/en-US` => is extension per ABNF, I suggest to treat it as such

peinjoh:
[...]

PR #1606, opened 2026-09-10
SSVC - implementation guidance

tschmidtb51:
- resolveshttps://github.com/oasis-tcs/csaf/issues/1549
- add language check as part of test 6.1.46
- add implementation hint to 6.2.36 and 6.3.15 about extension detection
- add implementation hint to 6.2.34, 6.2.35, 6.3.14 about base namespace determination
- add invalid examples
- add valid example

### DIFF
--- a/csaf_2.1/prose/edit/src/tests-01-mndtr-46-invalid-ssvc.md
+++ b/csaf_2.1/prose/edit/src/tests-01-mndtr-46-invalid-ssvc.md
@@ -1,6 +1,8 @@
 ### Invalid SSVC
 
 It SHALL be tested that the given SSVC object is valid according to the referenced schema.
+If the object contains a language code, for example as a part of a `namespace`,
+it SHALL also be tested that the language code is valid and exists.
 
 The relevant path for this test is:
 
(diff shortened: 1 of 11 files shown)


csaf_2.1/prose/edit/src/tests-01-mndtr-46-invalid-ssvc.md
```diff
@@ -1,6 +1,8 @@
 ### Invalid SSVC
 
 It SHALL be tested that the given SSVC object is valid according to the referenced schema.
+If the object contains a language code, for example as a part of a `namespace`,
+it SHALL also be tested that the language code is valid and exists.
```
sthagen, line 5: Can we drop `anmd exists`? Seems like a distinction without a difference to me.

sthagen (approved):

justmurphy (approved):

csaf_2.1/prose/edit/src/tests-01-mndtr-46-invalid-ssvc.md
tschmidtb51: Expectation is that there are different error messages for:
- language tags are (partially) not exist in the registry at all 
- language tags exists but compiled language code is not valid (e.g. in the wrong place)

See #1610

tschmidtb51:
Motion passes in https://github.com/oasis-tcs/csaf/issues/1549#issuecomment-5617898833

